PR #3, opened 2018-03-27
 Update types & pal. for the needs of desktop app

meriadec:


src/types/account.js
```diff
@@ -0,0 +1,135 @@
...
+  // minimal nb of blocks to consider an operation confirmed (set by the user)
+  minConfirmations: number,
+
+  // Actually used in the desktop app
+  // TODO: should get rid of that if possible
+  index: number,
+
+  // Actually used in the desktop app
+  // TODO: should be moved to separate reducer, I think. Because balance is
+  // derived data, it's not really data.
```
gre: yeah, basically could use the helpers/account functions ;)

PR #5, opened 2018-03-28
add exporter/importer of 'bridgestream'

gre:
# bridgestream

bridgestream is a bridge that allow one-directional stream communication from one legder app to another.

It currently allows to export Accounts from the desktop app to the mobile app.

### Desktop code gist

```js
import { makeChunks } from "@ledgerhq/wallet-common/lib/bridgestream/exporter";
const chunks = makeChunks({
  accounts,
  exporterName: "desktop",
  exporterVersion: "0.0.0"
});
// loop for each chunk of chunks[i]: display chunk QRCode, sleep a bit of time
```

### Mobile code gist

```js
import {
  parseChunkR, areChunksComplete, chunksToResult
} from "@ledgerhq/wallet-common/lib/bridgestream/importer";

class Scanning extends Component<{
  onResult: Result => void
}> {
  lastData: ?string = null;
  chunks: Data[] = [];
  completed: boolean = false;
  onBarCodeRead = ({ data }: { data: string }) => {
    if (data && data !== this.lastData && !this.completed) {
      this.lastData = data;
      this.chunks = parseChunksReducer(this.chunks, data);
      if (areChunksComplete(this.chunks)) {
        this.completed = true;
        this.props.onResult(chunksToResult(this.chunks));
      }
    }
  };

  ...
```

### DIFF
--- a/src/__tests__/bridgestream/__snapshots__/exporter.js.snap
+++ b/src/__tests__/bridgestream/__snapshots__/exporter.js.snap
@@ -1,0 +1,19 @@
+// Jest Snapshot v1, https://goo.gl/fbAQLP
+
+exports[`basic export 1`] = `
+Array [
+  "[4,0,\\"meta\\",1,\\"test\\",\\"0.0.0\\"]",
+  "[4,1,\\"account\\",\\"mock_account_export_0\\",\\"pQNKgto\\",141]",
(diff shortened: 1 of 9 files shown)


gre:
@meriadec yeah i'm not sure neither. in my mind I was thinking that our format will always be backward compatible. e.g. importer version 9 understand all version before of the exporter. however, if you have a new exporter, an older importer don't understand it and would just display "upgrade your app" , but not sure. hopefully we never changes the format or only add more fields  that would get ignored by an older importer

PR #9, opened 2018-04-08
Clean scripts and babel config

meriadec:


gre:
test issue with jest 🤔

package.json
```diff
@@ -1,15 +1,12 @@
 {
   "name": "@ledgerhq/wallet-common",
   "version": "0.13.0",
-  "main": "src/index.js",
+  "main": "lib/index.js",
   "license": "MIT",
   "scripts": {
-    "preinstall": "node -e \"if (process.env.npm_execpath.indexOf('yarn') === -1) { console.log('\u001b[31mPlease use yarn\u001b[0m'); process.exit(1); }\"",
     "flow-typed": "flow-typed install -s",
-    "build": "yarn clean && yarn documentation && babel --source-maps -d lib src && flow-copy-source src lib",
```
gre: why no longer building the documentation?

package.json
```diff
@@ -1,15 +1,12 @@
...
-    "watch": "babel --watch --source-maps -d lib src && flow-copy-source -w -v src lib",
-    "prepare": "yarn flow-typed && yarn build",
```
gre: I think prepare was important, because it is called when doing a publish & you really want to build before publish.

that said, maybe we can have our own publish script and forbid us to use directly the `npm publish` / `yarn publish`

package.json
meriadec: I think it's better to clearly separate concerns rather than a big "do everything" button. We can still add a "build-lib-and-documentation" script if we want.

package.json
meriadec: The problem with having a prepare script is we no longer benefit of the yarn integrity check when we do `yarn`: everything is always rebuilt,  no matter something has changed. As you said, we may have our own publish script if we really want a "all-in-one" build and publish.

package.json
gre: so maybve just putting it in `prepare`, or whatever the script to be called for the publish

package.json
gre: yeah it's actually weird `yarn` would call `prepare` 😅 any other hook to override the "publish"? "prepublish" ?

package.json
meriadec: the problem with `prepare` is the fact it is *always* triggered when installing/reinstalling deps with `yarn` ( = `yarn install`), which is highly annoying.

PR #10, opened 2018-04-10
Mock account generation improvements

meriadec:
- prevent account from generate negative balance
- ability to specify number of generated operations

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "flow-typed": "flow-typed install -s",
-    "build": "rm -rf lib && babel -sd lib src && flow-copy-source src lib",
+    "build": "rm -rf lib && babel --ignore __tests__ -sd lib src && flow-copy-source -i \"__tests__/**\" src lib",
     "prepublish": "yarn build && yarn documentation",
     "watch": "babel -wsd lib src && flow-copy-source -wv src lib",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -5,7 +5,7 @@
   "license": "MIT",
   "scripts": {
     "flow-typed": "flow-typed install -s",
-    "build": "rm -rf lib && babel -sd lib src && flow-copy-source src lib",
+    "build": "rm -rf lib && babel --ignore __tests__ -sd lib src && flow-copy-source -i \"__tests__/**\" src lib",
```
gre: :+1:

src/mock/account.js
```diff
@@ -115,10 +122,16 @@ export function genAccount(id: number | string): Account {
...
+
+  let total = 0;
   account.operations = Array(operationsSize)
     .fill(null)
     .reduce(ops => {
       const op = genOperation(account, ops, currency, rng);
+      if (total + op.amount < 0) {
+        op.amount = -op.amount;
+      }
+      total += op.amount;
```
gre: how does it work? genOperation don't go back in time as i thought?

meriadec: Everytime before a new op is added to list, we ensure that the total balance will be positive. If not, we instead add the `-op.amount`.

gre: but genOperation goes back in time: https://github.com/meriadec/ledger-wallet-common/blob/8d899c190a5ec5149fadb72766c0ab36128dcedf/src/mock/account.js#L41-L51

so i'm not sure this works 🤔 balance needs to be calculated reversely

gre: let's discuss more tomorrow, but i'm not sure this works 🤔  maybe it does & i totally missed something the other day :o .

PR #14, opened 2018-05-03
adding etc and xrp

NastiaS:


src/data/cryptocurrencies.js
```diff
@@ -78,6 +78,14 @@ const cryptocurrenciesArray: CryptoCurrency[] = [
     color: "#ffae35",
     units: ethereumUnits
   },
+  {
+    id: "ethereum_classic",
+    name: "Ethereum Classic",
+    ticker: "ETC",
+    scheme: "ethereum",
+    color: "#ffae35",
+    units: ethereumUnits
```
gre: needs to be another list, where the first item will be `code: "ETC"`

also, check with designers what will be ETC color (if defined).

finally, I think we need a different scheme, maybe for now you can put "ethereumclassic" in scheme until we really figure it out.. i think in the future, we will rethink this scheme thing, because generating QRCode with this won't be supported everywhere maybe (like third party app that scan this.. maybe we should allow no scheme & fallback on diff string)

src/data/cryptocurrencies.js
```diff
@@ -167,6 +175,27 @@ const cryptocurrenciesArray: CryptoCurrency[] = [
       }
     ]
   },
+  {
+    id: "xrp",
+    name: "XRP",
+    ticker: "XRP",
+    scheme: "ripple",
+    color: "#000",
```
gre: sure about the color?

src/data/cryptocurrencies.js
```diff
@@ -167,6 +175,27 @@ const cryptocurrenciesArray: CryptoCurrency[] = [
       }
     ]
   },
+  {
+    id: "xrp",
+    name: "XRP",
```
gre: name: Ripple

gre:
@NastiaS can you run `yarn test -u` i think this is because our mock generator is randomly picking different coin now.

PR #18, opened 2018-05-11
Prettier & update ETH testnet config

meriadec:


gre:
oh yeah actually i think atom-prettier is stuck on the tests, it just don't want to prettify anymore in all the tests, don't ask why 🙃

PR #20, opened 2018-05-13
Adjust Account/AccountRaw types

meriadec:
- add `isSegwit`
- rename `rootPath` to `walletPath`

### DIFF
--- a/flow-defs/globals.js
+++ b/flow-defs/globals.js
@@ -1,0 +1,6 @@
+/* eslint-disable */
+
+declare var describe: Function
+declare var test: Function
+declare var it: Function
+declare var expect: Function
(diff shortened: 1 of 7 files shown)


flow-defs/globals.js
```diff
@@ -0,0 +1,6 @@
+/* eslint-disable */
+
+declare var describe: Function
+declare var test: Function
+declare var it: Function
+declare var expect: Function
```
gre: mmh, do you have the **jest** type def in the `flow-typed` directory? it's supposed to define them, if not it feels it's a regression maybe 🤔

meriadec: you're right. works well after using `yarn flow-typed` (maybe we should put it on postinstall :trollface: )

PR #21, opened 2018-05-19
add coinType back

gre:
we'll probably need it to calculate the bip32 path when we first communicate with device to scan accounts


meriadec:
welcome back

PR #22, opened 2018-05-19
drop balanceByDay (unused)

gre:


gre:
i think we'll merge this because we will try not to use balanceByDay for poc. we'll come back to this "cache" idea later

gre:
@meriadec it's not used to draw the charts anymore. now we are calculating with the list of operations. I think we'll continue to do this for the MVP & reconsider the balanceByDay cache only later when we only rely on libcore (otherwise that will be more work to implement everywhere)

gre:
see https://gist.github.com/gre/4daf6ab3a2c65fa5ffeaa6f0e2c11d7a

meriadec:
Ok makes sense :ok_hand: 

If we calculate on our side, what about `account.balance` which is currently [calculated by libcore](https://github.com/LedgerHQ/ledger-live-desktop/blob/f488f36900d074b0df1e31288e429104eb742d1d/src/internals/accounts/scanAccountsOnDevice.js#L188-L189) for BTC like cryptos?

As you said, if we rely only on operations to calculate balance, we need to have all operations loaded on JS side.

gre:
oh yeah, balance will definitely be calculated by libcore & implementations, because for this one we can't assume having all operations over. for the balance by day, we can assume to have enough recent operations to display the X last days

gre:
(with the total balance + some recent operations, we can have balanceByDay calculated without problem)

gre:
so the basic idea of this PR is to temporarily remove this from the model because unused & when we figure it out we can add it back

PR #29, opened 2018-05-25
add missing cryptos

gre:


gre:
I have not verified if the magnitude are correct. i was assuming it was like BTC but not verified, let's double check

PR #51, opened 2018-07-10
Add a new field in CryptoCurrency: managerAppName

gre:


src/data/cryptocurrencies.js
```diff
@@ -1,8 +1,7 @@
 //@flow
 /**
  * ADDING A NEW COIN to the frontend stack:
- * You need to add the coin in cryptocurrenciesById,
- * We sort coins by their current market cap rank so please respect this order.
```
gre: (nb we no longer will keep things ordered by marketcap. up to consumer to sort by whatever order they want)

PR #53, opened 2018-07-16
Migrate to BigNumber – fixes #47

### LINKED ISSUE #47: refactor using a BigNumber implementation
BigInt vs BigNumber ?
which library? (whatever is most popular i think)

see also https://github.com/tc39/proposal-bigint/issues/53

gre:
and also drop deprecated stuff, because we'll do a major for this anyway.


src/helpers/currencies/CurrencyURIScheme.js
```diff
@@ -42,10 +43,10 @@ export function decodeURIScheme(str: string): Data {
   const { amount, ...specificFields } = { ...query };
   Object.assign(data, specificFields);
   if (amount) {
-    const amountFloat = parseFloat(amount);
-    if (!isNaN(amountFloat) && amountFloat > 0) {
+    const amountFloat = BigNumber(amount);
+    if (!amountFloat.isNaN() && amountFloat.isPositive()) {
```
gre: FIXME isPositive actually includes ZERO. wasn't aware of this

gre: yeah i know aha. xD

gre: maybe it will just be better to always use `isGreaterThan(0)` because there is also `isGreaterThanOrEqualTo(0)` 
a bit verbose but explicit

gre: OH there is also `gt` and `gte` synthax

gre: [](https://raw.githubusercontent.com/MikeMcl/bignumber.js/gh-pages/API.png)

PR #56, opened 2018-07-29
[add] adds ubiq

drewrothstein:
Adds `ubiq`.
Looks like the `svg` is already present.
The value for `blockAvgTime` is taken from https://ubiqsmart.com/.
The value for `units` is the same as `ethereumUnits` save for the first.

First commit to this repository so I am happy to follow whatever process/procedure you would like.

Once https://github.com/LedgerHQ/ledger-live-common/pull/55 is shipped, I can run tests with this.
I plan on making the appropriate changes in `ledger-live-desktop` after this ships.

### DIFF
--- a/src/data/cryptocurrencies.js
+++ b/src/data/cryptocurrencies.js
@@ -22,7 +22,10 @@
  *
  * ~~ also ~~
  *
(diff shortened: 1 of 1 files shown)


gre:
thanks!

PR #64, opened 2018-08-21
Fix over-truncating amounts 0.5 or less away from the next power of 10

MortalKastor:
### Type

Bugfix

### Description

Fix a bug where, for an `amount` with `d` digits and `n` a number between `0.5` and `0.1`, if `amount + n` has `d + 1` digit, the last decimal is cut off. eg `9.8` gets truncated to `9.`

### Demo

Before:
9 5 buggy

After:
9 5 fixed

### Affected Ledger Live versions
All versions since 1.1.0

### DIFF
--- a/src/helpers/currencies/BigNumberToLocaleString.js
+++ b/src/helpers/currencies/BigNumberToLocaleString.js
@@ -53,18 +53,16 @@ export const toLocaleString = (
   const maxDecimals = bn.toFormat(maximumFractionDigits);
   if (maximumFractionDigits !== minimumFractionDigits) {
     const minDecimals = bn.toFormat(minimumFractionDigits);
-    let i = maxDecimals.length - 1;
+    let i = maxDecimals.length;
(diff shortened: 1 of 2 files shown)


src/helpers/currencies/BigNumberToLocaleString.js
```diff
@@ -53,18 +53,16 @@ export const toLocaleString = (
...
-        i--;
-      } else if (maxDecimals[i] === format.decimalSeparator) {
-        i--;
+      if (maxDecimals[i - 1] !== "0") {
+        if (maxDecimals[i - 1] === format.decimalSeparator) {
+          i--;
+        }
         break; // we reach decimal. stop now.
-      } else {
-        i++; // we eat one character that we shouldn't. we stop there and roll it back (nb slice won't overflow)
```
MortalKastor: This part was never reached for `9.5` because `minDecimals` is rounded to `10` so `i > minDecimals.length` was false by the time the last char has already been removed by the previous loop cycle.

PR #68, opened 2018-08-28
change p2sh version for dgb

amougel:


amougel:
DGB changed their p2sh version and now segwit p2sh addresses start with "S" instead of "3". The new DGB app for device will also use this version.

gre:
@amougel ouch, so potentially we will have to bump this at same time a new version is out on manager and urge everyone to update?

amougel:
Yes. The Live will receive an update for supporting new digibyte address format and the manager will provide an updated app at the same time

PR #72, opened 2018-08-30
change zen name to horizen

amougel:


gre:
you need to keep using zencash.svg filename

PR #81, opened 2018-10-05
Implement frames variants

gre:
- to be tested
- there are some breaking changes in the API, but no breaking changes in the data format itself


gre:
will refactor to using qrloop. this code had bugs.

idea would be to no longer have bridgestream but just utilities and some convention of an export data compressed format. then can just pass this data to qrloop

PR #84, opened 2018-10-10
Re-work the account types (seedIdentifier, derivationMode)

gre:
(WIP, will include some breaking changes and we'll write some migration in desktop app)


meriadec:
noice! :rocket:  is it possible to squash you commits for moar readability?

PR #88, opened 2018-10-24
add more legacy addresses for eth, etc and xrp

amougel:
I added the missing legacy derivations from chrome apps
These are single addresses (no iteration possible)

### DIFF
--- a/src/derivation.js
+++ b/src/derivation.js
@@ -36,9 +36,21 @@ export const getDerivationScheme = ({
     // old MEW derivation scheme for ETC
     return "44'/60'/160720'/0'/<account>";
   }
-  if (derivationMode === "rip") {
-    // XRP legacy that the old Chrome Ripple Wallet used to wrongly derivate addresses on.
-    return "44'/144'/0'/<account>'";
+  if (derivationMode === "ripWalletIdentifier") {
+    // XRP legacy that the old Chrome Ripple Wallet used to wrongly derivate addresse on.
+    return "44'/144'/0'/0'";
+  }
+  if (derivationMode === "ripWalletMetaPassword") {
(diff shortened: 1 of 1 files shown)


src/derivation.js
```diff
@@ -36,9 +36,21 @@ export const getDerivationScheme = ({
     // old MEW derivation scheme for ETC
     return "44'/60'/160720'/0'/<account>";
   }
-  if (derivationMode === "rip") {
```
gre: This must not disappear, otherwise will create regression in existing accounts.

src/derivation.js
```diff
@@ -36,9 +36,21 @@ export const getDerivationScheme = ({
...
-    // XRP legacy that the old Chrome Ripple Wallet used to wrongly derivate addresses on.
-    return "44'/144'/0'/<account>'";
+  if (derivationMode === "ripWalletIdentifier") {
```
gre: Basically call this one rip for backward compatible

src/derivation.js
```diff
@@ -36,9 +36,21 @@ export const getDerivationScheme = ({
...
+    // XRP legacy that the old Chrome Ripple Wallet used to wrongly derivate addresse on.
+    return "44'/144'/0'/0'";
+  }
+  if (derivationMode === "ripWalletMetaPassword") {
```
gre:
```suggestion
  if (derivationMode === "rip2") {
```

amougel:
1. > I think Ledger Live will expect all derivations to be iterable

That's what I thought so I looked into the code and realized that there just a `replace` occuring on 
the path string (for account). So it should not break

Regarding the comment they reflect the reality of the chrome app

gre:
i'll merge this and bring an iteration function in another PR

PR #117, opened 2018-12-11
Initial port of some Manager features

gre:
(will continue later with more src/hw/* files)

### DIFF
--- a/demo/src/demos/index.js
+++ b/demo/src/demos/index.js
@@ -5,6 +5,7 @@ import Derivations from "./derivations";
 import Currencies from "./Currencies";
 import bridgestream from "./bridgestream";
 import qrstreambenchmark from "./qrstreambenchmark";
+import transports from "./transports";
 
 export default {
   CountervalueDirect,
@@ -13,5 +14,6 @@ export default {
   Derivations,
   Currencies,
   bridgestream,
-  qrstreambenchmark
+  qrstreambenchmark,
+  transports
 };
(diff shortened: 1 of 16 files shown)


demo/src/demos/index.js
```diff
@@ -13,5 +14,6 @@ export default {
   Derivations,
   Currencies,
   bridgestream,
-  qrstreambenchmark
+  qrstreambenchmark,
+  transports
```
meriadec: `comma-dangle` :cry:

demo/src/demos/transports/index.js
```diff
@@ -0,0 +1,117 @@
...
+import { setNetwork } from "@ledgerhq/live-common/lib/network";
+import genuineCheck from "@ledgerhq/live-common/lib/hw/genuineCheck";
+import getDeviceInfo from "@ledgerhq/live-common/lib/hw/getDeviceInfo";
+
+setNetwork(axios);
+
+let queue = Promise.resolve();
+const execInQueue = <T>(job: () => Promise<T>) => {
+  const p = queue.then(job);
+  queue = p;
```
meriadec: don't need to use `p` ref, but I guess it's for clarity

meriadec: (btw this is demo we dont care)

gre: it actually is required for flow last time i checked.

gre: p is of time `Promise<T>` so flow needs to remember that.

queue is a `Promise<any>` which accepts `Promise<T>`

PR #119, opened 2018-12-14
Add more hw functions and generic system for deviceAccess

gre:


src/hw/installOsuFirmware.js
```diff
@@ -0,0 +1,18 @@
...
+import ManagerAPI from "../api/Manager";
+
+export default (
+  transport: Transport<*>,
+  targetId: string | number,
+  firmware: OsuFirmware
+): Observable<*> => {
+  const params = {
+    targetId,
+    ...firmware,
```
valpinkman: this is a bit nasty as we also send all the applications array as URL argument, which can be quite long... I guess it's safer/cleaner to just explode the keys we need (targetId, firmware, firmwareKey and perso)

src/hw/firmwareUpdate-prepare.js
```diff
@@ -0,0 +1,35 @@
+// @flow
+import { Observable, from, of, empty, concat } from "rxjs";
+import { mergeMap, delay, filter, map, throttleTime } from "rxjs/operators";
+
+import getDeviceInfo from "../hw/getDeviceInfo";
+import installOsuFirmware from "../hw/installOsuFirmware";
+import { withDevice } from "../hw/deviceAccess";
+import type { OsuFirmware } from "../types/manager";
+
+const wait3s = of({ type: "wait" }).pipe(delay(3000));
```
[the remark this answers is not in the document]
gre: would probably be overkill for the localized usage we have here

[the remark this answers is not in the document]
valpinkman: fine with me :)

src/hw/installOsuFirmware.js
valpinkman: I'll take it

valpinkman: seems like it's done already

PR #121, opened 2018-12-26
Provide a fallback for accountsBtcBalance - Allows fix for LL-766 and LL-767

juan-cortes:
When we are unable to resolve a BTC value for ordering an account, think Posw, HCash, that inability breaks the ordering since we have a condition before the sorting stating `if (allRatesAvailable)`. Removing that condition causes a crash since the map in common relies on that value existing. Setting a fallback value fixes the sort again and places those accounts as if they had -1 balance.


juan-cortes:
>_I'm sorry Juan, I'm afraid I can't let you do that_
image

meriadec:
@juan-cortes should be fixed :joy:

PR #135, opened 2019-01-21
WIP: Return zero if testnet coin

Luuk2016:
Testnet coins will now return 0 instead of their normal mainnet value.


src/countervalues/index.js
```diff
@@ -127,6 +127,7 @@ function createCounterValues<State>({
 
   const getRate = (store, pair, date) => {
     if (pair.from.ticker === pair.to.ticker) return 1;
+    if (pair.from.ticker === pair.to.isTestnetFor) return 0;
```
meriadec: not what we want here. it should be:

```js
if ("isTestnetFor" in pair.from) return 0;
```

did you ensure that it work well (using `yarn link` to use it directly in live-desktop)?
also, can you rebase you PR to only have 1 single commit? thanks!

gre: or maybe even `if ("isTestnetFor" in pair.from || "isTestnetFor" in pair.to) return 0;` 🤔  

yet there is one thing i'm not so sure about the present solution, in the code here, a rate of 0 is somehow invalid and it will make the calculate / reverse function to return undefined, as in "not available". it will likely make the price not available on live-desktop including not having a total portfolio graph at all.

maybe instead, you can try to explicitly returns `0` if `rate===0` in https://github.com/LedgerHQ/ledger-live-common/blob/204ab5b3c5ce807dd175627301a77dae322b39dc/src/countervalues/index.js#L163-L175 (the undefined value is used as "not available", it might be ok to assume a value of `0` can represent "not have any countervalue")

Luuk2016: Just run `yarn link` in the ledger-live-common folder.
Then run `yarn link "@ledgerhq/live-common"`, `yarn` and `yarn start`, but for some strange reason i am getting these errors: 
afbeelding
These errors even happen when i checkout master ¯\_(ツ)_/¯

gre: ok I suspect the live-desktop have a too old version of live-common that don't yet have the @ledgerhq/errors , and we don't even have released a version yet. in meantime you can do `yarn add $FULL_PATH_TO_LEDGER_COMMON_FOLDER` , it will do like a npm install but it will copy the file so you need to run it each time (it's not like a link)

personally i'm having so many trouble with link so I never remember if it works with our stack

Luuk2016: Yep that worked, whats the easiest way to add a Bitcoin Testnet account with some assets to my Ledger Live? I dont have my Nano S with me now atm.

gre: @Luuk2016 
- in settings general, check on developer mode
- with your keyboard, type in uppercase "DEVTOOLS"
- click on the new devtools section that appeared😆, and select Bitcoin Testnet, enter a xpub

**(only to use for dev)**

meriadec: Lol that's right I forgot I dev that

Luuk2016: Well.... i tried testing the fix but without any progress.
For some reason it doesn't let me see the changes i did to the ledger-live-common library 😞

juan-cortes: Welcome to the world of testing a local forked library with yarn :trollface:

gre: Have you run yarn build before on the live-common? Yeah it's annoying

gre:
this PR is outdated. we can use it as reference for later. AFAIK this was not considered a problem because testing testnet is about testing the Ledger Live app too and it allows to test the portfolio logic. Maybe this idea could be rediscussed but I don't see it being a big problem (testnet is for dev anyway)

PR #137, opened 2019-01-24
Add metamask ethereum derivation scheme 44'/60'/0'/0/<account>

gre:


gre:
forgot about the intersection of index=0, will have to go with a new logic on desktop/mobile.

PR #146, opened 2019-02-07
NEW: Energi (NRG) support

andvgal:
Performed yarn-based static analysis and tests.

Smoke tested with local Live Desktop: it's possible to select Energi for account creation.

However, full testing with Live Desktop was not possible due to unclear inability to synchronize accounts with device even for Bitcoin before all the changes.

Ledger's explorer configuration is temporary commented out.


gre:
ping me when there is an app on manager available. thanks.

PR #148, opened 2019-02-13
LL-1006 LL-1007 Extracted partners from mobile/desktop, created new svgs

juan-cortes:
Moved all the partner urls and logos to live-common in order to maintain them in just one place. It's a bit messy still since we still need two sets of icons for react/react-native but at least we can add/remove without having to change them in both repos (well... descriptions would need to be added too I guess so 🤷‍♂️ )

Fixes https://github.com/LedgerHQ/ledger-live-desktop/pull/1807 
Fixes https://github.com/LedgerHQ/ledger-live-mobile/pull/767

image


flow/react-native.js
```diff
@@ -0,0 +1,5 @@
+// @flow
+
+declare module "react-native" {
```
gre: mmh

gre:
I think it's good enough as a first step: the fact is shared in live-common will make it more maintainable in the current way it's implemented. Moving everything to SVG component have some impact on the bundle size but I think it's way acceptable, especially it's embedded (electron / RN). The cons of SVG in the future could be that some exchanges have non vectorial images but for now it's fine.

In the future, in long term, I think the real solution will be to have an API that returns these entries and the images.

gre:
does this PR also add the new ones?

meriadec:
This PR illustration ahaha :D

juan-cortes:
@gre I went over all the icons, grounds higher resolution images for those that didn't have an svg format, vectorized them, svgomg'ed them, created the react and react native js equivalents.

Including the new ones yes sir.

gre:
cool, i'll merge this and release a new live-common

PR #152, opened 2019-02-15
Update Substratum Contract address, decimals.

substratumservices:
Substratum deployed a new [smart contract](https://substratum.net/news/substratum-smart-contract-upgrade-token-swap/). Please help me understand how to provide a correct ledgerSignature value? Is that something I can do with a Ledger Blue or is that only something the LedgerHQ team can do?


substratumservices:
Apologies for the noise. I just found the documentation where LedgerHQ sources the ERC-20 tokens from MyEtherWallet/ethereum-lists. I already submitted a pull request to update this information and it has been merged.

gre:
The ERC20 data we have here are derivated from an internal database we maintain with signature. There is not yet a clear path on how to contribute to it but we'll make this more clear in the future. (Btw that erc20 data and the provideERC20TokenInformation is not yet doing anything but it's to prepare an upcoming update of eth app that will externalize the erc20 data. That's why we have also Ledger signatures with the data)

cc @FabriceDautriat @s0lness

PR #153, opened 2019-02-17
Add TomoChain

thanhson1085:


src/data/cryptocurrencies.js
```diff
@@ -1519,6 +1519,28 @@ const cryptocurrenciesById = {
...
+    name: "TomoChain",
+    managerAppName: "TomoChain",
+    ticker: "TOMO",
+    scheme: "tomo",
+    color: "#FF9933",
+    family: "ethereum",
+    ethereumLikeInfo: {
+      chainId: 88
+    },
+    blockAvgTime: 2,
```
gre: the blockchains update every 2 seconds !?

gre:
thanks for the PR, something is odd in the blockAvgTime that is in seconds.

src/data/cryptocurrencies.js
thanhson1085: yep, create new block every 2 seconds (150 block producer - masternodes, see https://stats.tomochain.com)

gre: neat

PR #154, opened 2019-02-19
Add bridge type & Ethereum & Ripple implementations

gre:
This is a first milestone for sharing the bridge code for mobile & desktop.

mobile will be trivial to migrate, but for desktop:
- we need to migrate the existing code to the new interface (we are using an older bridge version)
  - that means removing Desktop's `bridge/types` types and adapting all code that uses bridge in the UI. we also need two different type of bridge now, the account bridge and the currency bridge. we need to do something similar to mobile: https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/bridge/index.js
- Mock can be imported from live-common and as first step we can just work with them.
- for the **Libcore Bridge**, we are not ready yet to migrate this in, libcore interface need to be unified. For now, what we need to fix in desktop is just to adapt existing code to the new bridge interface! (split into the two different bridge, etc..). *The long term plan idea would be probably to proxy the libcore calls over the renderer->internal, but maybe there is a better take on this.*
- for **ETH and XRP bridge**, we can move to the present live-common implementation. It almost works, the remaining problem is that we need bridge to run in renderer, yet all the device interaction stuff needs to exec in internal thread. one way to solve this would mean to write a Transport proxy: for now we could just proxy the open(), exchange() and close() (later on we should do the same for the listen().. but not in this scope i would say). *We need to see if we go into this rabbit hole or not, but i don't really see another way except maybe putting the device related stuff OUT of the bridge? that means creating a bridge with some sort of interface. yet we have this deviceAccess mechanism that could be of use here!*


gre:
plan before we start the desktop port:

- migrate mobile & make sure everything works.
- make sure the mock bridge correctly works too.

gre:
might have a totally different take than this PR but it will means making the bridge fully async so we can impl proxy of the bridge

PR #156, opened 2019-02-20
LL-1063 Currency termination

juan-cortes:
Added the preliminary version of changes to `CryptoCurrency` type, used to determine whether a currency is to be disabled for install, show a warning, etc. When we want to mark a crypto as terminated we need to modify `cryptocurrencies.js` and add the appropriate `terminated` field.


gre:
I don't feel we are going to maintain translations for all future disruption. I would say step 1 is to make it work with English.

I didn't really understood your point about making it optional.

In fact, in the future, we will completely remove the remote disruption for these coins. basically in term of priority this comes first: if a currency get stopped, we just track it in our data and forget about it everywhere else. so a description needs to be set.

the "remote disruption" is only to prepare future interruption, before we switch to this final mode basically.

So I would say we need both a description or URL. unless we make it acceptable that a "specific message" is not required, but I think @FabriceDautriat wants it. to be discussed more.

gre:
I've misread the "fallback on default". ok ultimately i wonder if we can NOT have description. we'll discuss.

PR #157, opened 2019-02-21
Add more events in genuineCheck

gre:
This introduce more events to go through the genuine check that tells when the UI needs to display the allow manager.
It's a breaking change for genuine check usage so we'll HODL until we have the UI ready OR some code change to adapt with this new format.


gre:
@valpinkman can you fix the checkDeviceForManager so it also returns GenuineCheckEvent ?

The next step for this rework:

- to backport the desktop code to make it work with the new type (trivial)
- to backport the changes on mobile as well without considering yet the UI changes
- to start thinking how to refactor mobile code to support is in the step system. This is probably not so trivial but there are maybe quick ways that can be found? There are basically 3 potential different rendering. We can unify the "before" and "after" into a same "spinner" for now but it worth asking for thoughts from design team.
- (bonus) we have the opportunity to have a better error when the user refuse that. That would mean introducing a new @ledgerhq/errors and throwing it in the observable in the case we have the specific status code at nonce=3

src/api/Manager.js
```diff
@@ -273,16 +276,45 @@ const API = {
...
       })
-    }).pipe(
-      last(),
-      filter(o => o.type === "result"),
-      map(o => o.payload || "")
+    }).pipe(input =>
+      Observable.create(o => {
+        let timeout;
+        let requested;
+        input.subscribe({
```
gre: We need to return the subscription to support unsubscribe

demo/src/demos/transports/index.js
```diff
@@ -142,6 +164,7 @@ class GenuineCheckButton extends Component<*, *> {
       )
     ).subscribe({
       next: result => {
+        console.log(result);
```
meriadec: i mean

src/api/Manager.js
valpinkman: Yes I saw that on documentation. I'll fix it

demo/src/demos/transports/index.js
valpinkman: it's a WIP, calm your tits ^^

src/api/Manager.js
gre: How do we use the new GitHub wip feature?

demo/src/demos/transports/index.js
valpinkman: Also, it's the demo files

PR #159, opened 2019-02-27
Add erc20 full example

gre:
Fixes LL-1089

capture d ecran 2019-02-27 a 18 07 17


juan-cortes:
:man_shrugging: too soon

gre:
@juan-cortes there is no impact on the lib itself so it's ok :p

PR #167, opened 2019-03-07
fix repair feature

valpinkman:
This PR fixes an issue where using the `Repair tool` on a broken device would block the UI as well as the device polling.

After investigating with @gre, we discovered that repair tool would use recursion and poll the device recursively. This would cause the first stream to not end and block the following executions. 

We simplified the device access within the repair function so we allowed only one device access at a time. This will free the UI in Ledger Live because we either receive a `complete` or an `error` event, which terminates the stream.


gre:
good job

PR #173, opened 2019-03-13
emit DisconnectedDeviceDuringOperation from socket.js

gre:
to be tested


gre:
ideally we'll test this first in ledger-repl

PR #177, opened 2019-03-20
Libcore in live-common

gre:
This is step 2 of the "libcore modern unification" started by https://github.com/LedgerHQ/ledger-live-mobile/pull/819

**Here is the rough roadmap:**

1. ✅ ~~migrate mobile code to a unified API and platform independent codebase.~~
2.  ✅ ~~Move the code in live-common and only keep the specific parts (it's already isolated in libcore/react-native folder here)~~
3. 👋 **WE ARE HERE** Make a PoC from that live-common with the node libcore bindings to establish the minimal node code required and quickly check that all libcore functions are working in node.js bindings context. 
4. migrate desktop code carefully and any backport any desktop innovation that was not yet in live-common libcore (for instance, the getVersion, and some extra stuff we need)

**What was done**

- completely integrated the mobile libcore generic part in live-common
- backport some missing logic that was in desktop's sign and broadcast (for komodo)
- add a "test" project that allows to quickly test the implementation on top of node.js wrapping (that node.js wrapping will get re-used in desktop)

**How to test**

```
cd test && yarn && yarn build
```

and then you can try any of the scripts in `lib/examples`, for instance

```
node lib/examples/isValidRecipient.js
```


gre:
**Future notes**

- signAndBroadcast need to be a simpler api `(Account, Transaction, devicePath)=>..` . Transaction also will need to evolve because it will be different on different families of coin
- komodo have a todo comment on that we should use `getLockTime()` , check with libcore team
- implement ethereum codes. We need limit the number of branching and not doing ifs everywhere

PR #179, opened 2019-03-22
use a safer getAddress rather than getWalletPublicKey

valpinkman:


src/libcore/scanAccountsOnDevice.js
```diff
@@ -48,11 +49,11 @@ export const scanAccountsOnDevice = (
             : currency;
           const path = `${isSegwit ? "49" : "44"}'/${coinType}'`;
 
-          const hwApp = new Btc(transport);
-          const { publicKey: seedIdentifier } = await hwApp.getWalletPublicKey(
-            path,
-            false,
-            isSegwit
+          // const hwApp = new Btc(transport);
```
gre: can you remove this line?

PR #187, opened 2019-03-27
Add missing libcore methods

gre:
- add methods that are used in desktop but not exposed in the types (freshResetAll() and get version funcs)
- refine the signAndBroadcast contract so it can work on desktop
- add `libcoreJobBusy` to know if libcore have pending jobs or not.

### DIFF
--- a/package.json
+++ b/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@ledgerhq/live-common",
-  "version": "4.36.1",
+  "version": "4.37.0-beta.0",
   "main": "lib/index.js",
   "license": "MIT",
   "scripts": {
--- a/src/libcore/access.js
+++ b/src/libcore/access.js
@@ -1,5 +1,7 @@
 // @flow
 import type { Core } from "./types";
+import { Subject, Observable } from "rxjs";
+import { map, distinctUntilChanged } from "rxjs/operators";
 
 const GC_DELAY = 1000;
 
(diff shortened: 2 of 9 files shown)


gre:
```js

  function trackAllCalls() {
    /* eslint-disable */
    const logRes = res => {
      if (res && typeof res === 'object' && res.then) {
        res.then(logRes, () => {})
      } else {
        if (res !== undefined) console.log('=>', res)
      }
    }
    const decorateStaticFunc = (k, m) => {
      const module = statics[k]
      const method = module[m]
      module[m] = function(...args) {
        console.log(`core.${k}.${m}(`, ...args, ')')
        const res = method.apply(this, args)
        logRes(res)
        return res
      }
    }
    const decorateMethod = (k, m) => {
      const module = statics[k]
      const method = module.prototype[m]
      module.prototype[m] = function(...args) {
        console.log(`core.${k}#${m}(`, ...args, ')')
        const res = method.apply(this, args)
        logRes(res)
        return res
      }
    }
    Object.keys(statics).forEach(k => {
      const module = statics[k]
      Object.keys(module).forEach(m => {
        const method = module[m]
        if (typeof method === 'function') {
          decorateStaticFunc(k, m)
        }
      })
      Object.keys(module.prototype).forEach(m => {
        if (k === 'Runnable' && m === 'run') return
        const method = module.prototype[m]
        if (typeof method === 'function') {
          decorateMethod(k, m)
        }
      })
    })
    decorateStaticFunc('ExtendedKeyAccountCreationInfo', 'init')
    /* eslint-enable */
  }

  if (__DEV__) trackAllCalls()
```

helper to investigate libcore calls

PR #189, opened 2019-03-28
[Major] rework of derivation scheme, change getAddress signature, libcore functions to prepare eth/erc20, prepare native segwit too

gre:
(see the commit message for detail)


gre:
this PR will be important but ideally we'll first have https://github.com/LedgerHQ/ledger-live-common/pull/184 merged, so we can then drop more deprecated stuff.

gre:
also need to see if we want to bump some lib or not (look at changelogs)

src/derivation.js
```diff
@@ -161,12 +235,22 @@ export const getDerivationModesForCurrency = (
       all.push("segwit_unsplit");
     }
   }
+  if (currency.supportsSegwit) {
+    all.push("segwit_on_legacy");
+    all.push("legacy_on_segwit");
+  }
   all.push("");
   if (currency.supportsSegwit) {
```
valpinkman: same confition as above, is this on purpose ?

src/libcore/platforms/react-native.js
```diff
@@ -0,0 +1,211 @@
...
+
+    const mappings: Object = {};
+
+    const wrappers = {
+      // FIXME need to fix it in RN bindings
+      hex: value => value.replace(/[< >]/g, "")
+    };
+
+    function wrap(id, ref) {
+      if (!ref) return ref;
```
valpinkman: how can you/why do you return `ref` is there is no `ref` ?

src/libcore/platforms/react-native.js
```diff
@@ -0,0 +1,211 @@
...
+      const Clz = mappings[id];
+      if (!Clz) {
+        console.warn(`Unkown wrapping '${id}'`);
+        return ref;
+      }
+      return new Clz(ref);
+    }
+
+    function unwrap(id, instance, ctx) {
+      if (!instance) return instance;
```
valpinkman: same here ? why do you return it if it doesn't exists/ is falsy ?

src/libcore/platforms/react-native.js
gre: some method returns optional results. For instance `CoreOperation#getFees(): ?CoreAmount;`

in that case if it's a `null` we need to return a `null`

if it's a `ref` (like a `{ id: "uuid of a CoreAmount" }`) it needs to wrap it by a `new CoreAmount(ref)`

src/derivation.js
gre: yes because there is a `push("")` between the 2.

the order makes that we scan invalid paths first, then the "normal" `44'` accounts, then the legacy `49'` accounts

src/libcore/platforms/react-native.js
gre: same yes.

in that case it is to address optional parameters.

gre:
@meriadec for you to acknowledge. you can see that we've upgraded BigNumber. I imagine that's the only part that might affect you but i guess you were already on the 8.x !

gre:
the PR pass ci on https://circleci.com/gh/gre/ledger-live-common/tree/rework-derivationMode

src/derivation.js
valpinkman: 👍

src/libcore/platforms/react-native.js
valpinkman: 👍

PR #191, opened 2019-03-31
feat: Allow for dynamic ledger explorer switching for cryptos

juan-cortes:
> I was unable to test it because demo project was doing a recursive sync loop and it's sunday so 🤷‍♂️ 

- Removes ledgerExplorerId and ledgerExplorerVersion from the data/cryptocurrencies
- Adds a resolver to fetch the explorer id and version for cryptos supporting it
- Allows experimental explorer + specific experimentals through the use of `getEnv` and `setEnv`

The explorer is resolved something like this:
If the `USE_EXPERIMENTAL_EXPLORERS` flag is true, it will check if we also have a specific override using the `CRYPTOIDINCAPS_EXPLORER` env variable, defaulting to the experimental one if not. It will always attempt to fallback to default anyway.

This needs to be plugged in into places that rely on the v2/v3 logic such as `api/Ethereum` but the change would be simple enough, from `currency.ledgerExplorerVersion === "v2"` to `getCurrencyExplorer(currency).id`


gre:
Great!!

api/Ethereum is in live-common no?

juan-cortes:
Yeah but I wanted another set of eyes before I went over all the changes, and probably some help testing it, I can't seem to get wml to work with demo being inside common and having common as a dependency, it chokes

PR #195, opened 2019-04-03
WIP libcore encryption

gre:
**To test things**

```
yarn
yarn build
cd test
yarn
yarn build
```

each time you want to reset libcore state:

```
rm -rf dbdata
```

**Status**

- ✅  I am able to scan accounts with a password. `LIBCORE_PASSWORD=toto node lib/examples/scanAccountsOnDevice.js` **works**. 
- ✅  I am able to scan AGAIN accounts with same password. `LIBCORE_PASSWORD=toto node lib/examples/scanAccountsOnDevice.js` **works again**. 
- 🔴  I am able to change the password. `LIBCORE_PASSWORD=toto node lib/examples/changePassword.js "toto" "foo"` => `segmentation fault`

**to be done on desktop**

- each time user input password to unlock the app. we hash this password in memory and use this hash to encrypt libcore by doing `setEnv("LIBCORE_PASSWORD", hash)`. With the new system we're working on with @valpinkman, this will reboot the internal process each time we change envs.
- we will add a new command `changeLibcorePassword(newPassword)` that, when call will basically do `libcore.getPool().changePassword`. when implementing the change password UI, we hash the new password to call that command with it. We wait it is successful and only then we can apply the encryption over user accounts data as well and only then we do a `setEnv("LIBCORE_PASSWORD", hash)`.
- as the previous work is not guarantee to be atomic, we need a way to reset the libcore when UI password matches but libcore password don't! It sounds like we need to rimraf sqlite like before.
- same goes when we "remove" the password. we will need to call `changeLibcorePassword("")` awaited and followed by a `setEnv("LIBCORE_PASSWORD","")`

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -29,7 +29,8 @@ const envParsers = {
   BASE_SOCKET_URL: stringParser,
   EXPLORER_V2: stringParser,
   EXPLORER_V3: stringParser,
-  EXPERIMENTAL_EXPLORERS: boolParser
+  EXPERIMENTAL_EXPLORERS: boolParser,
+  LIBCORE_PASSWORD: stringParser
 };
 
 // This define the default values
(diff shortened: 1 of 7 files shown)


gre:
i'll merge the work in #189

PR #198, opened 2019-04-04
Preparing for v2->v3 transition

gre:
we're just going to move ALL coins to v3, not just ethereum! so we have a smooth plan to do this by overriding EXPERIMENTAL_EXPLORERS to test it as soon as it's ready.

> HODL the PR because we'll need confirm (hopefully this week) that this will be effectively the new URL. maybe https://explorers.api.live.ledger.com/blockchain/v3/$ledgerExplorerId ? 
we need to also confirm that the new ticker are identical OR not, and which one will change. I've identified `eth` to change to `ethereum-ropsten` ?

### DIFF
--- a/src/explorers.js
+++ b/src/explorers.js
@@ -32,187 +32,67 @@ export const getAccountOperationExplorer = (
     operation.hash
   );
 
-export const hasCurrencyExplorer = (currency: CryptoCurrency): boolean => {
-  return !!ledgerExplorers[currency.id];
+type LedgerExplorer = {
+  version: string,
+  id: string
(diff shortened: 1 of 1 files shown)


gre:
HODL the PR because we'll need confirm (hopefully this week) that this will be effectively the new URL.
we need to also confirm that the new ticker are identical OR not, and which one will change. I've identified `eth` to change to `ethereum-ropsten` ?

gre:
merged inside https://github.com/LedgerHQ/ledger-live-common/pull/189

PR #205, opened 2019-04-17
feat: LL-1264Testing for mock countervalue

juan-cortes:
This allows us to provide a custom `getDailyRatesImplementation`, `fetchExchangesForPairImplementation`, `fetchTickersByMarketcapImplementation`. They are present inside `countervalues/mock` that will fake the rates for a given pair and provide the rates for the past days too, accounting for a faked inflation for variance but keeping it consistent to allow Waldo or any other test suite to work with this.

I'll link a pr on mobile and desktop that will use the mock instead of the real ones.

### DIFF
--- a/src/countervalues/mock.js
+++ b/src/countervalues/mock.js
@@ -116,10 +116,16 @@ export const getDailyRatesImplementation = (
   }, rates);
 };
 
+/**
+ * [good]: Will return deterministic rates for the past 3 years for the pair
+ * [ugly]: Will respond, but return 0 as the latest rate.
+ * [bad]: Will not return anything for the pair.
+ */
 export const fetchExchangesForPairImplementation = async () => [
   { id: "good", name: "good", website: "#" },
   { id: "ugly", name: "ugly", website: "#" },
   { id: "bad", name: "bad", website: "#" }
 ];
-export const fetchTickersByMarketcapImplementation = async ():Promise<string[]> =>
-  listCryptoCurrencies().map(c => c.ticker);
+export const fetchTickersByMarketcapImplementation = async (): Promise<
+  string[]
+> => listCryptoCurrencies().map(c => c.ticker);
--- a/src/countervalues/mock.js
+++ b/src/countervalues/mock.js
@@ -1,7 +1,8 @@
 // @flow
-import type { CounterValuesState, Histodays } from "./types";
+import type { CounterValuesState, Histodays, PollAPIPair } from "./types";
 import type { Currency } from "../types";
 import { formatCounterValueDay } from ".";
+import { listCryptoCurrencies } from "../currencies";
 
 type Pair = {
   from: Currency,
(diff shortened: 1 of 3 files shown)


src/countervalues/mock.js
```diff
@@ -32,3 +33,93 @@ export const genStoreState = (pairs: Pair[]): CounterValuesState => {
...
+export const getDailyRatesImplementation = (
+  getAPIBaseURL: () => string,
+  pairs: PollAPIPair[]
+) => {
+  const rates = {
+    USD: { BTC: { SIMEX: { latest: 0.00521587214628 } } },
+    BTC: {}
+  };
+
+  // For pairs not reflected here we will be falling back on 1  ¯\_(ツ)_/¯
```
gre: I would suggest instead to fallback to no countervalues? because we want to test the "partial" aspect & we'll have it when doing erc20 ;)

for the record, this is a script that i've used to generate that list: https://github.com/LedgerHQ/ledger-live-common/blob/master/scripts/listMarketPriceFor1USD.js we might adapt in future to add more tokens ;) i think we have listTokens() already

src/countervalues/mock.js
```diff
@@ -32,3 +33,93 @@ export const genStoreState = (pairs: Pair[]): CounterValuesState => {
...
+    }
+    return rates;
+  }, rates);
+};
+
+export const fetchExchangesForPairImplementation = async () => [
+  { id: "good", name: "good", website: "#" },
+  { id: "ugly", name: "ugly", website: "#" },
+  { id: "bad", name: "bad", website: "#" }
+];
```
gre: please quickly document the behavior of each exchange

src/countervalues/types.js
```diff
@@ -93,7 +93,9 @@ export type Input<State> = {
   getDailyRatesImplementation?: (
     getAPIBaseURL: () => string,
     pairs: PollAPIPair[]
-  ) => Promise<mixed>
+  ) => Promise<mixed>,
+  fetchExchangesForPairImplementation?: ()=>  Promise<Exchange[]>,
+  fetchTickersByMarketcapImplementation?: ()=> Promise<string[]>
```
gre: `prettier`

src/countervalues/mock.js
juan-cortes: I'll look into it later, looks good

src/countervalues/mock.js
juan-cortes: makes sense

src/countervalues/mock.js
juan-cortes: ok, zero it is

gre: ok thanks! i'll later re-generate that rates list and probably having it located at a same place & a script to refresh it . but not blocking this PR , we'll only need it for ERC20 mocking

PR #216, opened 2019-04-27
Tests/send

Arnaud97234:
### Type
Tests

### Context
Automated test cases using both ledger-live cli and a Ledger Nano
- BTC coins synchronisation
- Send tBTC (Segwit/Legacy)

 ### Known issues
For some reason, legacy to segwit transactions outputs are not recorded.


gre:
invalid paths are only scanned with a special environment enabled

gre:
Mmh why is there all these commits

meriadec:
rebase ftw

gre:
i've sent #222 that contains the current work. I think we can close this PR and you can either starts from mine and/or create another one if you think it's not yet finished.

It seems you will need to record again some APDUs btw. I kinda feel it will be better to split one test per currency btw because as soon as you receive a coin on one account, it would fuck up this big "BitcoinsAccounts" test.

PR #227, opened 2019-05-02
Tests Sync

Arnaud97234:
Refacto sync testing. One test by currency instead of a global sync test, should be esy to debug and maintains in case of needs to improve testing data.


gre:
CircleCi not running?

gre:
@Arnaud97234  https://circleci.com/gh/arnaud97234/ledger-live-common

PR #232, opened 2019-05-03
Force using own build in tests

MortalKastor:
Tests were using `ledger-live` from `$PATH` rather than the current build.
This should fix that.


gre:
it doesn't seem to work @MortalKastor

PR #237, opened 2019-05-07
Provide solution to reorder erc20 accounts + getting portfolio distribution

gre:
## reorderTokenAccounts

default implementation is to order by balance countervalues and then by ticker name

- this is done by implementation of reorderTokenAccounts utility
- the libcore sync will preserve the order present in the provided "existing account"
- overall big picture of what will happen:
  - a first ETH synchronisation will retrieve the ERC20 token accounts
  - from thet erc20 token accounts, we determine the ERC20 used by the app, this is provided to countervalues api to fetch only the necessary, so by design if there are new erc20, their rates are fetched. 
  - at key times (typically after fetching countervalues / at open of portfolio) we trigger a reordering that will potentially reupdate the account objects (reorderTokenAccounts)

## `getAssetsDistribution(accounts, calculateCountervalue)`

returns a AssetsDistribution

```js
export type AssetsDistribution = {
  // if not available, we would not display anything
  isAvailable: boolean,
  // a sorted list of assets with data
  list: Array<{
    currency: CryptoCurrency | TokenCurrency,
    distribution: number, // % of the total (normalized in 0-1)
    amount: BigNumber,
    countervalue: BigNumber // countervalue of the amount that was calculated based of the rate provided
  }>,
  // number of accounts to show first (before the see all)
  showFirst: number
};
```

example

```js
getAssetsDistribution(
    Array(40)
      .fill(null)
      .map((_, i) => genAccount("gad1_" + i)),
    (currency, value) => {
      const rate = baseMockBTCRates[currency.ticker];
      if (rate) return value.times(rate);
    }
  )

{ isAvailable: true,
        showFirst: 6,
        list:
         [ { currency: '[Currency XRP]',
             amount: 'BigNumber(46382020622)',
             countervalue: 'BigNumber(287939584.021376)',
             distribution: 0.20732887782734052 },
           { currency: '[Currency ETC]',
             amount: 'BigNumber(1.758879989230665628497e+21)',
             countervalue: 'BigNumber(211593262.7044490751081891)',
             distribution: 0.1523562446665281 },
           { currency: '[Currency LTC]',
             amount: 'BigNumber(13518366243)',
             countervalue: 'BigNumber(210927068.489529)',
             distribution: 0.15187655619485138 },
           { currency: '[Currency DOGE]',
             amount: 'BigNumber(277666122448926)',
             countervalue: 'BigNumber(149717573.2244608992)',
             distribution: 0.10780308846093133 },
           { currency: '[Currency ETH]',
             amount: 'BigNumber(44768186839198358530)',
             countervalue: 'BigNumber(143182091.967808110086499)',
             distribution: 0.10309726102282941 },
           { currency: '[Currency PPC]',
             amount: 'BigNumber(10789694608)',
             countervalue: 'BigNumber(121233008.615488)',
             distribution: 0.08729297750883552 },
           { currency: '[Currency BTG]',
             amount: 'BigNumber(28269664213)',
             countervalue: 'BigNumber(91565442.385907)',
             distribution: 0.06593105453755452 },
           { currency: '[Currency ZEC]',
             amount: 'BigNumber(5959921817)',
             countervalue: 'BigNumber(79585219.9911278)',
             distribution: 0.057304779433096656 },
           { currency: '[Currency DASH]',
             amount: 'BigNumber(2391885926)',
             countervalue: 'BigNumber(55362926.50722564)',
             distribution: 0.03986368690846095 },
           { currency: '[Currency BCH]',
             amount: 'BigNumber(617826871)',
             countervalue: 'BigNumber(37699795.66842)',
             distribution: 0.02714547343957158 },
           { currency: '[Currency BTC]',
             amount: 'BigNumber(515021410)',
             countervalue: 'BigNumber(0)',
             distribution: 0 } ] }
```


gre:
don't merge yet.

this need more thoughts: i don't think reordering can happen at `syncAccount` time, because well, it's a 🐔 🥚  problem: we need to know the ERC20 tokens of eth accounts to load the rates first, but we can't do that until syncing, so we can't order at sync time...

Here are some ideas:

- the reference "existingAccount" object is the source of trust on what ordering is used. that would mean that the tokenAccounts shouldn't follow the libcore ordering [here](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/ethereum/libcore-buildTokenAccounts.js#L71) but instead we need to preserve the one in `existingAccount.tokenAccounts` if it exists and then append the rest.
- a reordering can happen a posteriori, typically after rates are reloaded but maybe not always.. the present method is great because we can make it purely on renderer side and it's ok to run it each time because we can make that account ref is preserved if order doesn't even change (need to implement a top level method `reorderAccountByCountervalues` that delegate to `reorderTokenAccountsByCountervalues` and preserve account ref).. we could do that at same time as doing the reorderAccounts that already exists in live-desktop/mobile.

gre:
Done.

PR #248, opened 2019-05-20
Fix XST: take into consideration XST v2 txs without timestamp

KhalilBellakrid:
Related to : https://ledgerhq.atlassian.net/browse/LLC-214


gre:
@KhalilBellakrid nice. Is that the only remaining thing to fix? Would it work with the last stable libcore version or would it require a new one?

KhalilBellakrid:
There is also a PR on `ledgerjs` and yes it would work with last stable version of libcore !

gre:
@KhalilBellakrid  you can update ledgerjs hw app btc:

in top level of live-common:

```
yarn add @ledgerhq/hw-app-btc@4.58.1-alpha.1
```

and commit :)


then, I'm falling into an error 0x6f01. More detail in the LLC-214

PR #264, opened 2019-06-05
Fixes QTUM to support segwit

gre:
this can only be merged when we have confirmation to have a new app that supports segwit.
in complement, we could look at having a backward compatibility or more explicit error when an app is asked to derive on segwit but it fails.


gre:
i'll provide another fix that is non breaking change

PR #271, opened 2019-06-12
Eth should be #1c1ce1

juan-cortes:
¯\\_(ツ)_/¯


gre:
i would say we won't change anything not until https://api.ledgerwallet.com/update/assets/icons/ethereum matches that. 
easy to reopen the PR if this happens.

PR #274, opened 2019-06-13
Fixes account discovery for the XRP bridge.

juan-cortes:
Updates the XRP bridge to the current rippled API

### DIFF
--- a/src/countervalues/index.js
+++ b/src/countervalues/index.js
@@ -598,7 +598,9 @@ function createCounterValues<State>({
       if (prevProps.pairsKey !== this.props.pairsKey) {
         if (log)
           log(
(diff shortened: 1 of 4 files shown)


src/countervalues/index.js
```diff
@@ -598,7 +598,9 @@ function createCounterValues<State>({
       if (prevProps.pairsKey !== this.props.pairsKey) {
         if (log)
           log(
-            `pairsKey changed:\n    ${prevProps.pairsKey}\n => ${this.props.pairsKey}`
+            `pairsKey changed:\n    ${prevProps.pairsKey}\n => ${
+              this.props.pairsKey
```
gre: I suspect you have diff prettier version? can you check if `yarn prettier` don't re-set this to original? @juan-cortes

juan-cortes:
This will not be needed since we will revert to a previous version of rippled in the meantime.

PR #276, opened 2019-06-18
adds cryptocurrency to appversion when available

valpinkman:


gre:
forget about our API being drunk. i'll later make sure our test logs everything in VERBOSE so we see more clear what is wrong. #229

PR #278, opened 2019-06-19
Added getTransactionStatus

juan-cortes:
Refactored some logic out of live-desktop into here to be reused on mobile.

### DIFF
--- a/src/transaction.js
+++ b/src/transaction.js
@@ -1,0 +1,53 @@
+// @flow
+import type { BigNumber } from "bignumber.js";
+import type { Account, Transaction } from "./types";
+import type { AccountBridge } from "./bridge/types";
+
+export async function getTransactionStatus(
+  bridge: AccountBridge<any>,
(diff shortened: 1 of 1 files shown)


src/transaction.js
```diff
@@ -0,0 +1,53 @@
...
+  const useAllAmount = bridge.getTransactionExtra(
+    account,
+    transaction,
+    "useAllAmount"
+  );
+
+  let maxAmount;
+  if (useAllAmount) maxAmount = await bridge.getMaxAmount(account, transaction);
+
+  const estimatedFees = totalSpent - (maxAmount || amount);
```
gre: unfortunately this is not true for ERC20.

the fees is not `totalSpent minus amount` because this is basically zero for erc20.

sounds like we need a bridge.getEstimatedFees. (or we need this whole function to be in bridge)

gre:
404 on the ci ..

PR #60, opened 2018-08-08
Update Bitcoin Cash scheme from `bch` to `bitcoincash`

meriadec:
Following [Bitcoin Cash cashaddr specs](https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/cashaddr.md#prefix), the prefix for the currency is `bitcoincash`

PR #286, opened 2019-06-20
check for cashAdrr in isValidRecipient

valpinkman:


src/libcore/isValidRecipient.js
```diff
@@ -17,6 +17,11 @@ export const isValidRecipient: F = withLibcoreF(core => async arg => {
     return res;
   }
   const { currency, recipient } = arg;
+
+  if (recipient.startsWith("bitcoincash:")) {
```
MortalKastor: IMHO it's not enough to only test for the `bitcoincash:` prefix, especially since  it collides with the URI scheme used in QR Codes (see #60 ):
- A cashaddr QR formatted as `bitcoincash:<addr>` rather than `bitcoincash:bitcoincash:<addr>` would not be caught by this test
- Wallets or websites omitting the `bitcoincash:` prefix or users only copying the last part of the address would not be caught either
- On the other hand, implementations displaying a supported bitcoin addresses with the `bitcoincash:` URI scheme would trigger a _not supported_ error

gre: so technically it would still catch if you manually copy paste in that field. only the QR code scanning can have a problem.

I kinda wonder if we should just make the qrcode scanner preserving `bitcoincash:*` as the recipient, meaning we will not support `bitcoincash:*` as a "uri scheme" qrcode bit will assume it's an address part already

MortalKastor: Regardless of the QR issue ([the official specs require to not have a double prefix in this context](https://github.com/bitcoincashorg/bitcoincash.org/blob/master/spec/cashaddr.md#double-prefix) btw), using CashAddr without prefix is not uncommon and should be handled: Bitcoin ABC own CashAddr announcement states:
> Note the prefix “bitcoincash:”, which is technically always part of the address, although the prefix may be optional or missing entirely, depending on the wallet or the implementation.

Why not use the reference `bitcoincashjs` lib to test the address format?
https://github.com/bitcoincashjs/bchaddrjs#test-for-address-format

gre: it's more libcore that should provide a way to do that.

gre:
for internal discussion

PR #288, opened 2019-06-21
Experimental ROI calculation

gre:
- share in live-common how we calculate the portfolio % and delta value.
- introduce a new implementation (EXPERIMENTAL_ROI_CALCULATION) that 
  - implements `delta = sum( operationValue * (currentRate - operationTimeRate) )`
  - percentage ROI is `delta / sum( all receive operation countervalue at operation time )`
  - in case there are no operation in the range, the ROI is calculated with comparing market value from beginning to end of the range. For portfolio it's `delta / (beginRangePortfolioSum + sum(send countervalue))`


.gitignore
```diff
@@ -1,5 +1,7 @@
 # See https://help.github.com/ignore-files/ for more about ignoring files.
 
+.yalc
```
meriadec: good candidate for `~/.gitignore_global` :slightly_smiling_face:

gre: i think we'll leave it in .gitignore in fact, because yalc is recommended in our README and is also used in the CI and not ignoring them will create problem, typically with checking there is no git diff / using with jest.

PR #291, opened 2019-06-25
V7

gre:
- requires to use a lib-ledger-core binding version 3.x (new libcore major => new live-common major)
- Removed `sortAccounts, reorderAccountByCountervalues, reorderTokenAccountsByCountervalues` (see equivalent in ordering.js)
- Have moved to **lib/api/Ledger**: `findCurrencyExplorer, hasCurrencyExplorer, getCurrencyExplorer`
- NotEnoughGas is thrown when sending ERC20 but not enough ETH.
- Add `sortByMarketcap`, `getMarketcapTickers`, `useMarketcapTickers`, `currenciesByMarketcap`, `useCurrenciesByMarketcap`
- CSV operations to return erc20 account operations
- Replace `createCountervalues` by `{ implementCountervalues, getCountervalues }`
- Introduce `setSupportedCurrencies`, `listSupportedCurrencies`, `isCurrencySupported`
- drop EXPERIMENTAL_NATIVE_SEGWIT : it's now stable if you use libcore @ 3.

### Change signature of countervalues to prepare future

The point is to have countervalues living in live-common more and more so we can more easily migrate codes in live-common. Typically we can getCountervalues() from inside live-common logic.

Replace code like this:

```js
import createCountervalues from "@ledgerhq/live-common/lib/countervalues";
const Countervalues = createCountervalues(a);
```

by

```js
import { implementCountervalues, getCountervalues } from "@ledgerhq/live-common/lib/countervalues";
implementCountervalues(a);
const Countervalues = getCountervalues();
```


demo/package.json
```diff
@@ -3,19 +3,19 @@
...
-    "@ledgerhq/hw-transport-u2f": "^4.46.0",
-    "@ledgerhq/hw-transport-web-ble": "^4.46.0",
-    "@ledgerhq/hw-transport-webusb": "^4.46.0",
-    "@ledgerhq/live-common": "file:..",
-    "axios": "^0.18.0",
+    "@ledgerhq/hw-transport": "^4.63.2",
+    "@ledgerhq/hw-transport-u2f": "^4.63.2",
+    "@ledgerhq/hw-transport-web-ble": "^4.63.2",
+    "@ledgerhq/hw-transport-webusb": "^4.63.2",
+    "@ledgerhq/live-common": "^7.0.0-beta.0",
```
juan-cortes: Why does it depend on itself?

gre: oops

gre: `demo/` depends on live-common

PR #293, opened 2019-07-02
Add Qtum Segwit derivation

Arnaud97234:


gre:
it's a dup of #264 right?
as this requires a new app but it's also a breaking change, i'm not sure how we should proceed.

gre:
i'll provide a fix that is non breaking change

PR #294, opened 2019-07-03
Ability to hide token accounts with balance 0

MortalKastor:
Adds `HIDE_EMPTY_TOKEN_ACCOUNTS` environment variable to remove token accounts with zero balance from sync results

### Context

https://ledgerhq.atlassian.net/browse/LL-1440
https://ledgerhq.atlassian.net/browse/LL-1477


src/libcore/syncAccount.js
```diff
@@ -90,6 +96,20 @@ export async function syncCoreAccount({
   return account;
 }
 
+function hideEmptyTokenAccounts(tokenAccounts: TokenAccount[]): TokenAccount[] {
```
gre: if you don't mind, i would prefer this logic to live in the core account iteration https://github.com/LedgerHQ/ledger-live-common/blob/2ea7e1105cbd9e6fd65368ad4ba64e2faa4be554/src/families/ethereum/libcore-buildTokenAccounts.js#L76

you can notice there is a function called buildERC20TokenAccount, we could even `return null` here https://github.com/LedgerHQ/ledger-live-common/blob/2ea7e1105cbd9e6fd65368ad4ba64e2faa4be554/src/families/ethereum/libcore-buildTokenAccounts.js#L20-L22 to indicate that "we skip this account", which is then just about doing `if(account)accounts.push(account)` there

wdyt?

MortalKastor: I totally missed that part, I assumed the token accounts array was coming straight from the libcore binary 😅
I'm refactoring right away.

gre:
hold on. we'll merge v7 first.

PR #304, opened 2019-07-17
LL-1681 Throw a AccountNotSupported error if derivation mode is not supported

juan-cortes:
When trying to get an account's bridge, we will now throw an `AccountNotSupported` error if the derivation mode is not supported. This pr also drops ethW1, ethW2, rip and rip2 derivation modes.

### Type

Feature

### Context

https://ledgerhq.atlassian.net/browse/LL-1681


gre:
@juan-cortes can you rebase with just your commit? (+fix flow) thx

PR #307, opened 2019-07-22
LL-1549 Add checkCurrencyApp and checkAccountApp commands

juan-cortes:
Adds two commands to diferentiate between the open app and the correct device logic. Allowing us to see if the current open app is the one we are expecting. I'll need to ping @dasilvarosa to get the wording because obviously we can't go with what's in the screenshot :trollface: but for that, it's another pr on desktop.

Screenshot 2019-07-22 at 19 17 09

### DIFF
--- a/src/hw/checkAccountApp.js
+++ b/src/hw/checkAccountApp.js
@@ -1,0 +1,30 @@
+// @flow
+import Transport from "@ledgerhq/hw-transport";
+import { WrongDeviceForAccount } from "@ledgerhq/errors";
+import getAddress from "./getAddress";
+import checkCurrencyApp from "./checkCurrencyApp";
+import type { Account, CryptoCurrency } from "../types";
+import { isSegwitDerivationMode } from "../derivation";
+
+export default async (
+  transport: Transport<*>,
+  account: Account,
+  currency: CryptoCurrency,
+  devicePath: string
+): Promise<void> => {
+  await checkCurrencyApp(transport, currency, devicePath);
+  const { address } = await getAddress(transport, {
+    derivationMode: account.derivationMode,
+    devicePath,
+    currency,
+    path: account.freshAddressPath,
+    segwit: isSegwitDerivationMode(account.derivationMode)
+  });
+
+  const { freshAddress } = account;
+  if (freshAddress !== address) {
+    throw new WrongDeviceForAccount(`WrongDeviceForAccount ${account.name}`, {
+      accountName: account.name
+    });
+  }
+};
--- a/src/hw/checkCurrencyApp.js
+++ b/src/hw/checkCurrencyApp.js
@@ -1,0 +1,47 @@
+// @flow
+import Transport from "@ledgerhq/hw-transport";
+import getAppAndVersion from "./getAppAndVersion";
+import {
+  DeviceOnDashboardUnexpected,
+  WrongAppForCurrency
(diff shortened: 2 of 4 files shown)


src/hw/checkAccountApp.js
```diff
@@ -0,0 +1,39 @@
...
+
+export default async (
+  transport: Transport<*>,
+  account: Account,
+  currency: CryptoCurrency,
+  devicePath: string
+): Promise<void> => {
+  await checkCurrencyApp(transport, currency);
+  const { address } = await getAddress(transport, {
+    derivationMode: account ? account.derivationMode : "",
```
gre: mmh, according to the type of the function, it can't be falsy 🤔

src/hw/checkAccountApp.js
```diff
@@ -0,0 +1,39 @@
...
+    devicePath,
+    currency,
+    path: account
```
gre: same

src/hw/checkAccountApp.js
```diff
@@ -0,0 +1,39 @@
...
+      ? account.freshAddressPath
+      : runDerivationScheme(
+          getDerivationScheme({ currency, derivationMode: "" }),
+          currency
+        ),
+    segwit: account ? isSegwitDerivationMode(account.derivationMode) : false
```
gre: same

src/hw/checkAccountApp.js
```diff
@@ -0,0 +1,39 @@
...
+  });
+
+  const { freshAddress } = account;
+  if (account && freshAddress !== address) {
```
gre: same

src/hw/checkCurrencyApp.js
```diff
@@ -0,0 +1,33 @@
...
+  const currentApp = await getAppAndVersion(transport).catch(e => {
+    if (e.status === 0x6e00) return null;
+    throw e;
+  });
+
+  if (currentApp) {
+    if (currentApp.name === currency.managerAppName) {
+      return;
+    } else if (
+      // nano x 1.2.4-1 dashboard app name
```
gre: if possible, put the array in a global and comment only in the OLOS line, for more clarity

src/hw/checkCurrencyApp.js
```diff
@@ -0,0 +1,33 @@
...
+      ["BOLOS", "OLOS\u0000"].includes(currentApp.name)
+    ) {
+      throw new DeviceOnDashboardUnexpected();
+    }
+    throw new WrongAppForCurrency(
+      `wrong app ${currentApp.name}, expected ${currency.managerAppName}`,
+      { current: currentApp.name, expected: currency.managerAppName }
+    );
+  }
```
gre: as a else case, we should fallback on just checking a getAddress works


```
await getAddress(transport, {
  derivationName: "",
  devicePath,
  currency,
  path: runDerivation....(),
  segwit: false,
})
```

src/hw/checkCurrencyApp.js
juan-cortes: gotcha

src/hw/checkCurrencyApp.js
juan-cortes: :ah:

gre:
v3 eth staging seems to have issue with 504 which is why CI is failing, we'll assume it works

PR #311, opened 2019-07-25
Flag for enforcing following HIDE_EMPTY_TOKEN_ACCOUNTS in flattenAccounts()

MortalKastor:
Keeps the current behavior of ignoring `HIDE_EMPTY_TOKEN_ACCOUNTS` as the default.
But is this what we want, though? 🤔

### DIFF
--- a/src/account/helpers.js
+++ b/src/account/helpers.js
@@ -65,14 +65,17 @@ export function listTokenAccounts(account: Account): TokenAccount[] {
 }
 
 export function flattenAccounts(
-  topAccounts: Account[] | TokenAccount[] | (Account | TokenAccount)[]
+  topAccounts: Account[] | TokenAccount[] | (Account | TokenAccount)[],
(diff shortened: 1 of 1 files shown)


src/account/helpers.js
```diff
@@ -65,14 +65,17 @@ export function listTokenAccounts(account: Account): TokenAccount[] {
 }
 
 export function flattenAccounts(
-  topAccounts: Account[] | TokenAccount[] | (Account | TokenAccount)[]
+  topAccounts: Account[] | TokenAccount[] | (Account | TokenAccount)[],
+  enforceHideEmptyTokenAccounts?: boolean
```
gre: for readability i suggest this is a `{ enforceHideEmptyTokenAccounts?: boolean }`

MortalKastor:
It's for `SelectAccount`, in the _Send_ flow in particular

PR #317, opened 2019-08-05
Attempt to implement freshAddresses

pbondoer:
# freshAddresses proposal

> :diamond_shape_with_a_dot_inside: **LL-1701** for context

* Deprecate (but still support) `freshAddress` and `freshAddressPath`. They now correspond to `freshAddresses[0]`.
* Add `freshAddresses` from libcore, which is guaranteed to have items if synced. Otherwise this array is empty.

Feedback very much appreciated, I'm not 100% sure how to test this properly :mag:


pbondoer:
@meriadec I had the same question, I would assume so since the idea is to deprecate those fields but I'm not sure if there's specific things to take into consideration while doing that

pbondoer:
@meriadec Oh, that's on desktop! Not for userland yet, not right now at least. There's some references to those fields in `live-common` still though that we might need to take care of.

gre:
I think it's safe to merge this and later start migrating all the freshAddress. if the tests pass on this it's a good first step and we'll bump live-common and confirm it doesn't break anything before before we migrate it all.

PR #319, opened 2019-08-08
Allow exporting to BRIDGESTREAM_DATA and use app.json as source

juan-cortes:
Since https://github.com/LedgerHQ/ledger-live-mobile/pull/994 got merged already we have a way to trigger an import response on mobile (android), this  pr build on that premise allowing the previously called `liveQR` command (now `exportAccounts`) to _either_ output to a LiveQR in the terminal or directly  return a `BRIDGESTREAM_DATA` string that we can then pipe into `yarn android:import` on mobile.

If we don't pass the `--out` flag it will behave as it did before by outputting to a liveqr stream.

### Sample exports 
```
ledger-live exportAccounts --appjsonFile "/path/to/app.json" --out
ledger-live exportAccounts --xpub XPUB_STRING --currency bitcoin --out
```

### How to use to import on mobile
Simply pipe the output of any of the previous commands to `yarn android:import $(YOUR_EXPORT)`

### DIFF
--- a/tool/src/scan.js
+++ b/tool/src/scan.js
@@ -1,7 +1,7 @@
 // @flow
 
 import { BigNumber } from "bignumber.js";
-import { from, defer, of } from "rxjs";
+import { from, defer, of, throwError } from "rxjs";
 import { skip, take, reduce, mergeMap, map, concatMap } from "rxjs/operators";
 import { fromAccountRaw } from "@ledgerhq/live-common/lib/account";
 import { syncAccount } from "@ledgerhq/live-common/lib/libcore/syncAccount";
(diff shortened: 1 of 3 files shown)


tool/src/scan.js
```diff
@@ -109,7 +116,24 @@ export const inferCurrency = ({ device, currency, file, xpub }) => {
 };
 
 export function scan(arg) {
-  const { device, xpub: xpubArray, file, scheme, index, length } = arg;
+  const { device, xpub: xpubArray, file, appjsonFile, scheme, index, length } = arg;
+
+  if (typeof appjsonFile === "string") {
+    const appjsondata = appjsonFile
+      ? JSON.parse(fs.readFileSync(appjsonFile, "utf-8"))
+      : { data: { accounts: [] } };
```
gre: i'm not sure why a else case is required?

tool/src/scan.js
```diff
@@ -109,7 +116,24 @@ export const inferCurrency = ({ device, currency, file, xpub }) => {
...
+
+    if (typeof appjsondata.data.accounts === "string") {
+      return throwError(
+        new Error("encrypted ledger live data is not supported")
+      );
+    }
+    return from(appjsondata.data.accounts.map(a=>fromAccountRaw(a.data))).pipe(
```
gre: can you prettier it?

gre: i'll do it from master

PR #321, opened 2019-08-13
Update lru-cache dependency

meriadec:
having this out-of-date sub-dep prevent us from bumping live-common to latest
no real changes except it now use class.

see https://github.com/isaacs/node-lru-cache/commit/9ce69193c4cefbeee27e759b1cf94e02a3eadf8b


gre:
To be carefully tested because I think the reason of not upgrading was that it was just not working on reactnative

gre:
we'll give a try to this and if it breaks somewhere we can still rollback.

PR #325, opened 2019-08-19
LL-1768 Changed on the string sanitazition logic

juan-cortes:
The issue was brought up by @Arnaud97234 , on some input fields deleting the 1 from 1000 would automatically return 0 (which is fine) but put then typing a 1 (sending 01) to validation would also return zero meaning you couldn't type anymore, this wasn't the only buggy behaviour related to leading zeros.

This pr removes leading zeros that are not immediatelly followed by a decimal separator.

### DIFF
--- a/src/currencies/sanitizeValueString.js
+++ b/src/currencies/sanitizeValueString.js
@@ -24,11 +24,14 @@ export const sanitizeValueString = (
       if (decimals >= 0) {
         decimals++;
         if (decimals > unit.magnitude) break;
-        value += c;
+        value = value === "0" ? c : value + c;
         display += c;
       } else if (value !== "0") {
         value += c;
         display += c;
+      } else {
+        value = c;
+        display = c;
       }
     } else if (decimals === -1 && (c === "," || c === ".")) {
       if (unit.magnitude === 0) {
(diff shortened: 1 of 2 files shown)


src/currencies/sanitizeValueString.js
```diff
@@ -18,6 +18,8 @@ export const sanitizeValueString = (
   let display = "";
   let value = "";
   let decimals = -1;
+  const trimLeadingZeros = new RegExp(`^([0]+)(?!$)(?!\\${dot})`, "g");
```
gre: can we try to implement this in the `for` loop under it? it's really designed to be as efficient as possible because it's a critical part to be efficient, especially for mobile (otherwise the keyboard input does not answer fast enough and it's going jumpy)

juan-cortes: makes sense, we don't really need this regexp anyway

gre:
@juan-cortes can you recreate the commit & reset the authors? it was weirdly squashed into a commit `v7.11.0-alpha.2`

juan-cortes:
> @juan-cortes can you recreate the commit & reset the authors? it was weirdly squashed into a commit `v7.11.0-alpha.2`

I don't know why I ended up like that, a bad rebase I guess. Fixed.

PR #338, opened 2019-08-29
Remove mandatory account flatten for getPortfolio

juan-cortes:
-----
### ⚠️ This is a breaking change that requires updating [desktop](https://github.com/LedgerHQ/ledger-live-desktop/blob/0a0b7c75ab3a75617116ee29b374d759f34cf95f/src/actions/portfolio.js#L57) and [mobile](https://github.com/LedgerHQ/ledger-live-mobile/blob/5c74a84979669432e43da6984395071decf6b394/src/actions/portfolio.js#L60)
-----

This allows us to get the combined history of a subset of accounts without assuming that we want the to be flattened. This is needed for the crypto assets page for example since I dont want to get the token accounts included when I'm looking at the ETH page

### DIFF
--- a/src/__tests__/portfolio.js
+++ b/src/__tests__/portfolio.js
@@ -11,6 +11,7 @@ import {
 } from "../portfolio";
 import { genAccount } from "../mock/account";
 import { baseMockBTCRates } from "../countervalues/mock";
+import { flattenAccounts } from "../../lib/account";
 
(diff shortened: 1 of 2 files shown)


juan-cortes:
We don't need this anymore (does no harm to merge it but no need) due to https://github.com/LedgerHQ/ledger-live-common/pull/339

PR #343, opened 2019-09-05
add Zcash transaction tests

Arnaud97234:


gre:
good for now but we probably should merge zcash and bitcoin because it's same family so it's probably we'll have more code sharing. also wait a bit before adding more coins, we definitely should code share more logic and need to see what smart helper we can write. talk to you more tomorrow.

PR #347, opened 2019-09-08
Added Galilel (GALI) support for ledger-live-common

mbroemme:
Added Galilel (GALI) Support for ledger-live-common libs


mbroemme:
Any comments, ACK, nACK, what else is needed? Hard to track status otherwise. :)

gre:
The first step in Ledger integration is to develop an app on the device. Only then, we start working on the front end.

Please reach our team on our slack. Thanks.

(this PR can still be referred and reopened in the future, closing for now)

PR #358, opened 2019-10-09
polish bridge impl

gre:


gre:
tests are broken because funds were moved. we'll recover tests later

PR #359, opened 2019-10-10
wip app auto update logic

gre:


gre:
no tests at the moment because it's a mvp. 
we can merge it for testing purpose (not used anywhere)
when we finish that work, we'll also add tons of tests which should be pretty easy to do as it's all implemented in reducers.

src/apps/list.js
```diff
@@ -97,14 +105,22 @@ const listApps = async (
...
+  const apps = sortedCryptoApps.concat(filtered);
+
+  const installed = installedList.map(({ name, hash }) => {
+    const ins = apps.find(i => name === i.name);
     return {
-      ...app,
-      installed: !!ins,
-      updated: ins ? ins.hash === app.hash : false
+      name,
+      updated: ins ? ins.hash === hash : false
```
juan-cortes: I think that you guys talked about changing this to `uptodate` in the call.
In case it's relevant.

src/hw/installApp.js
```diff
@@ -20,6 +22,19 @@ export default function installApp(
   }).pipe(
     filter(e => e.type === "bulk-progress"), // only bulk progress interests the UI
     throttleTime(100), // throttle to only emit 10 event/s max, to not spam the UI
-    map(e => ({ progress: e.progress })) // extract a stream of progress percentage
+    map(e => ({ progress: e.progress })), // extract a stream of progress percentage
+    catchError((e: Error) => {
+      if (!e || !e.message) return throwError(e);
+      const status = e.message.slice(e.message.length - 4);
+      if (status === "6a83") {
```
juan-cortes: it would be great if we had constants with the meaning of the apdus instead of codes somehow -.-

src/apps/list.js
gre: reminder for next iteration. we'll rediscuss, i'm not really annoyed by the "updated" wording personally

src/hw/installApp.js
gre: yeah, we in fact have with `TransportStatusError` but it turns out this error directly comes in a string from the HSM and in fact not the device (that proxy the apdu). we'll see later what we can do)

PR #360, opened 2019-10-13
WIP Stellar integration with a JS bridge

juan-cortes:
image

### Preamble
I decided to revisit the version of the initial integration from last year. This is a work in progress and it has some **breaking changes** that we need to make sure are solvable to begin with. I will list all the issues and solutions (if any) below and we can work from there.

The current status of the pr is that we can use the cli to scan, receive, send (without memo support). I managed to enable it locally on LLD but it unreliable and I hit a nested dependency version mismatch with bignumber.js and I can't successfully sign a transaction (more on this below).

### What is done
- [x] Account scanning
- [x] Bridge js integration
- [x] Receive flow (getAddress com)
- [x] Send flow through cli
- [x] Types
- [x] Asset update ~When I integrated it into desktop I noticed it still shows with the little rocket logo instead of the one above~ New logo and color integrated now.
- [x] **Code cleanup** This is a dirty weekend attempt, I'd like to clean it up before a possible merge.
- [x] Refactor to the new jsHelpers bridge style
- [x] **Memo support** at the cli and lld level, the bridge should be ready for it though.

### What is not done
- [ ] **Correct fee calculation**. Current approach is a naive one assuming we spend 0.00001XLM
- [ ] **Correct base reserve calculation**. According to the docs regarding the minimum balance an account needs to have we need to follow `Minimum Balance = (2 + # of entries) * base reserve`. Currently it assumes `1 XLM` since it's the minimum
- [ ] Either drop stellar-sdk or figure out a way to not crash on bignumber
--------

### Concerns ⚠
Maybe @gre or someone else can figure this out but I'm currently only able to build this by using node `10.16.3` this seems to be a requirement from stellar-sdk. This means that we need to rebuild the modules and there's a chance of breaking things.

When running on desktop I hit an error on the signing step that (took me forever to figure it out) is caused by the version mismatch of bignumber used between stellar-base and live-common. Ideally things would just work but for whatever reason it just breaks, attempting to call a dropped method (to the eyes of the up to date library). Maybe this is caused by yalc? I have no ideas anymore.


src/families/stellar/bridge/js.js
```diff
@@ -0,0 +1,379 @@
...
+          operations: mergeOps(a.operations, newOperations),
+          blockHeight: _blockHeight || 0
+        }));
+      } else {
+        const lastLedger = await api.getLastLedger();
+
+        o.next(a => ({
+          ...a,
+          balance,
+          blockHeight: +lastLedger.records[0].paging_token||0,
```
gre: :wtf:

src/derivation.js
```diff
@@ -25,6 +25,10 @@ const modes = Object.freeze({
   // this is "default" by convention
   "": {},
 
+  stellar: {
+    overridesDerivation: "44'/148'/<account>'",
+    tag: "stellar"
```
gre: I don't think the tag is correct. Do you have a hint where this derivation scheme would be used? If not, maybe we call this legacy.

stellar on bip44 is standard / used by any app today?

juan-cortes: Right, remember when I wrote most of this code I had no idea what anything was. Tag can be removed then since we're not using bip44 and it's not really legacy

juan-cortes:
Maybe next time

PR #363, opened 2019-10-16
TRON: add send transaction on CLI

KhalilBellakrid:
Since we rely already on `trongrid` for sync of accounts and we want quick support of new currencies, I relied on `trongrid` to forge and broadcast transactions :trollface:  .
This was tested many times with CLI + NanoS of @FabriceDautriat 🚀  , SFYL 🚎, FYI I made transactions between (`44'/195'/0'/0/0` => `44'/195'/0'/0/1`) since self-transactions are forbidden.

### DIFF
--- a/src/derivation.js
+++ b/src/derivation.js
@@ -81,6 +81,9 @@ const modes = Object.freeze({
       TEZOS_XPUB_CURVE: "SECP256K1" // FIXME bug in libcore? it should be ED25519
     }
   },
+  tron: {
+    overridesDerivation: "44'/195'/<account>'/0/0"
+  },
   native_segwit: {
     purpose: 84,
     libcoreConfig: {
--- a/src/derivation.js
+++ b/src/derivation.js
@@ -81,9 +81,6 @@ const modes = Object.freeze({
       TEZOS_XPUB_CURVE: "SECP256K1" // FIXME bug in libcore? it should be ED25519
     }
   },
-  tron: {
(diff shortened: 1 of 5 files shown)


src/derivation.js
```diff
@@ -81,6 +81,9 @@ const modes = Object.freeze({
       TEZOS_XPUB_CURVE: "SECP256K1" // FIXME bug in libcore? it should be ED25519
     }
   },
+  tron: {
+    overridesDerivation: "44'/195'/<account>'/0/0"
```
gre: i don't think it's required because it's basically the default derivation no? (bip44)

src/families/tron/bridge/js.js
```diff
@@ -188,7 +273,35 @@ const getTransactionStatus = a =>
...
+        o.error(e);
+      }
+    );
+    return () => {
+      cancelled = true;
+    };
+  });
 
 const prepareTransaction = async (a, t: Transaction): Promise<Transaction> =>
   Promise.resolve(t);
```
gre: we'll need to impl the prepareTransaction & getTransactionStatus to get full support (required for send form basically). we can do this later in second step.

src/derivation.js
KhalilBellakrid: Yes it is, I thought it was mandatory field.

src/families/tron/bridge/js.js
KhalilBellakrid: +1 better do that later !

PR #367, opened 2019-10-16
Add pending Operation time retention strategy

gre:
this is to workaround operation instability before it gets available in blockchain/explorer/libcore.
we will always keep a pending op for 30mn (env configurable) and we will just always dedup when getting op list.

Should Fixes LL-1816

To be tested more.


gre:
hodl for now, need more testing.

PR #369, opened 2019-10-17
NEO: add send transaction on CLI

KhalilBellakrid:
Add `send` on CLI for NEO 🚀 .
So I managed to test: 
- [x] Build of transaction, 
- [x] Stream to device, 
- [x] Display of right informations on device.

Unfortunately, the only command I can validate on device is `Deny` (which is a good feature :trollface: haha), this is the only remaining issue to debug. I already checked the hex of the raw unsigned tx and the streamed data to device and both look fine to me, but I will spend more time to find why `Sign Tx` and `Sign Now Tx` are blocked :rage4: !

I've already implemented: 
- [x] Signature injection in raw transaction,
- [x] Broadcast of transaction. 

But of course because of above reasons I couldn't test (since `tester c'est douter` it should be ok :trollface: ).

TODO:
- [x] Find remaining issue on tx validation on device,
- [x] (Alright let's )Test broadcast, 
- [x] Think about strategy of fee selection (please refer to `families/neo/bridge/js.js`).

Update: 
Everything is fixed, I managed to sign and broadcast a transaction 🚀  , @gre I gave you back the 1 NEO you sent to me 🎊 : [d9bc8acfe90d3e136272b59d0b2e44e9e71f5665b6d07a58dee2dbc348ea908f](https://neoscan.io/transaction/d9bc8acfe90d3e136272b59d0b2e44e9e71f5665b6d07a58dee2dbc348ea908f)

### DIFF
--- a/package.json
+++ b/package.json
@@ -43,6 +43,7 @@
     "react": "*"
   },
   "dependencies": {
+    "@cityofzion/neon-js": "^4.7.1",
     "@ledgerhq/compressjs": "1.3.2",
     "@ledgerhq/errors": "4.72.1",
     "@ledgerhq/hw-app-btc": "4.72.1",
--- a/src/families/neo/hw-app-neo/index.js
+++ b/src/families/neo/hw-app-neo/index.js
@@ -6,7 +6,6 @@ import { getAddressFromScriptHash, getScriptHashFromPublicKey } from "./crypto";
 import { foreach } from "./utils";
 const CLA = 0x80;
 const PATH_SIZE = 4;
-const PATHS_LENGTH_SIZE = 1;
 const CHUNK_SIZE = 240;
 const SIGN = 0x02;
 /**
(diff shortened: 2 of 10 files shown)


src/families/neo/hw-app-neo/index.js
```diff
@@ -35,4 +40,48 @@ export default class Neo {
...
+      if (i === buffers.length - 1) {
+        startByte = 0x80;
+      } else {
+        startByte = 0x00;
+      }
+      return [startByte, p];
+    });
+
+    let response;
+    return foreach(buffers, ([startByte, data]) => {
```
gre: we can rewrite that with a simple `for` loop with `await` in it.

package.json
```diff
@@ -43,6 +43,7 @@
     "react": "*"
   },
   "dependencies": {
+    "@cityofzion/neon-js": "^4.7.1",
```
gre: that's the most annoying part. to be estimated if it's good to dep too much on libs. if minimal integration would be possible without the lib, otherwise we can still use it.

recently @juan-cortes had bad surprise with Bignumber.js version mismatch so we'll be careful about these kind of dep (which bring many dep).

need to see if stuff can be inlined. I already did a bunch of inlining in https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/neo/hw-app-neo/crypto.js

KhalilBellakrid: Yeah I've seen it ! I went with the pretty straightforward/fast implementation using existing libraries, we can think later about getting rid of those or when we support it from libcore side !

gre:
this PR is outdated. we can use it as reference for later.

PR #377, opened 2019-10-18
Fix extracting minor version inside seVersion

meriadec:
The regex was extracting only the first char of the minor.

Didn't caused trouble until now because all our minor were 1 char LOL :smile:
But Nano S firmw upgrade 1.6.0 is flashing MCU from bootloader `0.11`, which was extracted as `0.1` and this was very likely causing the failure during our tests.


meriadec:
testing is douting

gre:
Please have a test that covers this.

PR #382, opened 2019-10-22
fixes missing an error on not enough fee on send max bitcoin

gre:


gre:
and yeah, it would require a test to cover this!

PR #396, opened 2019-11-05
Currency Bridge, dropping deprecated methods, Account#spendableBalance

gre:
This PR will contain a series of breaking change that will be the next live-common MAJOR bump.


gre:
- [ ]  `available_capacity` can be negative

src/families/bitcoin/bridge/libcore.js
```diff
@@ -135,6 +135,7 @@ const prepareTransaction = async (
 };
 
 const getCapabilities = () => ({
+  canDelegate: false,
```
gre: Maybe it should be canStake

gre: ok let's keep this canDelegate for now. there is a clear semantic of delegation atm. we'll see later what we generalize.

PR #400, opened 2019-11-11
Manager v2 requirements integration

gre:
- [ ] apps/hw listApps to not fetch marketcap (that stuff is to solve on UI side)
- [x] apps/hw listApps to not fallback on the listApps apdu (we want to do HSM listapp always for now and let the error throw)


gre:
@juan-cortes i let you take over this PR the next days. you should try to move in all the sorting and filtering logic (i've created a file apps/filtering.js)

gre:
ok we'll merge this in fact, it's working with your desktop PR so we can move on, but it would be great to move in the filter/sort logic then ;)

the model "App" is a new object and you should have everything in that to be able to sort things out (there is the dependencies in that model as well as the market cap index)

PR #403, opened 2019-11-12
LL-2013 Improved demo/partners, added Crypto.com

juan-cortes:
### Description
This pr brings back the /demo/partners page where we can check the links and logos used by desktop/mobile for the partners. This was broken by https://github.com/LedgerHQ/ledger-live-common/commit/3b893e2ae58e87dadc40f4246c76cb449b621443#diff-70713b253c47c97cab35a8d690d44b21R14 since the introduction of the dark parameter. I've added a dark mode toggle in order to make sure that we have the light versions of the partners too.

-------------------

~In the process of add crypto.com, **we lack the dark mode version and the light version renders incorrectly** when converted to react SVG, passing the torch to design to give me pretty logos, please. You can find the original SVG in the linked issue, let me know if you need anything else, it's the same as usual, remove all the junk and simplify it so that react doesn't choke. @khalilbenihoud @bscholving @dekkaki (you guys decide who). For the dark mode, there are some references on their website but no clear downloadable candidate.~ 

Thanks to @bscholving for providing clean beautiful versions of the SVGs
 
### Related
https://ledgerhq.atlassian.net/browse/LL-2013

### Screenshots
image
image


juan-cortes:
Was supposed to be draft until I get the correct svgs, adding hodl tag to prevent merging.

demo/package.json
```diff
@@ -8,7 +8,7 @@
     "@ledgerhq/hw-transport-web-ble": "^4.73.2",
     "@ledgerhq/hw-transport-webhid": "^4.73.2",
     "@ledgerhq/hw-transport-webusb": "^4.73.2",
-    "@ledgerhq/live-common": "^9.0.0-alpha.1",
+    "@ledgerhq/live-common": "file:.yalc/@ledgerhq/live-common",
```
gre: :ah:

juan-cortes: I honestly left it on purpose because, how does this even work? I neer a bump regardless for the demo to work 🤷‍♂

gre: demo must always point to a release live-common otherwise it can't build and deploy to netlify.

I think your changes is compatible with `^9.0.0-alpha.1` so i suggest you just rollback to it.

it's ok if it does not yet includes your icon addition, it will appear later when we rebump.

gre:
for the other CI failure, i'll check on master.

demo/package.json
juan-cortes: Removed the yalced, _demo_ will not have latest icons until it's bumped though

PR #404, opened 2019-11-14
throw NotEnoughBalanceInParentAccount if parent can't afford fees

juan-cortes:
> Requires https://github.com/LedgerHQ/ledgerjs/pull/420 and a bump

For Tezos child accounts, check if the parent account can afford the transaction fees and throw the appropriate error if it can't.


juan-cortes:
Once again, this was supposed to be a draft.

src/families/tezos/bridge/libcore.js
```diff
@@ -132,6 +132,10 @@ const getTransactionStatus = async (a, t) => {
     );
   }
 
+  if (estimatedFees.gt(a.balance)) {
```
gre: to me we should do that if:
- there is no errors.amount already
- it's a "subAccuont" transaction 🤔

juan-cortes: current version should do that

gre:
I can't find this PR it keeps saying #404

PR #411, opened 2019-11-26
hourly support

gre:
unfortunately this is not performant enough to get to prod anytime soon. we need a more performant API first.

it works if you rollback this https://github.com/LedgerHQ/ledger-live-desktop/pull/2279/files


gre:
work has been integrated in https://github.com/LedgerHQ/ledger-live-common/pull/437

PR #412, opened 2019-11-26
babel 7 upgraded

IAmMorrow:


src/families/ethereum/cli-transaction.js
```diff
@@ -74,7 +74,7 @@ function inferTransactions(
       userGasLimit: new BigNumber(opts.gasLimit),
       estimatedGasLimit: null
     };
-  });
+  }).map();
```
gre: Pretty obscure.

`undefined is not a function` no?

if there is magic behind that, well, please let's just explicitly lodash.`flatMap`.

src/families/ripple/cli-transaction.js
```diff
@@ -21,14 +21,14 @@ function inferTransactions(
...
-  return transactions.flatMap(({ transaction, account }) => {
+  return transactions.map(({ transaction, account }) => {
     invariant(transaction.family === "ripple", "ripple family");
     return {
       ...transaction,
       fee: inferAmount(account, opts.fee || "0.001xrp"),
       tag: opts.tag
     };
-  });
+  }).map();
```
gre: same

src/families/tezos/cli-transaction.js
```diff
@@ -69,7 +69,7 @@ function inferTransactions(
       gasLimit: opts.gasLimit ? new BigNumber(opts.gasLimit) : null,
       storageLimit: opts.storageLimit ? new BigNumber(opts.storageLimit) : null
     };
-  });
+  }).map();
```
gre: same

.babelrc
```diff
@@ -1,17 +1,22 @@
...
     [
-      "env",
+      "@babel/preset-env",
       {
         "targets": {
-          "browsers": ["last 2 versions"],
-          "node": "current"
-        },
-        "forceAllTransforms": true
+          "node": "current",
```
MortalKastor: Can we use a [browserlist](https://github.com/browserslist/browserslist) query in a `. browserslistrc` (or under a `browserslist` key in `package.json`) instead of targets in `.babelrc`?

This is what `@babel/preset-env` doc recommends, as it's followed not only by Babel but other tools as well.

Also we should explicitly target the lowest `JSC` version Ledger Live Mobile has to deal with. For Android we ship the latest version, but on iOS we depend on what the OS offers.
AFAIK the lowest iOS version we support is the same as RN itself, so that would be 9.0.

And while we're at it, target our specific lowest Electron we use (unless we don't merge before migrating LLD?).

As a browserlist query that would be something akin to `electron 1.2, ios_saf 9, current node` (`ios_saf` is iOS Safari, which version follows the OS it's shipped with)

PR #421, opened 2019-12-13
Sorting and filtering utils for the managerv2 for mobile and desktop

juan-cortes:
### What is this
Shared logic for sorting/filtering the list of apps in the catalog. Currently supporting filtering by installed, not installed, needing an update, supported by live (maybe we can do more?) and sorting by name and marketcap.

### Usage
Since most of the time we would like to compound a search and a filter we have the `sortFilterApps` which  receives both sets of options and returns the filtered and ordered app array. The only thing to keep in mind is that we need to provide the installed app list (which is not of type `App[]` but rather `InstalledItem[]`) from the state in order to match with the catalog. Same applies for sorting from marketcap, the function expects us to provide the sorted tickers list, which can be obtained in a number of ways with the easiest being `const marketcapTickers = useMarketcapTickers();`.

### Testing?
I believe @gre wanted some tests on live-common for this so, that part is still pending. I don't want to write tests before the implementation is approved and makes sense.

### DIFF
--- a/src/apps/filtering.js
+++ b/src/apps/filtering.js
@@ -2,11 +2,11 @@
 import type { App } from "../types/manager";
 import type { InstalledItem } from "./types";
 import { getCryptoCurrencyById, isCurrencySupported } from "../currencies";
+import { useMemo } from "react";
 
 export type SortOptions = {
-  type: "name" | "marketcap",
-  order: "asc" | "desc",
-  marketcapTickers?: string[]
+  type: "name" | "marketcap" | "default",
+  order: "asc" | "desc"
 };
 
 export type FilterOptions = {
@@ -17,10 +17,6 @@ export type FilterOptions = {
 
 type UpdateAwareInstalledApps = {
   [string]: boolean // NB [AppName]: isUpdated
-};
-
-type ScoredApps = {
-  [string]: number // NB [AppName]: computed score for sorting
 };
 
 const searchFilter = (query?: string) => ({ name, currencyId }) => {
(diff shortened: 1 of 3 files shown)


src/apps/filtering.js
```diff
@@ -1,10 +1,124 @@
...
+  const terms = `${name} ${
+    currency ? `${currency.name} ${currency.ticker}` : ""
+  }`;
+  return terms.toLowerCase().includes(query.toLowerCase().trim());
+};
+
+const typeFilter = (type = "all", installedApps?: InstalledItem[]) => app => {
+  switch (type) {
+    case "installed":
+      invariant(installedApps, missingInstalledApps);
```
gre: invariant not needed if you just make it mandatory in options

src/apps/filtering.js
```diff
@@ -1,10 +1,124 @@
...
+};
+
+export const sortApps = (apps: App[], _options: SortOptions): App[] => {
+  const { type, order = "asc", marketcapTickers } = _options;
+  const asc = order === "asc";
+  let sortedApps = [...apps];
+
+  switch (type) {
+    case "marketcap":
+      invariant(marketcapTickers, missingMarketcapTickers);
```
gre: same

src/apps/filtering.js
```diff
@@ -1,10 +1,124 @@
...
+
+      return [...currencyApps, ...otherApps];
+    default:
+      sortedApps.sort((a1, a2) => {
+        const [app1, app2] = asc ? [a1, a2] : [a2, a1];
+        return app1.name.localeCompare(app2.name);
+      });
+  }
+
+  return sortedApps;
```
gre: I think the code above here is way too complex for what it's doing. the O() complexity is not that great here. it's better to build up an array for performance.

src/apps/filtering.js
```diff
@@ -1,10 +1,124 @@
...
+      return true;
+  }
+};
+
+export const sortApps = (apps: App[], _options: SortOptions): App[] => {
+  const { type, order = "asc", marketcapTickers } = _options;
+  const asc = order === "asc";
+  let sortedApps = [...apps];
+
+  switch (type) {
```
gre: we could use sub functions per type in a map.

alternatively an idea is to have some sort of score. especially handy in the "second level sorting" (marketcap but then what)

see https://github.com/LedgerHQ/ledger-live-common/blob/master/src/account/ordering.js#L27

src/apps/filtering.js
```diff
@@ -1,10 +1,124 @@
...
+    case "installed":
+      invariant(installedApps, missingInstalledApps);
+      return installedApps.find(a => a.name === app.name);
+    case "not_installed":
+      invariant(installedApps, missingInstalledApps);
+      return !installedApps.find(a => a.name === app.name);
+    case "updatable":
+      invariant(installedApps, missingInstalledApps);
+      return installedApps.find(a => a.name === app.name && !a.updated);
```
gre: it's better if we have beforehand a map of {installed} so we don't need to .find() at each app iteration

src/apps/filtering.js
```diff
@@ -1,10 +1,104 @@
...
+export const useSortedFilteredApps = (
+  apps: App[],
+  _filterOptions: FilterOptions,
+  _sortOptions: SortOptions
+) => {
+  const { query, installedApps, type: filterType } = _filterOptions;
+  const { type: sortType, order } = _sortOptions;
+
+  return useMemo(
+    () => sortApps(filterApps(apps, _filterOptions), _sortOptions),
```
gre:
```suggestion
    () => sortFilterApps(apps, { query, installedApps, type: filterType }, { type: sortType, order }),
```

PR #422, opened 2019-12-14
Workaround socket

gre:
Workaround for a socket error to appear before the channel normally close


gre:
not working properly. more work will be needed

PR #426, opened 2019-12-20
Rewording and remove block style

dasilvarosa:
- rewording
- Replaced block styles with regular paragraphs

### DIFF
--- a/tools/src/demos/lld-signature/index.js
+++ b/tools/src/demos/lld-signature/index.js
@@ -185,10 +185,10 @@ const LLDSignature = () => {
 
       <h2>Verify my Ledger Live install binary</h2>
 
-      <block>
+      <Block>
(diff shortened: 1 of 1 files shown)


tools/src/demos/lld-signature/index.js
```diff
@@ -185,11 +185,10 @@ const LLDSignature = () => {
 
       <h2>Verify my Ledger Live install binary</h2>
 
-      <Block>
-        You can verify with <strong>sha512</strong> that Ledger Live binary
-        installation file is correct by comparing its hash to the one available
+      <p>
+        You can verify the authenticity of the Ledger Live binary installation file by comparing its <strong>sha512</strong> hash to the one available
         in this file:
-      </Block>
```
gre: please undo the `<Block>` change, i already did in fact. `Block` is the current ok display you wanted.

PR #429, opened 2019-12-22
[DRAFT] Implement getBalanceHistory from libcore

juan-cortes:
The first approach to integrating `getBalanceHistory` from libcore in order to be able to fetch the balance history for an account without having to compute it from the transactions themselves. This could open the door for a more performant graph (with more features @IAmMorrow :trollface:)

Current draft has added the balance history to the Account model but no further logic has been added to actually populate that during sync, this was something we talked about during our peer programming sessions @gre 

In order to get the CLI to work, you'll need to yalc the live-common version.

**Note:** Behavior doesn't match the js computed version, there are some inconsistencies following https://ledgerhq.atlassian.net/browse/LLC-475


juan-cortes:
Rebased with develop

gre:
work has been integrated in https://github.com/LedgerHQ/ledger-live-common/pull/437

PR #430, opened 2019-12-28
Device test all apps Install / Update / Remove

Arnaud97234:
Use the cli ledger-live appsUpdateTestAll command to install, update and uninstall all Manager app on a Nano device.

Using a cron job on the remote computer, the script will update ledger-live-common repo then run the command. If an error occurs during the process an issue will be created on ledger-live-common. 

Prerequisites:
- Plug device on the remote computer
- Unlock device
- Allow Manager on device


cli/tests/device-appsUpdateTestAll.sh
```diff
@@ -0,0 +1,30 @@
...
+
+LOG_FILE=cli/tests/tmp/error.log
+
+if [ ! -f $LOG_FILE ]; then
+    echo "Log file couldn't be find"
+elif grep -i "NoDevice" $LOG_FILE; then
+    echo "A Nano device needs to be connected"
+elif grep -i -E "FAILED|ERROR" $LOG_FILE; then
+    echo "An error has occured. Creating an issue..."
+    node cli/tests/create-issue.js "`cat $LOG_FILE`"
```
gre: probably better to pass the filename (and to read from the JS) instead of the cat result.

cli/yalc.lock
```diff
@@ -0,0 +1,9 @@
+{
```
gre: file to not commit.

cli/tests/device-appsUpdateTestAll.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+
+GIT_REMOTE=git@github.com:LedgerHQ/ledger-live-common.git
+
+git clean -xdf &&
```
gre: if you add `set -e` at beginning of bash, it will have the same effect as all the `&&` so you can remove them

PR #432, opened 2019-12-30
Enables compound filtering + tests

juan-cortes:
As pointed out by @LFBarreto we were unable to perform compound filtering on the catalog/installed apps and, this addresses that feature and adds the appropriate tests to make sure it works. 🚀

### DIFF
--- a/src/apps/filtering.js
+++ b/src/apps/filtering.js
@@ -37,46 +37,34 @@ const searchFilter = (query?: string) => ({ name, currencyId }) => {
 };
 
 const typeFilter = (
-  type: AppType[] = ["all"],
+  filters: AppType[] = ["all"],
   updateAwareInstalledApps: UpdateAwareInstalledApps
-) => app => {
-  let validApp = true;
-
-  type.forEach(filter => {
+) => app =>
+  filters.every(filter => {
     switch (filter) {
       case "installed":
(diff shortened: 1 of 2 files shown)


src/apps/filtering.js
```diff
@@ -29,27 +37,44 @@ const searchFilter = (query?: string) => ({ name, currencyId }) => {
...
-      return (
-        app.currencyId &&
-        isCurrencySupported(getCryptoCurrencyById(app.currencyId))
-      );
-    default:
-      return true;
-  }
+  let validApp = true;
+
+  type.forEach(filter => {
```
gre: logic can be simplified a bit

- `type` naming is a bit weird, can be just "filters"
- use of `.every` would simplify the logic and not requires to have a `validApp` variable.

```js
const typeFilter = (
  filters: AppType[] = ["all"],
  updateAwareInstalledApps: UpdateAwareInstalledApps
) => app => filters.every(filter => {
  switch (filter) {
    case "installed": return !!updateAwareInstalledApps.hasOwnProperty(app.name);
    ...
  }
})`

juan-cortes: noice

juan-cortes: You have it like you asked for now

PR #439, opened 2020-01-06
LL2080 (ERC20): use ERC20Balances

LFBarreto:
(ERC20): use ERC20Balances to fetch in batch instead of loop calls

### DIFF
--- a/src/families/ethereum/types.js
+++ b/src/families/ethereum/types.js
@@ -237,7 +237,7 @@ export const reflect = (declare: (string, Spec) => void) => {
         returns: "BigInt"
       },
       getERC20Balances: {
-        returns: ["CoreBigInt"]
+        returns: ["BigInt"]
       }
     }
   });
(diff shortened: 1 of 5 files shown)


src/families/ethereum/types.js
```diff
@@ -234,6 +235,9 @@ export const reflect = (declare: (string, Spec) => void) => {
       },
       getEstimatedGasLimit: {
         returns: "BigInt"
+      },
+      getERC20Balances: {
+        returns: ["CoreBigInt"]
```
juan-cortes: `["BigInt"]` ?

gre: you can check running the project `mobile-test-app` btw, it's only in React Native bindings that this would break ;)

gre:
there is a few files that i also modified in #437 , i'll cherry-pick this branch from my PR.

gre:
thx for the work. works great. ETH sync with ~10 erc20 accounts are about 3 times faster.

will be merged as part of
https://github.com/LedgerHQ/ledger-live-common/pull/437

to we can batch the "QA effort" on testing everything

PR #440, opened 2020-01-06
Add `transportId` parameter to withDevice, withDevicePolling & open

meriadec:
:book:  A bit of context from this summer: https://github.com/LedgerHQ/ledger-vault-front/pull/1000#discussion_r317095477

:star: The purpose of this PR is to provide a way to target a specific transport if you:
- have registered multiple transports
- want to use `withDevice` / `withDevicePolling`

:raised_hand_with_fingers_splayed:  The today behavior is to automatically select the [first working registered transport](https://github.com/LedgerHQ/ledger-live-common/blob/77ef279cf71fedf926c6ab0680781f3096ecb03c/src/hw/index.js#L61-L65) which is not our use case (we want to let user choose what he want).

So now you can:

```js
registerTransportModule({ id: "u2f", ... });
registerTransportModule({ id: "webusb", ... });
registerTransportModule({ id: "quantum-mechanics", ... });

const userChoice = localStorage.getItem("PREFERRED_TRANSPORT") || "webusb";

withDevicePolling(deviceID, userChoice)(transport => { /* blablabl */ })
```

Oh, I hear you shouting:

# > "Why don't you just register one transport based on the user preference?"

That's what we do now. But the whole thing is to allow dynamically change the transport without reloading the page (which is [our current workaround now](https://github.com/LedgerHQ/ledger-vault-front/blob/51ce50ec91b2d0ab222b045ac74e5d224fc77c0a/src/components/TransportChooser.js#L30-L36)).

### DIFF
--- a/src/hw/deviceAccess.js
+++ b/src/hw/deviceAccess.js
@@ -81,7 +81,7 @@ export const cancelDeviceAction = (transport: Transport<*>) => {
 
 const deviceQueues = {};
 
-export const withDevice = (deviceId: string) => <T>(
+export const withDevice = (deviceId: string, transportId?: string) => <T>(
   job: (t: Transport<*>) => Observable<T>
 ): Observable<T> =>
   Observable.create(o => {
@@ -107,7 +107,7 @@ export const withDevice = (deviceId: string) => <T>(
 
     // for any new job, we'll now wait the exec queue to be available
     deviceQueue
-      .then(() => open(deviceId)) // open the transport
+      .then(() => open(deviceId, transportId)) // open the transport
       .then(async transport => {
         if (unsubscribed) {
           // it was unsubscribed prematurely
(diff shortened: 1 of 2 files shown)


gre:
I suggest you put your "transportId" in the deviceId itself

id: `${transportId}:${deviceId}`

that's a bit a pattern we've been using on LLM and CLI ( https://github.com/LedgerHQ/ledger-live-common/blob/c410a6eced6a767c943f031f906347fbcd7479e8/cli/src/live-common-setup.js#L37 )

meriadec:
Sorry but as I answered [in the previous discussion](https://github.com/LedgerHQ/ledger-vault-front/pull/1000#discussion_r317117738) I don't get what it has to do with `deviceId` which is used or not by the transport itself to instanciate (and for ex, u2f transport ***does not*** use `deviceId`).

The need here is to _choose_ which transport we want. I don't understand how we can achieve this in another way.

The second parameter to `withDevice`, `withDevicePolling` & `open` is totally optional, I don't see where it could have un-wanted impact.

But not opiniated on the impl, if there is a better solution that can allow us to dynamically choose transport without adding a param, we will embrace it.

meriadec:
After IRL discussion... Sounds that you are right and there is a way to _not touch_ the actual code AND achieve what we want. 

See you!

PR #445, opened 2020-01-07
(formatting): formatSize allow for defaultString value

LFBarreto:
allow for formatSize to take a default value.
usefull for managerV2 integration.


gre:
@LFBarreto could you cover it in a minimal test? thx

PR #446, opened 2020-01-07
put all tests together

gre:
### Tradeoffs

- I initially wanted to have our bridge tests under each family `src/families/*/bridge.test.js` but the problem with that is Jest (OR something else?) does not like to run in parallel file, it need to be in the same test file. so we are keeping the datatest files idea for now.
- The way it was unified is by changing Jest module resolver to `[ node_modules, cli/node_modules ]`, that way Jest will know where to find the libcore implementation (in the CLI). The tradeoff here is that you need to yarn once the cli folder before running tests.

### Future plan that is not yet implemented in this PR

- most of the tests of `cli/tests` will be dropped: they can be replaced by the new solution implemented here.

### What this PR actually do

- the tests of `cli/src/__tests__` has been moved in `src/__tests__`
- introduce `src/__tests__/test-helpers/bridge` that is a helper to make a bridge test
- introduce a way to test `scanAccounts`
- introduce a way to test `signOperation`

### What does this means for our CI and for QA

- We are going to have more reliable code coverage
- Performance might be better (even tho yarn in cli takes some time)
- We express the tests at the same place
- as soon as the future plan of testing signOperation is achieved, QA team to help on writing more tests coverage for existing coins. Implement some minimal scanAccounts APDUs and some minimal signOperation cases with APDU.

### What does this means for developers

- You only need to yarn in `cli` but you don't need to yalc and recompile every time to be able to test your latest code changes.


gre:
Capture d’écran 2020-01-13 à 12 06 56

posting idea for future, (we could do this in another private repo)

PR #447, opened 2020-01-09
(test): added minimal test for formatSize

LFBarreto:
added minimal test for formatSize changes introduced in #445


gre:
gonna run it in master

PR #448, opened 2020-01-09
Minor updates

gre:
- SignedOperation: add signatureRaw
- add coreAccount to signTransaction


src/transaction/signOperation.js
```diff
@@ -10,22 +10,36 @@ import type {
...
-});
+): SignedOperation => {
+  const { operation, signature, expirationDate, signatureRaw } = signedOp;
+  const out: SignedOperation = {
+    operation: fromOperationRaw(operation, accountId),
+    signature,
+    expirationDate: expirationDate ? new Date(expirationDate) : null
+  };
+  if (signatureRaw) {
+    out.signatureRaw = signatureRaw;
```
nukky94: To avoid mutation?
```
const out: SignedOperation = {
    operation: fromOperationRaw(operation, accountId),
    signature,
    expirationDate: expirationDate ? new Date(expirationDate) : null,
    signatureRaw
}
```

src/transaction/signOperation.js
```diff
@@ -10,22 +10,36 @@ import type {
...
-});
+): SignedOperationRaw => {
+  const { operation, signature, expirationDate, signatureRaw } = signedOp;
+  const out: SignedOperationRaw = {
+    operation: toOperationRaw(operation, preserveSubOperation),
+    signature,
+    expirationDate: expirationDate ? expirationDate.toISOString() : null
+  };
+  if (signatureRaw) {
+    out.signatureRaw = signatureRaw;
```
nukky94: Same:
```
const out: SignedOperationRaw = {
    operation: toOperationRaw(operation, preserveSubOperation),
    signature,
    expirationDate: expirationDate ? expirationDate.toISOString() : null,
    signatureRaw
  };
```

src/transaction/signOperation.js
gre: it's not equivalent.

in your version, `"signatureRaw in out` is true. not in mine.

`...(signature ? {signature} : {})` would be the equivalent to do it in "one" go. but it's not good for performance and these methods are heavily used in LLD so i prefer it's made in the most performant way possible (from end point of view of using the method, it's immutable)

src/transaction/signOperation.js
gre: and same answer :)

src/transaction/signOperation.js
gre: in general when something is really a specific field and we don't want to enforce it for all, I tend to do `field?: type` (aka the field itself it optional) vs when we want to enforce it (like the expirationDate here, it's really to push it's usage on all chain even tho we have set null today) I tend to `field: ?type`.

doing `field?: ?type` is possible but overkill.

in fact i wonder if `field?: type` would accept `field: undefined`, i think it does but it's sometimes we've seen problem arise with the type system.
sometimes we tested the present of a field to do logic, I don't know if it's a very good idea but it's possible and flow would be able to prove things. (if "field in out" is true, you are ensured that your type is not falsy)

nukky94: Wow, thanks for explanations. Didn’t know about this 2 kind of type (optional type param and maybe type). In TS there is no such distinction (I guess? Or I should miss it too :p)

PR #449, opened 2020-01-13
formatSize with blockSize

LFBarreto:
format size function can now take a block size as parameter

### DIFF
--- a/src/apps/formatting.js
+++ b/src/apps/formatting.js
@@ -1,8 +1,8 @@
 // @flow
 import type { AppOp, InstalledItem } from "./types";
 
-/** format bytes size according to given block size */
-export const formatSize = (size?: number, blockSize: number = 1) => {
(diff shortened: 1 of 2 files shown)


src/apps/formatting.js
```diff
@@ -1,8 +1,12 @@
 // @flow
 import type { AppOp, InstalledItem } from "./types";
 
-export const formatSize = (size?: number, defaultString?: string = "") =>
-  !size ? defaultString : Math.round(size / 1024) + "Kb";
+/** format bytes size according to given block size (defaults to 4096) */
+export const formatSize = (size?: number, blockSize: number = 4096) => {
+  const formatedSize =
+    size && Math.ceil((Math.ceil(size / blockSize) * blockSize) / 1024);
```
gre: I think only one ceil is needed.  https://github.com/LedgerHQ/ledger-live-common/blob/master/src/apps/logic.js#L105

PR #455, opened 2020-01-15
Update bitcoin tests

Arnaud97234:
The accounts used for the sync tests was not correct.
New apdus snapshot recorded.

In the bridge testing, there is still an XRP test that fails using mock implementation of the bridge, the same test is successful with ripplejs.


gre:
i wasn't able to commit in your fork, i just merged it manually by pushing on master

PR #457, opened 2020-01-17
Use one wallet for each account

gre:
This is an attempt to address problem with our current usage of libcore.

We've often noticed report from users that got their account data badly swapped and we always have suspected problem related to wallet index iteration.

In recent iteration, we've introduced new derivation path for Tezos, and they have intersection between each other (`44'/1729'/x'/0'` and `44'/1729'/0'/x'`) so we have special rules to "skip" an index (starts the second at index 1). This obviously reveal the bug.

The last usecase we have is we want to be able to have different configuration per account, and today configuration is at the wallet level.. which is not a concept we have in Live so we can't offer users a way to configure that settings.

The solution is simple: I don't think we need the concept of wallet, so we're just going to always create one wallet for each account. If this works without a glitch, i think we can move to this unless someone have a problem.

interestingly, on the way doing this, i've found an optimisation to cache the derivation so we don't ask `44'/0'` to the device N times.

### DIFF
--- a/src/account/accountId.js
+++ b/src/account/accountId.js
@@ -55,11 +55,13 @@ export function decodeAccountId(accountId: string): AccountIdParams {
 export function getWalletName({
   seedIdentifier,
   derivationMode,
-  currency
+  currency,
+  index
 }: {
   seedIdentifier: string,
   derivationMode: DerivationMode,
-  currency: CryptoCurrency
+  currency: CryptoCurrency,
+  index: number
 }) {
-  return `${seedIdentifier}_${currency.id}_${derivationMode}`;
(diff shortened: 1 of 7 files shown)


gre:
unfortunately, the code does not currently work for Bitcoin likes. libcore seems to highly rely on index to derivate the xpub, so we can't just do thing like `KEYCHAIN_DERIVATION_SCHEME="44'/0'/1'/<node>/<address>` with an index 0. 😢

gre:
currently not viable but we'll look forward to do this in the future: fact I had to rewrite the derivation that libcore is trying to do to rewrite the index is not viable. libcore uses that to also reconstruct fresh address...

PR #462, opened 2020-01-20
make operation fees from subaccounts be 0 in export

valpinkman:


src/csvExport.js
```diff
@@ -36,10 +36,12 @@ const fields: Field[] = [
   {
     title: "Operation Fees",
     cell: (account, parentAccount, op) =>
-      formatCurrencyUnit(getAccountCurrency(account).units[0], op.fee, {
-        disableRounding: true,
-        useGrouping: false
-      })
+      ["TokenAccount", "ChildAccount"].includes(account.type)
```
gre: I think it can be just `parentAccount ? ""`

gre: but it's not really important anyway. let's merge!

PR #463, opened 2020-01-21
Tezos scan accounts

Arnaud97234:
Tezos scanAccounts


cli/package.json
```diff
@@ -13,7 +13,7 @@
     "@ledgerhq/hw-transport-node-ble": "5.6.0",
     "@ledgerhq/hw-transport-node-hid": "5.6.0",
     "@ledgerhq/ledger-core": "^5.3.0",
-    "@ledgerhq/live-common": "^11.1.1",
+    "@ledgerhq/live-common": "file:.yalc/@ledgerhq/live-common",
```
gre: this should not be committed. 
in fact you no olnger need to do `yalc publish` or `yalc add` . it's only for developers now.

src/families/tezos/datasets/tezos.scanAccounts.1.js
```diff
@@ -0,0 +1,25 @@
+// @flow
```
gre: this file needs to be imported from the main test-dataset (see in similar family) otherwise it's not going to do anything.

gre:
I imagine this PR can be closed as it's in the other?

gre:
This is included in the other PR that was merged

PR #465, opened 2020-01-22
add mock bridge for tezos family

valpinkman:


src/families/tezos/bridge/mock.js
```diff
@@ -0,0 +1,207 @@
...
+  signOperation,
+  broadcast,
+  sync,
+  isInvalidRecipient
+} from "../../../bridge/mockHelpers";
+import {
+  // fetchAllBakers,
+  // hydrateBakers,
+  // asBaker,
+  isAccountDelegating
```
gre: at some point, we should not import `../bakers` I think.
we need to mock that.
we'll have to ask ourself how we want to address mocking bakers too.

until we don't mock bakers, it's ok to keep that, but it's going to be a real challenge. maybe bakers idea should be integrated as part of the bridge at some point? either that, or we branch some code based on `getEnv("MOCK")` but it's less powerful (i like that mock is a bridge because you can literally simultaneously have mock account and regular account. much powerful^^ – bakers.js instead is a global thing we import and not linked to a account context. we also want it to be global somehow so it's cached globally anyway...).


cc @nukky94 @henri-ly ^ your thoughts on this too would be interesting. nothing that will happen anytime soon I think but it's good we discuss at some point.

gre:
this is really curious that we are increasing the code coverage of src/families/tezos/bridge/libcore.js

PR #468, opened 2020-01-22
Update sync account

Arnaud97234:
- Record scanAccount in order to use frozen accounts (btc family + tezos)
- Add scanAccount test on Ethereum Classic
- Update device-appsUpdateTestAll.sh


cli/tests/device-appsUpdateTestAll.sh
```diff
@@ -8,9 +8,8 @@ cd `mktemp -d`
 git clone $GIT_REMOTE
 cd ledger-live-common
 yarn
-yalc publish
 cd cli
-yalc add @ledgerhq/live-common
+yarn add @ledgerhq/live-common
```
gre: this makes no sense and could have incompatibility issues. we should just yarn. so probably remove the line.

gre:
this PR is outdated.

PR #471, opened 2020-01-23
LL-2136 Account ordering prioritize starred accounts + tests

juan-cortes:
Allows to sort accounts taking into account starred accounts by receiving an id array. We will need to update the call to `sortAccountsComparatorFromOrder` on lld and eventually llm when we have stars over there too.


src/account/ordering.js
```diff
@@ -29,13 +29,20 @@ export const sortAccountsComparatorFromOrder = (
   calculateCountervalue: (
     currency: TokenCurrency | CryptoCurrency,
     value: BigNumber
-  ) => ?BigNumber
+  ) => ?BigNumber,
+  starredAccountIds: string[]
```
gre: can you make it so it's backward compatible? third param to fallbacks to `[]`

juan-cortes: Makes sense, yeah I can do that

juan-cortes: Done, with test 🚀

src/__tests__/accounts/ordering.js
```diff
@@ -0,0 +1,167 @@
...
+      "libcore:1:ethereum:xpub3:",
+      "libcore:1:ethereum:xpub3B:",
+      "libcore:1:ethereum:xpub1B:"
+    ]
+  );
+  const sortedAccounts = accounts.sort(compareFn);
+  expect(sortedAccounts.map(a => a.name)).toEqual(["AA", "C", "CA", "B", "A"]);
+});
+
+test("Accounts ordering | backwards compatible when no starred accounts are passed", () => {
```
gre: Thanks for doing this test too :rocket:

PR #474, opened 2020-01-25
Hack to disable sync token

gre:
honestly I still don't understand why we need /syncToken on our explorers api,
this PR introduces a DISABLE_SYNC_TOKEN flag to be used on the CLI for now,
i've even managed to pass our (extensive) tests by bypassing any usage of syncToken. Not only they all pass but the performance are better..

for now, this PR is here for @juan-cortes  to test the explorer locally.

### DIFF
--- a/cli/src/live-common-setup-base.js
+++ b/cli/src/live-common-setup-base.js
@@ -121,6 +121,49 @@ axios.interceptors.response.use(
   }
 );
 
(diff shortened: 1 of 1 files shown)


gre:
we'll merge to give more extensive testing on the CLI itself. (it only affects cli)

PR #477, opened 2020-01-29
Rewrite socket logic to fix race condition in transport

gre:
- rewrite the socket logic in order to make it work with upcoming script runner as well as the previous one
- add command appsInstallAll
- make appsUpdateTestAll to do a listApps after EACH install. It reveals the bug in current script runner and assess it's in fact fixed in the upcoming. We do that to also verify that the hash of install app matches the hash of the api (and we spot and reported problems on Nano X)
- Introduce a way to mock the socket that is enabled in MOCK=1 mode (will be eventually used for UI test purpose)
- the socket.js logic have the mecanism to detect the "allow manager" built in into it which allowed to simplify some code
- the logs observable exported from api/socket has been dropped because we are just using the @ledgerhq/logs like other parts
- the progress event out of the apps logic is throttle by 100ms

This will require heavy testing and we'll do it in context of "lldv2" and "llm", i'll release a .alpha for this purpose

### DIFF
--- a/cli/src/commands-index.js
+++ b/cli/src/commands-index.js
@@ -1,4 +1,5 @@
 import app from "./commands/app";
+import appsInstallAll from "./commands/appsInstallAll";
 import appsUpdateTestAll from "./commands/appsUpdateTestAll";
 import balanceHistory from "./commands/balanceHistory";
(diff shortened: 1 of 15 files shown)


meriadec:
:rocket:

PR #480, opened 2020-01-30
Documenting currency.md

gre:
Bootstrap currency.md (some parts need to be continued)

### DIFF
--- a/docs/currency.md
+++ b/docs/currency.md
@@ -1,0 +1,159 @@
+# The Currency models
+
+Ledger Live deals with different kind of currencies and assets: fiat currencies, crypto-currencies and tokens.
+They all share similar concept but have specifics.
+
+## The types
+
+The types are defined in [types/currencies](../src/types/currencies.js).
+
+live-common exposes 3 different types: `CryptoCurrency`, `TokenCurrency` and `FiatCurrency`.
+It also exposes a union type `Currency` that is either of these 3 types. _Make sure you use the correct type, sometimes you want the union type Currency, sometimes you are really in context of a CryptoCurrency_.
+
+To ease the differentiation of types, live-common data also introduces a _discriminator field_, `.type`, which have the same name of the type.
+
+### Common fields
+
+Essentially all currencies share this common ground:
+
+- `name: string`: the display name of a currency. For instance `"Bitcoin"` or `"Euro"`.
+- `ticker: string`: the ticker name in exchanges and used for rates (for instance BTC or EUR).
+- `units: Unit[]`: all available `Unit` of the currency (see below what is a Unit). by convention, [0] is the default and have "highest" magnitude.
+- `disableCountervalue?: boolean`: if available, this field express that we should assume there is not valable countervalue for this coin. (because either we can't trade it or it's not available enough to assume enabling its value)
+- `symbol?: string`: if available, it's a short symbol for the coin. _in practice it's currently not used and might be dropped._
+
+#### Unit
+
+```js
+type Unit = {
+  // display name of a given unit (example: satoshi)
+  name: string,
+  // string to use when formatting the unit. like 'BTC' or 'USD'
+  code: string,
+  // number of digits after the '.' in context of this unit.
+  magnitude: number,
+  // should it always print all digits even if they are 0 (usually: true for fiats, false for cryptos)
+  showAllDigits?: boolean
+};
+```
+
+A unit describes a given representation of a currency for humans. A currency can have many units, for instance we can assume Euro have euros and cents. We can define Bitcoin to have: bitcoin, mBTC, bit, satoshi (but that's up to us really).
+
+There are however two essential property we must respect:
+
+- the first unit (`[0]`) in an array of unit is the **highest unit**, typically the default and commonly used unit (euro, bitcoin)
+- the last unit must have a **magnitude of 0**: it is the smallest unit which in fine determine the magnitude of the coin (the most atomic unit representable in this currency) and drive integers representation.
+
+Here is an example of Bitcoin units:
+
+```js
+[
+  {
+    name: "bitcoin",
+    code: "BTC",
+    magnitude: 8
+  },
+  {
+    name: "mBTC",
+    code: "mBTC",
+    magnitude: 5
+  },
+  {
+    name: "bit",
+    code: "bit",
+    magnitude: 2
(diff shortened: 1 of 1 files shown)


docs/currency.md
```diff
@@ -0,0 +1,159 @@
+# The Currency models
+
+Ledger Live deals with different kind of currencies and assets: fiat currencies, crypto-currencies and tokens.
```
juan-cortes:
```suggestion
Ledger Live deals with different kinds of currencies and assets: fiat currencies, cryptocurrencies, and tokens.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+They all share similar concept but have specifics.
```
juan-cortes:
```suggestion
They all share similar concepts but have specifics.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+
+## The types
+
+The types are defined in [types/currencies](../src/types/currencies.js).
+
+live-common exposes 3 different types: `CryptoCurrency`, `TokenCurrency` and `FiatCurrency`.
+It also exposes a union type `Currency` that is either of these 3 types. _Make sure you use the correct type, sometimes you want the union type Currency, sometimes you are really in context of a CryptoCurrency_.
```
juan-cortes:
```suggestion
It also exposes a union type `Currency` that is either of these 3 types. _Make sure you use the correct type, sometimes you want the union type Currency, sometimes you are really in the context of a CryptoCurrency_.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+
+To ease the differentiation of types, live-common data also introduces a _discriminator field_, `.type`, which have the same name of the type.
```
juan-cortes:
```suggestion
To ease the differentiation of types, live-common's data also introduces a _discriminator field_, `.type`, which has the same name as the type.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+
+### Common fields
+
+Essentially all currencies share this common ground:
+
+- `name: string`: the display name of a currency. For instance `"Bitcoin"` or `"Euro"`.
+- `ticker: string`: the ticker name in exchanges and used for rates (for instance BTC or EUR).
+- `units: Unit[]`: all available `Unit` of the currency (see below what is a Unit). by convention, [0] is the default and have "highest" magnitude.
```
juan-cortes:
```suggestion
- `units: Unit[]`: all available `Unit` of the currency (see below what is a Unit). by convention, [0] is the default and has the "highest" magnitude.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+- `disableCountervalue?: boolean`: if available, this field express that we should assume there is not valable countervalue for this coin. (because either we can't trade it or it's not available enough to assume enabling its value)
```
juan-cortes:
```suggestion
- `disableCountervalue?: boolean`: if available, this field expresses that we should assume there is not valid counter value for this coin. (because either we can't trade it or it's not available enough to assume enabling its value)
```

I think 🤷‍♂

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+  // string to use when formatting the unit. like 'BTC' or 'USD'
+  code: string,
+  // number of digits after the '.' in context of this unit.
+  magnitude: number,
+  // should it always print all digits even if they are 0 (usually: true for fiats, false for cryptos)
+  showAllDigits?: boolean
+};
+```
+
+A unit describes a given representation of a currency for humans. A currency can have many units, for instance we can assume Euro have euros and cents. We can define Bitcoin to have: bitcoin, mBTC, bit, satoshi (but that's up to us really).
```
juan-cortes:
```suggestion
A unit describes a given representation of a currency for humans. A currency can have many units, for instance, we can assume Euro have euros and cents. We can define Bitcoin to have: bitcoin, mBTC, bit, satoshi (but that's up to us really).
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+
+There are however two essential property we must respect:
```
juan-cortes:
```suggestion
There are however two essential properties we must respect:
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+
+- the first unit (`[0]`) in an array of unit is the **highest unit**, typically the default and commonly used unit (euro, bitcoin)
```
juan-cortes:
```suggestion
- the first unit (`[0]`) in an array of units is the **highest unit**, typically the default and commonly used unit (euro, bitcoin)
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+- the last unit must have a **magnitude of 0**: it is the smallest unit which in fine determine the magnitude of the coin (the most atomic unit representable in this currency) and drive integers representation.
```
juan-cortes:
```suggestion
- the last unit must have a **magnitude of 0**: it is the smallest unit which in fine determines the magnitude of the coin (the most atomic unit representable in this currency) and drives integers representation.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+  },
+  {
+    name: "satoshi",
+    code: "sat",
+    magnitude: 0
+  }
+];
+```
+
+The field `showAllDigits` is generally unset for all crypto currencies and set for fiat currencies to true (but there are exceptions). It forces a unit to displayed all digits even if it's all zeros.
```
juan-cortes:
```suggestion
The field `showAllDigits` is generally unset for all cryptocurrencies and set for fiat currencies to true (but there are exceptions). It forces a unit to display all digits even if it's all zeros.
```

docs/currency.md
```diff
@@ -0,0 +1,159 @@
...
+
+We don't want showAllDigits for currencies because they generally have a lot of magnitude, for instance Ethereum have 18 which would turn `ETH 42.100000000000000000`. Instead we want `ETH 42.1`.
```
juan-cortes:
```suggestion
We don't want showAllDigits for currencies because they generally have a lot of magnitudes, for instance, Ethereum has 18 which would turn `ETH 42.100000000000000000`. Instead, we want `ETH 42.1`.
```

PR #488, opened 2020-02-01
throwing ManagerDeviceLockedError when device is unresponsive during socket

gre:
This fixes an issue in timeout of bulk and also will throw ManagerDeviceLockedError if device is locked during the initial socket exchange.

The goal will be later to handle specific interruption cases (for now just locked mode of the device) during a series of apps Install/Uninstall/Update in a way that does not completely interrupt everything (queue of install/uninstalls) but will handle recovering from that state.

This only will be covered in a second PR but we first need this PR in to fixes timeout issue.

### DIFF
--- a/src/api/socket.js
+++ b/src/api/socket.js
@@ -19,7 +19,9 @@ import type { SocketEvent } from "../types/manager";
 const warningsSubject = new Subject();
 export const warnings: Observable<string> = warningsSubject.asObservable();
 
-const UNRESPONSIVE_DELAY = 1000;
+const ALLOW_MANAGER_DELAY = 500;
+const UNRESPONSIVE_DELAY = 3000;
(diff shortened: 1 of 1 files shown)


gre:
*quick draft idea for the next PR to come:*

I was thinking we can introduce some kind of `state.userActionError` in app state, that if present, will require the UI to implement a modal that ask user something (like please unlock your device if error is ManagerDeviceLockedError). maybe it could just be `state.isLocked` but I anticipate we might want to do the same even for unplugging a device in middle of an update? in any case, the logic (of live-common) would need to automatically detect the unlock and recover & continues from it.

the thing is the apdu is blocked with the device, so maybe we want to globally listen to that to know when something hangs and to know when it is "released". this is a general concept we could introduce on transport and in order to prevent race condition.

the current implementation will close the socket if the device hangs after 3s, i think it's a good practice because in that sleep mode case, we're liking going to have a timeout from server anyway so we better "release" that socket, ask user to unlock, and redo the action again later.

PR #499, opened 2020-02-07
Introduce starred field on Account

gre:
We are introducing the field `starred` on all Account likes.

it's better this is collocated with the account data:
- simpler to use and set (like we do for `name`)
- we don't have to deal with removing the stars when account is removed
- it gets into the encryption too (storring ids separately need encryption as ids contains xpub)


gre:
@juan-cortes we could, but i kinda wanted this to be in a new "major" as part as some other siblings PRs we have that deserve to be a major bump. we also are about to have a new libcore node bindings that will be breaking, so i'm waiting it to make a big "super bump" with a PR in lld-v2 🙏

PR #511, opened 2020-02-13
LL-2141 Parse stringified error object

LukeSugiura:
## Result
Now it shows human readable error message consistantly accross accounts.
Screen Shot 2020-02-13 at 4 39 30 PM


src/libcore/syncAccount.js
```diff
@@ -72,8 +71,7 @@ export async function syncCoreAccount({
 
     return account;
   } catch (e) {
-    if (e.name !== "Error") throw remapLibcoreErrors(e);
-    throw new SyncError(e.message);
+    throw remapLibcoreErrors(e);
```
gre: I would recommend this:

```js
const betterError = remapLibcoreErrors(e);
if (betterError.name !== "Error") throw betterError; // we found a better error to remap to
throw new SyncError(betterError.message); // otherwise we fallback into a SyncError
```

i think the initial idea (that was later broken) is that if we can't find a "precise" error (which maps to a wording, like a no network case) we will fallback into a "SyncError" (that have a "good" wording for the user)

LukeSugiura:
@gre I pulled from master on this repo and linked it to the desktop app. Now the desktop app doesn't start. Do you aware of this issue?

LukeSugiura:
Confirmed it's working with https://github.com/LedgerHQ/lld-v2/tree/live-common-12

LukeSugiura:
@gre I tried with https://github.com/LedgerHQ/ledger-live-mobile/pull/1151
Are these look okay to you?

Screen Shot 2020-02-14 at 12 48 42 PM


Screen Shot 2020-02-14 at 12 48 40 PM

gre:
👌

PR #515, opened 2020-02-16
Documenting bridges

gre:


docs/AccountBridge.md
```diff
@@ -0,0 +1,318 @@
+# `AccountBridge`: synchronize an account and perform transaction
+
+> - Prerequisite: [Account](./account.md).
+> - See also: [gist: transaction with a Ledger device](./docs/gist-tx.md)
+
+CurrencyBridge offers a generic abstraction to synchronize accounts and perform transactions.
```
juan-cortes: AccountBridge?

docs/AccountBridge.md
```diff
@@ -0,0 +1,318 @@
...
+  broadcast({
+    account: Account,
+    signedOperation: SignedOperation
+  }): Promise<Operation>;
+}
+```
+
+## `getAccountBridge`
+
+`getCurrencyBridge` is the entry point to use to get a bridge instance for a given account.
```
juan-cortes: same

docs/AccountBridge.md
```diff
@@ -0,0 +1,318 @@
...
+  initialAccount: Account,
+  syncConfig: SyncConfig
+) => Observable<(Account) => Account>
+```
+
+The first parameter `initialAccount` is the account object as the frontend knows it. It will be used to determine what needs to be refreshed in the most minimal way possible (sync tries to preserve object references).
+
+The second parameter allows to configure the synchronization (similarly to `CurrencyBridge#scanAccounts`). See [SyncConfig](../src/types/pagination.js).
+
+**The returned value is an `Observable` of updater function (`Account=>Account`). Let's dig a bit in this concept:**
```
juan-cortes: nice

PR #517, opened 2020-02-17
Support TypeScript on CLI

LukeSugiura:
`yarn build` command both transpile flow and compile typescript files with babel.
`yarn type-check` command checks flow with outputs and typescript without emitting files.


gre:
😍

meriadec:
:rocket:

PR #539, opened 2020-02-28
resume mobile testing on ci

valpinkman:


gre:
we'll have to reboot when it works

PR #543, opened 2020-03-02
appConnect and ManagerConnect commands to hw

valpinkman:


src/hw/actions/shared.js
```diff
@@ -0,0 +1,29 @@
...
+  modelId: DeviceModelId
+};
+
+export type Action<Request, HookState, Result> = {|
+  useHook: (?Device, Request) => HookState,
+  mapResult: HookState => ?Result
+|};
+
+// emit value each time it changes by reference. it replays the last value at subscribe time.
+export function useReplaySubject<T>(value: T): ReplaySubject<T> {
```
gre: could we move this into a src/observable.js maybe ?

valpinkman: ok

PR #545, opened 2020-03-03
Bootstrap derivation.md

gre:


docs/derivation.md
```diff
@@ -0,0 +1,106 @@
...
+```
+
+where
+
+- `purpose` will be 44 for legacy, 49 for segwit and 84 for native segwit.
+- `coin_type` is a number unique per crypto currency and specified in [SLIP 44](https://github.com/satoshilabs/slips/blob/master/slip-0044.md). Typically, Bitcoin is `0`.
+- `account` is the index of the account
+- `change` and `address_index` are ways to generate new addresses inside the account. This follow a strict rule described in the spec (with the importance of the Gap limit)
+
+So a typically derivation is
```
valpinkman: typical*

docs/derivation.md
```diff
@@ -0,0 +1,106 @@
...
+
+**There is a very important specification in BIP 44:** when a client (like Ledger Live / live-common) will search for accounts, it needs to start from account=0 and increment account until finding an empty account. By empty, it means an account without transaction received on its addresses (with respect to Gap limit).
+
+So for instance, if `44'/0'/0'/0/0` is empty, by following the spec, it's assumed that no addressed under the `44'/0'/0'` account has been filled with transaction: the account is considered empty and it's therefore the only available Bitcoin Legacy account that should be exposed to the user. In other words, allowing the user to create the account `44'/0'/1'` would be considered illegal regarding this spec.
+
+This is a strict limit that allows the scanning algorithm that search for accounts to terminate.
+
+Now, for Bitcoin, we have more than one "purpose": there were important upgrade in how to format addresses and make transactions on the Bitcoin network that required to move to a fundamentally different account: this is why Segwit accounts are going to be on `49'/0'/account'` and Native Segwit are going to be on `84'/0'/account'`.
```
valpinkman: upgrades*

docs/derivation.md
```diff
@@ -0,0 +1,106 @@
...
+
+We have also some special "custom" accounts (by enabling _Experimental Settings > Extended account search_) that would enable more derivation modes:
+
+```
+[ "legacy_on_bch", "segwit_on_legacy", "legacy_on_segwit", "", "segwit", "native_segwit" ]
+```
+
+`segwit_on_legacy` and `legacy_on_segwit` are ways to recover from classical user mistake of having sent funds on a Segwit addressed derived on a legacy account, or reversely. That way allows these users to create (in advanced mode way) the accounts to recover the funds.
+
+The returned value really different for each coin, because some coins supports segwit some don't, same for native segwit. Some coins also are **hard fork of Bitcoin** which force us to support a special derivation (we called `"unsplit"`, because it's the derivation path before splitting the funds out of the parent derivation path).
```
valpinkman: either "The returned values *are* really different for each coin" or "The returned values really differ for each coin"

PR #552, opened 2020-03-04
LL-2264 Allow accounts with ChildAccount to export operations

juan-cortes:
image

Investigating https://ledgerhq.atlassian.net/browse/TECSUPPORT-579 we found that while we were handling `TokenAccount` type for operation exports, we weren't doing this for `ChildAccount` type. Meaning the export operations has been broken for Tezos with children accounts since the beginning of time.

### How to test this
- Try to export a Tezos account with child accounts on develop branch. This should ask for the destination file but not produce any actual file. There'll be some traces on the console if it's open. 
- Try to do the same on this branch, it should generate the csv file correctly.

### DIFF
--- a/src/csvExport.js
+++ b/src/csvExport.js
@@ -74,7 +74,7 @@ const accountRows = (
 const accountsRows = (accounts: Account[]) =>
   flattenAccounts(accounts).reduce((all, account) => {
     const parentAccount =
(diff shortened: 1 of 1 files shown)


gre:
QA will be done when we bump live-common on the other side.

PR #558, opened 2020-03-09
Implement BridgeSync in a generic way

gre:
- generalized for mobile and desktop ( https://github.com/LedgerHQ/ledger-live-desktop/pull/2697 )
- more optimized with internal states and hooks

---

QA will be done in context of PRs i'll then send to desktop and mobile. I already wrote desktop and it should work the same.

Testing this is a tricky topic and for now we'll have no test on live-common itself.


src/account/helpers.js
```diff
@@ -138,3 +139,28 @@ export const shortAddressPreview = (addr: string, target: number = 20) => {
...
 };
+
+export const isUpToDateAccount = (account: ?Account) => {
+  if (!account) return true;
+  const { lastSyncDate, currency } = account;
+  const { blockAvgTime } = currency;
+  if (!blockAvgTime) return true;
+  const outdated =
+    Date.now() - (lastSyncDate || 0) >
+    blockAvgTime * 1000 + getEnv("SYNC_OUTDATED_CONSIDERED_DELAY");
```
juan-cortes: Why `*1000`?
https://github.com/LedgerHQ/ledger-live-common/blob/a60b6766654436699628469b39482c9f8bbb4d9a/src/types/currencies.js#L65

src/bridge/react/BridgeSync.js
```diff
@@ -0,0 +1,384 @@
...
+    const schedule = (ids: string[], priority: number) => {
+      if (priority < skipUnderPriority.current) return;
+      // by convention we remove concurrent tasks with same priority
+      // FIXME this is somehow a hack. ideally we should just dedup the account ids in the pending queue...
+      syncQueue.remove(o => priority === o.priority);
+      log("bridge", "schedule " + ids.join(", "));
+      syncQueue.push(ids, -priority);
+    };
+
+    // don't always sync in same order to avoid potential "never account never reached"
```
juan-cortes:
```suggestion
    // don't always sync in the same order to avoid potential "account never reached"
```

src/account/helpers.js
juan-cortes: Either it's not `*1000` or the comment on the field should be `// in milliseconds`

gre: 🤔

PR #566, opened 2020-03-18
Fix getTrustedInput to support Zencash raw transactions with extra data

KhalilBellakrid:
This is related to : https://ledgerhq.atlassian.net/browse/TECSUPPORT-356

Thanks @juan-cortes for the help and to add and expose this useful command for future debug sessions of the infamous "Missing Inputs" errors 😄 

**HODL:** we should wait for https://github.com/LedgerHQ/ledgerjs/pull/472 to be merged, bump `ledgerjs` version and test the whole thing again !


cli/src/commands/hash.js
```diff
@@ -0,0 +1,56 @@
...
+  if (!hex) return `Backend returned no hex for this hash`
+
+  const tx = btc.splitTransaction(
+    hex,
+    currency.supportsSegwit,
+    currency.hasTimestamp,
+    true
+  );
+
+  const outHash = await btc.getTrustedInput(transport, 0, tx, ['zencash']);
```
gre: shouldn't it be `currency.id` here?

cli/src/commands-index.js
```diff
@@ -65,6 +66,7 @@ export default {
   repl,
   send,
   signMessage,
+  hash,
```
gre: i recommend a more specific name for this command, like `testZencashHash` or something. on the long term i'm not sure if it makes so much sense to have this command here, but maybe we could made a test later based on this idea.

src/types/currencies.js
```diff
@@ -62,6 +62,7 @@ export type CryptoCurrency = CurrencyCommon & {
   // used for UI
   color: string,
   family: string,
+  hasTimestamp?: boolean,
```
gre: probably should be moved into `bitcoinLikeInfo` level ?

src/families/bitcoin/libcore-signOperation.js
```diff
@@ -59,6 +59,8 @@ async function signTransaction({
     additionals.push("decred");
   } else if (currency.id === "stealthcoin") {
     additionals.push("stealthcoin");
+  } else if (currency.id === "zencash") {
+    additionals.push("zencash");
   }
```
gre: maybe for later, but i think we could replace some of these ifs by `additionals.push(currency.id)`

gre:
left some comment. you should also be able to `yarn add @ledgerhq/hw-app-btc` to update the version on both the root and the cli/ folder.

cli/src/commands-index.js
KhalilBellakrid: This command is generic, but we could rename it `testGetTrustedInputFromTxHash`, WDYT ?

gre: good to me!

cli/src/commands/hash.js
```diff
@@ -0,0 +1,56 @@
...
+import { deviceOpt } from "../scan";
+import { from } from "rxjs";
+import axios from "axios";
+import Btc from "@ledgerhq/hw-app-btc";
+import { findCurrencyExplorer } from "@ledgerhq/live-common/lib/api/Ledger";
+import { findCryptoCurrencyById } from "@ledgerhq/live-common/lib/data/cryptocurrencies";
+
+const command = async (transport, currencyId, hash) => {
+  const btc = new Btc(transport);
+  const currency = findCryptoCurrencyById(currencyId);
```
gre: (minor) you can also add an `invariant(currency.family==="bitcoin", "currency of bitcoin family only")`

gre:
`yarn prettier` for the test to pass

PR #571, opened 2020-03-21
Add AccountBridge#estimateMaxSpendable

gre:
(from documentation that is also in this PR)

---

### `estimateMaxSpendable`
```js
estimateMaxSpendable({
  account: AccountLike,
  parentAccount?: ?Account,
  transaction?: ?Transaction
}): Promise<BigNumber>;
```
is an heuristic that provide the **estimated** max amount that can be set to a send.
This is often the balance minus the fees, but there are exceptions that depends between coins (reserve, burn, frozen part of the balance,...).
By "estimated", it means this is not necessarily correct and there might be a delta that is +- the actual reality (meaning trying to send with this amount can either exceed the spendable or leave some dust). It should not be used to do an actual SEND MAX (there is `useAllAmount` for this) but it can be used for an informative UI or also for "dry run" approaches (try a transaction with that, and refine).
The strategy of implementations should be to prefer pessimistic estimation rather than potentially impossible transaction, in other words implementation should consider the worse case scenario (like sending all UTXOs to a legacy cost more fees to it have less spendable)
the parameters `account` and `parentAccount` are the regular account parameter for normal account and token accounts.
The parameter `transaction` allows to refine the estimation to a better precision. If provided, implementation need to be accurate and try to have an actual SEND MAX value. If information of `transaction` are partial, it can still help the calculation but it's not guaranteed.


docs/AccountBridge.md
```diff
@@ -279,6 +285,28 @@ broadcast({
...
+
+```js
+estimateMaxSpendable({
+  account: AccountLike,
+  parentAccount?: ?Account,
+  transaction?: ?Transaction
+}): Promise<BigNumber>;
+```
+
+is an heuristic that provide the **estimated** max amount that can be set to a send.
```
juan-cortes:
```suggestion
is a heuristic that provides the **estimated** max amount that can be set on a send.
```

docs/AccountBridge.md
```diff
@@ -279,6 +285,28 @@ broadcast({
...
+
+This is often the balance minus the fees, but there are exceptions that depends between coins (reserve, burn, frozen part of the balance,...).
```
juan-cortes:
```suggestion
This is often the balance minus the fees, but there are exceptions that depend between coins (reserve, burn, frozen part of the balance,...).
```

docs/AccountBridge.md
```diff
@@ -279,6 +285,28 @@ broadcast({
...
+
+By "estimated", it means this is not necessarily correct and there might be a delta that is +- the actual reality (meaning trying to send with this amount can either exceed the spendable or leave some dust). It should not be used to do an actual SEND MAX (there is `useAllAmount` for this) but it can be used for an informative UI or also for "dry run" approaches (try a transaction with that, and refine).
+
+The strategy of implementations should be to prefer pessimistic estimation rather than potentially impossible transaction, in other words implementation should consider the worse case scenario (like sending all UTXOs to a legacy cost more fees to it have less spendable)
```
juan-cortes:
```suggestion
The strategy of implementations should be to prefer pessimistic estimation rather than a potentially impossible transaction, in other words, an implementation should consider the worst-case scenario (like sending all UTXOs to a legacy address has higher fees resulting in a lower max spendable)
```

docs/AccountBridge.md
```diff
@@ -279,6 +285,28 @@ broadcast({
...
+
+the parameters `account` and `parentAccount` are the regular account parameter for normal account and token accounts.
+
+The parameter `transaction` allows to refine the estimation to a better precision. If provided, implementation need to be accurate and try to have an actual SEND MAX value. If information of `transaction` are partial, it can still help the calculation but it's not guaranteed.
```
juan-cortes:
```suggestion
The parameter `transaction` allows refining the estimation to better precision. If provided, an implementation needs to be accurate and try to have an actual SEND MAX value. If the information on `transaction` is partial, it can still help the calculation but it's not guaranteed.
```

src/families/bitcoin/publicAddresses.js
```diff
@@ -0,0 +1,36 @@
+// @flow
+/**
+ * These addresses are PUBLIC addresses
+ * We use them for tests and also for dry-run estimations
+ * DO NOT USE AS RECIPIENT OR SIGN TRANSACTION INTO
```
juan-cortes:
```suggestion
 * DO NOT USE AS RECIPIENT OR SIGN TRANSACTIONS INTO THEM
```
Unless I'm mistaken, do you mean "sign transactions _to these accounts_" or something else?

src/families/bitcoin/publicAddresses.test.js
```diff
@@ -0,0 +1,17 @@
+// @flow
+
+import { abandonSeedLegacyPerCurrency } from "./publicAddresses";
+import { listCryptoCurrencies } from "../../currencies";
+
+const ignore = ["bitcloud", "bitcore", "bitsend", "clubcoin", "megacoin"];
+
+test("all bitcoin forks that have a manager app have a defined address in abandonSeedLegacyPerCurrency", () => {
```
juan-cortes: impressive

src/families/ripple/bridge/libcore.js
```diff
@@ -133,6 +134,22 @@ const prepareTransaction = async (a, t) => {
...
+  account,
+  parentAccount,
+  transaction
+}) => {
+  const mainAccount = getMainAccount(account, parentAccount);
+  const t = await prepareTransaction(mainAccount, {
+    ...createTransaction(),
+    ...transaction,
+    useAllAmount: true,
+    recipient: "rHsMGQEkVNJmpGWs8XUBoTBiAAbwxZN5v3" // public testing seed abandonx11,about
```
juan-cortes: If all currencies will have this, wouldn't it make sense to have this in `cryptocurrencies.js` in case we want to reuse it somewhere else?

src/types/bridge.js
```diff
@@ -85,6 +87,19 @@ export interface AccountBridge<T: Transaction> {
     transaction: T
   ): Promise<TransactionStatus>;
 
+  // heuristic that provide the estimated max amount that can be set to a send.
```
juan-cortes:
```suggestion
  // heuristic that provides the estimated max amount that can be set to a send.
```

src/types/bridge.js
```diff
@@ -85,6 +87,19 @@ export interface AccountBridge<T: Transaction> {
...
+  // this is usually the balance minus the fees, but it really depends between coins (reserve, burn, frozen part of the balance,...).
+  // it is an heuristic in that this is not necessarily correct and it can be +-delta (so the info can exceed the spendable or leave some dust).
```
juan-cortes:
```suggestion
  // it is a heuristic in that this is not necessarily correct and it can be +-delta (so the info can exceed the spendable or leave some dust).
```

src/types/bridge.js
```diff
@@ -85,6 +87,19 @@ export interface AccountBridge<T: Transaction> {
...
+  // it's used as informative UI and also used for "dry run" approaches, but it shouldn't be used to determine the final SEND MAX amount.
+  // it returns an amount in the account unit
+  // if a transaction is provided, it can be used to precise the information
+  // if it not provided, you can assume to take the worse case scenario (like sending all utxo to a legacy cost more fees to it have less spendable)
```
juan-cortes:
```suggestion
  // if it not provided, you can assume to take the worst-case scenario (like sending all UTXOs to a legacy address has higher fees resulting in a lower max spendable)
```

src/families/ripple/bridge/libcore.js
gre: I think `cryptocurrencies.js` is meant to be general data of cryptos, here it feels a bit specific to our internal algorithm of dry run estimation. not all bridge implementation & currency will need real addresses (e.g. ETH can just use 0x000...0000).

it's too implementation-specific at the moment to consider generalisation imo.

juan-cortes: 👍

PR #575, opened 2020-03-26
LL-2323 Fix portfolio year delta calculation

juan-cortes:
image

### Description of the issue
We have a way of calculating the delta for an account and a given range that disregards the left dates until the first one that is greater than zero. This works great for individual accounts since it will allow the user to see a delta that starts from the moment they first added funds to the accounts.

However, when we apply the same logic to a portfolio view we are essentially calculating the delta from an existing data point which combines all the _first date with a balance greater than zero_ for all accounts. Using the above graphic as a base we would be calculating a delta between `(A1+B1)-(A2+B2)` while we really want `A1-(A2+B2)` since that's what our _portfolio_ looks like.

### Proposed solution
I disabled the `effectiveFrom` calculation when we are in the portfolio graph mode and the associated cache to that portfolio data. This means that we will still have this for individual accounts and asset views but not for the portfolio graph, allowing for a more realistic delta.

This solution is not perfect though since, ideally, we would calculate the first date with balance for all accounts in the portfolio and use that as the left range. Since I'm currently not doing that the delta would still be off for portfolios that don't have any balance at the beginning of the year range, it will calculate the delta between 0 and the ending point **only in this case**. I think it's a fair tradeoff for now until we can improve on it further. Maybe @gre you can take a look at this, I went a bit dizzy trying to do it.

I also increased the `percentageHighThreshold` variable but perhaps @gre has a reason for it to exist in the first place. I increased it (feel free to suggest an edit if you disagree) because a balance can grow more than 100 for sure and we shouldn't hide the percentage delta in that case as we are currently doing.

### Test plan
With a desktop version running this branch it would be easy to compare, you should see that the delta on portfolio reflects the difference between the first datapoints and the last datapoints instead of an unknown and difficult calculation like it currently does. Account and asset views should remain unaffected between develop and this branch.

### DIFF
--- a/src/portfolio/index.js
+++ b/src/portfolio/index.js
@@ -93,18 +93,21 @@ type GetBalanceHistoryWithCountervalue = (
     TokenCurrency | CryptoCurrency,
     BigNumber,
     Date
-  ) => ?BigNumber
+  ) => ?BigNumber,
+  useEffectiveFrom?: boolean
 ) => AccountPortfolio;
 
 // hash the "stable" part of the histo
 // only the latest datapoint is "unstable" meaning it always changes because it's the current date.
-const accountRateHashCVStable = (account, r, cvRef) =>
-  `${accountRateHash(account, r)}_${cvRef ? cvRef.toString() : "none"}`;
+const accountRateHashCVStable = (account, r, cvRef, useEffectiveFrom) =>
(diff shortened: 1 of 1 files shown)


gre:
If it passes test we'll merge and involve QA on the target projects

juan-cortes:
If it doesn't pass tests perhaps the tests are wrong :trollface:

gre:
probably a tmp back issue

PR #578, opened 2020-03-30
Tron - implement useAllAmount

nukky94:


src/families/tron/test-dataset.js
```diff
@@ -27,6 +27,7 @@ const dataset: DatasetTest<Transaction> = {
   currencies: {
     tron: {
       FIXME_ignoreAccountFields: [
+        "tronResources.bandwidth",
```
gre: can you comment **in the code** why we need to ignore these? it's not clear to me

src/families/tron/bridge/js.js
```diff
@@ -71,6 +71,11 @@ const signOperation = ({ account, transaction, deviceId }) =>
           ? account.subAccounts.find(sa => sa.id === transaction.subAccountId)
           : null;
 
+      const balance =
+        subAccount ? subAccount.balance : account.spendableBalance;
```
gre: in case of parent account, isn't spendableBalance minus fees? that's what i did when implementing the "max spendable"

src/families/tron/test-dataset.js
nukky94: Because this account was not totally frozen but it should be the case right now.
I will delete this ignored fields and regenerate snapshots

src/families/tron/bridge/js.js
nukky94: you're right

PR #579, opened 2020-03-30
LL-2279

LFBarreto:
(Tron): added voting flow react hooks + some tests

### DIFF
--- a/src/families/tron/react.js
+++ b/src/families/tron/react.js
@@ -1,18 +1,36 @@
 // @flow
-import { useState, useEffect } from "react";
+import { useState, useEffect, useMemo, useReducer, useRef } from "react";
 import { getTronSuperRepresentatives, getNextVotingDate } from "../../api/Tron";
 
 import { BigNumber } from "bignumber.js";
 
-import type { SuperRepresentative, Vote } from "./types";
+import type { SuperRepresentative, Vote, TronResources } from "./types";
 import type { Account } from "../../types";
 
 import { getCryptoCurrencyById } from "../../currencies";
+
+export type Action = {
+  type: "updateVote" | "resetVotes" | "clearVotes",
+  address: string,
+  value: string
+};
+
+export type State = {
+  votes: { [address: string]: number }, // formatted Map of votes
+  votesAvailable: number, // total of available TP
+  votesUsed: number, // total of TP used
+  votesSelected: number, // number of SR votes selected
+  max: number, // votes remaining
+  initialVotes: { [address: string]: number } // initial Map of votes
+};
 
 const oneTrx = BigNumber(10).pow(
   getCryptoCurrencyById("tron").units[0].magnitude
 );
 export const MIN_TRANSACTION_AMOUNT = oneTrx;
+
+export const SR_THRESHOLD = 27;
+export const SR_MAX_VOTES = 5;
 
 /** Fetch the list of super representatives */
 export const useTronSuperRepresentatives = (): Array<SuperRepresentative> => {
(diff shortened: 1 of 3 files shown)


src/families/tron/react.js
```diff
@@ -68,3 +88,165 @@ export const formatVotes = (
...
+  superRepresentatives: SuperRepresentative[],
+  votes: Vote[]
+): {
+  sr: SuperRepresentative,
+  name: ?string,
+  address: string,
+  rank: number,
+  isSR: boolean
+}[] {
+  const [initialVotes] = useState(votes.map(({ address }) => address));
```
LukeSugiura: useMemo?

src/families/tron/react.js
```diff
@@ -68,3 +88,165 @@ export const formatVotes = (
...
+        name: sr.name,
+        address: sr.address,
+        rank: rank + 1,
+        isSR: rank <= SR_THRESHOLD
+      })),
+    [superRepresentatives]
+  );
+
+  const sortedVotes = useMemo(() => {
+    return SR.filter(({ address }) => initialVotes.includes(address)).concat(
```
LukeSugiura: Can omit return

src/families/tron/react.js
LFBarreto: not here i don't want it to recompute just initialize and that it :+1:

src/families/tron/react.js
```diff
@@ -68,3 +88,165 @@ export const formatVotes = (
...
+    () => (search ? SR.filter(searchFilter(search)) : sortedVotes),
+    [search, SR, sortedVotes]
+  );
+
+  return sr;
+}
+
+/** Tron vote flow state reducer */
+function votesReducer(state, action: Action) {
+  const { type, address, value } = action;
```
LukeSugiura: why not destructing directly on argument?

src/families/tron/react.js
```diff
@@ -68,3 +88,165 @@ export const formatVotes = (
...
+  const sr = useMemo(
+    () => (search ? SR.filter(searchFilter(search)) : sortedVotes),
+    [search, SR, sortedVotes]
+  );
+
+  return sr;
+}
+
+/** Tron vote flow state reducer */
+function votesReducer(state, action: Action) {
```
LukeSugiura: Can State type be used here?

src/families/tron/react.js
LukeSugiura: In that case, i would suggest useRef. useState shouldn't be used without set func.

src/families/tron/react.js
```diff
@@ -68,3 +88,166 @@ export const formatVotes = (
...
+  onChangeVotes: (votes: Vote[]) => void,
+  tronResources: ?TronResources
+): [State, (action: Action) => void] {
+  const [state, dispatch]: [State, (action: Action) => void] = useReducer(
+    votesReducer,
+    initState(votes, tronResources)
+  );
+
+  useEffect(
+    debounce(
```
gre: this makes no sense

src/families/tron/react.js
```diff
@@ -68,3 +88,166 @@ export const formatVotes = (
...
+    votesSelected: initialVotes.length,
+    max: Math.max(0, votesAvailable - votesUsed),
+    initialVotes: votes
+  };
+}
+
+/** Hook to retrieve and update voting flow state */
+export function useVotesReducer(
+  votes: Vote[],
+  onChangeVotes: (votes: Vote[]) => void,
```
gre: having both an external state and an internal state of "votes" feels like it's an antipattern; there is something to improve here. there should be no "onChangeVotes" I feel. `votes` param should be called `initialVotes` if it's truely what's happening here.

LFBarreto: initially i wanted to debounce the update of the transaction votes and have an easier to use data structure for this but updating the transaction after each update doesn't seem to make the whole thing lag anymore :man_shrugging:

gre: we should `useDebounce` then. because in your code, it's never calling onChangeVotes

LFBarreto: the useEffect with a debounce does that, it's weird but it works i'll try with the fixed use debouncedCallback on the lld side without the debouncing here :+1:

gre: ok i understand the code, but it's not optimal and it's very dangerous: if you have a lot of re-render happening, the debouncing will never be called, it's only called when there is no re-render but you don't know that

src/families/tron/react.js
gre: is there a reason why @LukeSugiura ? I used that pattern at a few places too and i didn't find a better one. useRef is not equivalent, is it?

PR #586, opened 2020-03-31
Fix listApps always returning Nano S `deviceModel` on any transport but RN BLE

MortalKastor:
`listApps` tries to get current `deviceModel` from `transport` and fallbacks to Nano S if it's not there. The issue is, AFAIK, only `react-native-hw-transport-ble` exposes a `deviceModel` in its class, so any device connected through a different transport will be detected as a Nano S.

Counterintuitively, the issue doesn't stay confined to `listApps` as `useAppsRunner` uses the results from it to seed the `initState` and contaminates the whole manager on mobile. On desktop the app doesn't use this state at the top level of the manager but only in the nested app list, thus avoiding any breakage from the bug.

This PR adds a third optional parameter to `listApps` to enforce a specific `deviceModelId`.

### DIFF
--- a/src/apps/hw.js
+++ b/src/apps/hw.js
@@ -2,6 +2,7 @@
 
 import Transport from "@ledgerhq/hw-transport";
 import { getDeviceModel } from "@ledgerhq/devices";
+import type { DeviceModelId } from "@ledgerhq/devices";
 import { UnexpectedBootloader } from "@ledgerhq/errors";
 import { Observable, throwError } from "rxjs";
 import type { Exec, AppOp, ListAppsEvent, ListAppsResult } from "./types";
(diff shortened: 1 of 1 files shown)


gre:
I think the real fix should be to fix react-native-hid to expose a  deviceModel the same way, 
i know it's not part of Transport

wdyt @MortalKastor ? can you try to fix it in ledgerjs too?

MortalKastor:
> I think the real fix should be to fix react-native-hid to expose a deviceModel the same way

I considered it, but since listApps isn't meant to only be used in React Native, the change should be made for all our transports, and I wasn't feeling like changing every one of them.

Also listApps actually barely use `deviceModel` itself, only as a fallback for computing a blocksize that doesn't look like it's used anywhere on mobile (maybe on desktop?), so maybe it would make more sense to not use it in listApps and rather pass it to `useAppsRunner`?

> i know it's not part of Transport but we probably should make it so

It makes sense for an instantiated transport object to expose `deviceModel` and would be nice to have everywhere.

Anyway, since the bug only breaks the implementation relying on `react-native-hid` right now and we probably (I'd like it) will add `deviceModel` to `Transport` in the long run, it makes sense to add it only to `react-native-hid` as a quick fix for now, so I'm doing that right away 👍

gre:
yeah I feel the fix is easier on ledgerjs and we would not have to introduce this, especially it would quickly become outdated if it's part of transport (which is the case for a bunch of transport, not all i'm afraid)

yet, i feel the real fix should be to have a way to infer device id over apdu (with the deviceInfo in fact i think we do have this info! i just don't know exactly how to do it, maybe we can ask oto, it's something with the `targetId` )

MortalKastor:
Closed in favor of LedgerHQ/ledgerjs#483

PR #591, opened 2020-04-02
Add lastWithdrawnRewardDate and lastVotedDate in tronResources

nukky94:


src/api/Tron.js
```diff
@@ -491,24 +482,12 @@ export const getTronResources = async (
   };
 
   const delegatedFrozen = {
-    bandwidth:
-      delegatedFrozenBandwidth && lastDelegatedFrozenBandwidthOp
-        ? {
-            amount: BigNumber(delegatedFrozenBandwidth),
-            expiredAt: new Date(
```
gre, line 498: mmh we seem to need it for https://github.com/LedgerHQ/ledger-live-desktop/blob/42c6e4927576e37f7c37b799671dba082586775f/src/renderer/modals/ManageTron/index.js#L126-L129 ?

nukky94: This is not the same field `delegatedFrozen` vs `frozen`.
`delegatedFrozen` is an edge weird case and represent a tron user which have frozen assets for another one tron user. But in fact we don't really care.

PR #594, opened 2020-04-02
Tron polishes

gre:


src/families/tron/bridge/js.js
```diff
@@ -576,7 +576,7 @@ const getTransactionStatus = async (
       errors.amount = useAllAmount
         ? new NotEnoughBalance()
         : new AmountRequired();
-    } else if (totalSpent.gt(balance)) {
+    } else if (amount.gt(balance)) {
```
nukky94, line 579: I'm not sure of this change 🤔 
totalSpent = amount + fees, so we need to compare the balance with it no?

gre: in the past we used to have totalSpent=amount+fees and balance=spendableBalance
in your recent changes, we have balance=spendableBalance-fees, so i now counterbalance that on the other side.

PR #599, opened 2020-04-05
LL-1382 Fix broken reconciliation in balanceHistory

juan-cortes:
image

Following on the investigations that led to disabling the libcore implementation of the balance history for an account we've identified what could be the cause of the balance history not being updated. It seems that the way we checked in https://github.com/LedgerHQ/ledger-live-common/blob/6bce8ba5c431d6d2acbe764e8f4a91e18f77e324/src/reconciliation.js#L232-L233 if a patch was needed meant that we would only update the `balanceHistory` for an account **if** and only if the balance had changed.

If the balance hadn't changed we would be left with the old balance history. This on its own should be ok, but [we are then relying on the balance history's datapoint's `date` ](https://github.com/LedgerHQ/ledger-live-common/blob/f016dcbad04f0ba8d1887a1cc3598486574f0836/src/portfolio/index.js#L134)to fetch the counter value for a given amount, meaning that the values we were showing were, in fact, the last week/month/year from the last time we synchronized our account after our balance had changed.
https://github.com/LedgerHQ/ledger-live-common/blob/f016dcbad04f0ba8d1887a1cc3598486574f0836/src/portfolio/index.js#L134
Where `calc` is the local implementation in desktop (maybe on mobile too) calling `calculateWithIntermediarySelector` with the wrong date.

I've tested this fix while keeping the Libcore's implementation enabled *and* loading the original app.json that we received from our colleague from tech-support and after syncing the account balance would automatically reflect the correct value. Meaning it wouldn't even require a cache drop or any other action by the user.

Basically, following the drawing above, the graph was using the counter value from the dates of `A` range to display the balance values of the `C` range, or something like that. Where `B` is the last time the user's balance changed (received, send, etc).

### DIFF
--- a/src/reconciliation.js
+++ b/src/reconciliation.js
@@ -10,7 +10,8 @@ import type {
   Account,
   AccountRaw,
   SubAccount,
-  SubAccountRaw
+  SubAccountRaw,
+  BalanceHistoryRawMap
 } from "./types";
 import {
   fromAccountRaw,
@@ -151,6 +152,28 @@ export function minimalOperationsBuilderSync<CO>(
   return operations;
 }
 
+const shouldRefreshBalanceHistory = (
+  balanceHistory: BalanceHistoryRawMap,
+  account: Account
+): boolean => {
+  const { week } = balanceHistory;
+  const accountWeek = account.balanceHistory && account.balanceHistory.week;
+  if (!week || !accountWeek) return true; // there is no week yet || there were no balance history yet
+
+  const [firstDate] = week[0];
+  const [, lastValue] = week[week.length - 1];
+  const { date: firstAccountDate } = accountWeek[0];
+  const { value: lastAccountValue } = accountWeek[accountWeek.length - 1];
+
+  const isSameDate = firstDate === firstAccountDate.toISOString();
+
+  return (
+    !isSameDate || // start date of the range has changed
+    lastValue !== lastAccountValue.toString() || // final balance has changed
+    week.length !== accountWeek.length // number of endpoints has changed
+  );
+};
+
 export function patchAccount(
   account: Account,
   updatedRaw: AccountRaw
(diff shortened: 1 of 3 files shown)


src/reconciliation.js
```diff
@@ -221,19 +221,29 @@ export function patchAccount(
...
-    const refreshBalanceHistory =
-      !week || // there is no week yet
-      !accountWeek || // there were no balance history yet
-      // last datapoint changes will force a refresh
-      week[week.length - 1][1] !==
-        accountWeek[accountWeek.length - 1].value.toString();
-    if (refreshBalanceHistory) {
-      next.balanceHistory = fromBalanceHistoryRawMap(updatedRaw.balanceHistory);
-      changed = true;
+    if (week && accountWeek) {
```
gre, line 228: if there is no week or account week, we still should reconciliate to "empty" the object, otherwise you can be stuck when we disable the data.

I suggest to make a function "shouldRefreshBalance" history and to return true in these case. Also please try to keep previous comments.

src/reconciliation.js
```diff
@@ -221,19 +221,29 @@ export function patchAccount(
...
+      const [lastDateRaw, lastValue] = week[week.length - 1];
+      const { date: lastAccountDate, value: lastAccountValue } = accountWeek[
+        accountWeek.length - 1
+      ];
+      const lastDate = new Date(lastDateRaw);
+
+      const isSameDay =
+        lastDate.getFullYear() === lastAccountDate.getFullYear() &&
+        lastDate.getMonth() === lastAccountDate.getMonth() &&
+        lastDate.getDate() === lastAccountDate.getDate();
```
gre, line 238: Would it work with `lastDateRaw === lastAccountDate.toISOString()` ? (Hourly is coming soon)

src/reconciliation.js
juan-cortes: 🤷‍♂ sure

juan-cortes: Solved in current code

src/reconciliation.js
juan-cortes: Solved in current code

PR #605, opened 2020-04-07
LL-2373 Import the blacklistedTokenIds array from LiveQR

juan-cortes:
I forgot we had to do this so https://github.com/LedgerHQ/ledger-live-desktop/pull/2806 will not work until we have this in mobile side.


gre:
@juan-cortes can we have test coverage for this?

PR #617, opened 2020-04-20
bootstrap mock for cosmos

gre:
it can be tested with `EXPERIMENTAL_CURRENCIES=cosmos MOCK=whatever`

### DIFF
--- a/src/families/cosmos/types.js
+++ b/src/families/cosmos/types.js
@@ -1,0 +1,130 @@
+// @flow
+
+import type { BigNumber } from "bignumber.js";
+import type {
+  TransactionCommon,
+  TransactionCommonRaw
+} from "../../types/transaction";
+
+export type CoreStatics = {};
+export type CoreAccountSpecifics = {};
+export type CoreOperationSpecifics = {};
+export type CoreCurrencySpecifics = {};
+
+export type CosmosDelegationStatus =
+  | "bonded" //  in the active set that generates rewards
+  | "unbonding" // doesn't generate rewards. means the validator has been removed from the active set, but has its voting power "frozen" in case they misbehaved (just like a delegator undelegating). This last 21 days
+  | "unbonded"; // doesn't generate rewards. means the validator has been removed from the active set for more than 21 days basically
+
+export type CosmosDelegation = {|
+  validatorAddress: string,
+  amount: BigNumber,
+  pendingRewards: BigNumber,
+  status: CosmosDelegationStatus
+|};
+
+export type CosmosResources = {|
+  delegations: CosmosDelegation[],
+  delegatedBalance: BigNumber,
+  pendingRewardsBalance: BigNumber,
+  unboundingBalance: BigNumber,
+  withdrawAddress: string
+|};
+
+export type CosmosDelegationRaw = {|
+  validatorAddress: string,
+  amount: string,
+  pendingRewards: string,
+  status: CosmosDelegationStatus
+|};
+
+export type CosmosResourcesRaw = {|
+  delegations: CosmosDelegationRaw[],
+  delegatedBalance: string,
+  pendingRewardsBalance: string,
+  unboundingBalance: string,
+  withdrawAddress: string
+|};
+
+// NB this must be serializable (no Date, no BigNumber)
+export type CosmosValidatorItem = {|
+  validatorAddress: string,
+  name: string,
+  votingPower: number, // value from 0.0 to 1.0 (normalized percentage)
+  commission: number // value from 0.0 to 1.0 (normalized percentage)
+|};
+
+export type CosmosRewardsState = {|
+  targetBondedRatio: number,
+  communityPoolCommission: number,
+  assumedTimePerBlock: number,
+  inflationRateChange: number,
+  inflationMaxRate: number,
+  inflationMinRate: number,
+  actualBondedRatio: number,
+  averageTimePerBlock: number,
+  totalSupply: number,
+  averageDailyFees: number,
+  currentValueInflation: number
+|};
+
+// by convention preload would return a Promise of CosmosPreloadData
+export type CosmosPreloadData = {
+  validators: CosmosValidatorItem[],
+  rewardsState: CosmosRewardsState
+};
+
+export type CosmosOperationMode =
+  | "send"
+  | "delegate"
+  | "undelegate"
+  | "redelegate"
+  | "claimReward"
+  | "claimAllRewards";
+
+export type NetworkInfo = {|
+  family: "cosmos",
+  fees: BigNumber
+|};
+
+export type NetworkInfoRaw = {|
+  family: "cosmos",
+  fees: string
+|};
+
+// TODO suggest to rename to CosmosValidatorInput?
+export type CosmosValidator = {
+  address: string,
+  amount: BigNumber
+};
+
+export type CosmosValidatorRaw = {
+  address: string,
+  amount: string
+};
+
+export type Transaction = {|
+  ...TransactionCommon,
+  family: "cosmos",
+  mode: CosmosOperationMode,
+  networkInfo: ?NetworkInfo,
+  fees: ?BigNumber,
+  gasLimit: ?BigNumber,
+  memo: ?string,
+  validators: ?(CosmosValidator[]),
+  cosmosSourceValidator: ?string
+|};
+
+export type TransactionRaw = {|
+  ...TransactionCommonRaw,
+  family: "cosmos",
+  mode: CosmosOperationMode,
+  networkInfo: ?NetworkInfoRaw,
+  fees: ?string,
+  gasLimit: ?string,
+  memo: ?string,
+  validators: ?(CosmosValidatorRaw[]),
+  cosmosSourceValidator: ?string
+|};
+
+export const reflect = (_declare: *) => {};
--- a/src/account/helpers.js
+++ b/src/account/helpers.js
@@ -168,6 +168,8 @@ export const getVotesCount = (
     return isAccountDelegating(account) ? 1 : 0;
   } else if (mainAccount.tronResources) {
     return mainAccount.tronResources.votes.length;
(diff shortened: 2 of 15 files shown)


src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+export type CoreStatics = {};
+export type CoreAccountSpecifics = {};
+export type CoreOperationSpecifics = {};
+export type CoreCurrencySpecifics = {};
+
+export type CosmosDelegation = {|
+  validatorAddress: string,
+  amount: BigNumber,
+  pendingRewards: BigNumber,
+  isActive: boolean, // TBD is there more status case? if so, switch to status field with an enum string.
```
gagbo, line 18: A Delegation is always "active" ; if `isActive` means "generatesRewards" here (what I suspect), then `isActive` is only depending on the validator state.

Technically there are 3 statuses 
- "bonded" : in the active set that generates rewards
- "unbonding" : doesn't generate rewards
- "unbonded" : doesn't generate rewards.

I'm okay with the boolean here, we can discuss it further on other channels

src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+|};
+
+export type CosmosRedelegation = {|
```
gagbo, line 21: On the network level, redelegations are stored as `(delegator,sourceValidator,destValidator) -> [(amount, completionTime)....]`: the `completionTime,Amount` doesn't have to be unique for a `(delegator, sourceValidator, destValidator)`. It's okay if we unroll the list into a list of this type, we just have to be sure to catch it

src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+  pendingRewardsBalance: string, 
+  unboundingBalance: string,
+  withdrawAddress: string,
+|};
+
+// NB this must be serializable (no Date, no BigNumber)
+export type CosmosValidatorItem = {|
+  validatorAddress: string,
+  name: string,
+  votingPower: number, // value from 0.0 to 1.0 (normalized percentage)
```
gagbo, line 78: Slight warning, we'll have to check how to compute this percentage, it might not be easy

src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+  commission: number, // value from 0.0 to 1.0 (normalized percentage)
+|};
+
+// by convention preload would return a Promise of CosmosPreloadData
+export type CosmosPreloadData = {
+  validators: CosmosValidatorItem[]
```
gagbo, line 84: We can add a `RewardsState` object or something to hold the data necessary for reward rate computation.

We need :
Network parameters : 
- target bonded ratio
- community pool commission
- assumed time per block
- inflation rate change
- inflation max rate
- inflation min rate

Network dynamic values : 
- actual bonded ratio
- average time per block
- total supply of ATOM
- average daily fees
- current value of inflation on the network

(The only other value is the validator commission that we have in the validators array).

We'll just mock all these data to start, and I'll help with the js implementation later

src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+export type CosmosOperationMode =
+  | "send"
+  | "delegate"
+  | "undelegate"
+  | "redelegate"
+  | "claimReward"
+  | "claimAllRewards";
```
gagbo, line 93: There's no `claimAllRewards`, nor will we support the REST API feature for that ; at least for the moment. I suppose it can stay there because it's a model and we don't want to change it too often, I'm just pointing this out.

If you want to support all types of operations on the network the list will have to be bigger, we can discuss it later (Basically, we'd need one OperationMode per CosmosMsgType here : https://github.com/gagbo/lib-ledger-core/blob/18e81d2626bd26b2a9450ee7fecad805d76edeb7/core/idl/wallet/cosmos/messages.djinni#L126-L141 - without the validator-specific messages I suppose)

src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+
+export type NetworkInfoRaw = {|
+  family: "cosmos",
+  fees: string
+|};
+
+// TODO suggest to rename to CosmosValidatorInput?
+export type CosmosValidator = {
+  address: string,
+  amount: BigNumber
```
gagbo, line 108: What is this amount supposed to represent ?

src/families/cosmos/types.js
```diff
@@ -0,0 +1,140 @@
...
+};
+
+export type CosmosValidatorRaw = {
+  address: string,
+  amount: string
+};
+
+export type Transaction = {|
```
gagbo, line 116: In Cosmos, a Transaction is a list of messages (each with its own `mode`), wrapped into `memo`, `fees`, `accountNumber`, `sequence`.... elements.

So storing a Transaction with this model would only work for uni-message transactions, and that would not work for multiple message transactions like [this one](https://hubble.figment.network/cosmos/chains/cosmoshub-3/transactions/0DBFC4E8E9E5A64C2C9B5EAAAA0422D99A61CFC5354E15002A061E91200DC2D6) (WithdrawRewards + WithdrawCommission messages ; don't mind these message types, WithdrawCommission is Validator-specific)

src/families/cosmos/types.js
gre: Ok! let's aim for something that is closer to technical reality then.

src/families/cosmos/types.js
gre: ok. we'll try and if it's a problem we'll remove that feature.

src/families/cosmos/types.js
gre: i'm not sure to really understand this part. @henri-ly can you help me reiterating on the type here? thanks!

src/families/cosmos/types.js
gre: ok. @henri-ly could you help me adding the necessary parts?

src/families/cosmos/types.js
gre: ok! this was taken from the current wip work done but we can remove that one on both side then.

src/families/cosmos/types.js
gre: this was part of the existing stuff. cc @henri-ly 
this is user's input, how much he set to vote. that's why i suggested to rename CosmosValidatorInput

IDK a good wording btw, shouldn't it be like "CosmosVote" ?

src/families/cosmos/types.js
gre: i don't understand your feedback. this is exactly from your current ongoing implementation. cc @henri-ly .
A Transaction here is a UX concept which holds all the state the UI need (user inputs, other volatile infos,..)

src/families/cosmos/types.js
gagbo: Oh ok, then I think the amount is the voting power of the validator, because that information is already in `CosmosDelegation` (At some point we'll have to decide if we follow the naming schemes of the coins, or our own naming scheme that we document somewhere. `TronVote` = `CosmosDelegation` and `CosmosVote` is kind of confusing because that's another feature that's uncovered here)

src/families/cosmos/types.js
gagbo: Oh, alright. I haven't been looking too much the code in `live-common` for a while. I suppose we manually unwrap operations then, sorry for the noise

src/families/cosmos/types.js
gagbo: We'll have to store either this or the votingPower in it's raw, bigNumber form anyway. It would be nicer to have the percentage, but I just don't know how to get it from the top of my head right now so it's more a "marker" to remember later to check this

src/families/cosmos/types.js
gagbo: For the record :
- unbonding means the validator has been removed from the active set, but has its voting power "frozen" in case they misbehaved (just like a delegator undelegating). This last 21 days
- unbonded means the validator has been removed from the active set for more than 21 days basically

src/families/cosmos/types.js
gagbo: @henri-ly basically I'm saying : "If we're flattening the `entries` part of redelegation, no problem, but we must remember that this flattening must happen`

src/families/cosmos/types.js
henri-ly: It's what @gre said it's the user input of how much we are going to delegate and to who.
it works like this : 

> --cosmosValidator addr1, addr2 --cosmosAmountValidator 0.33, 1.322

wasn't really inspired for the name and if we can change to something clearer i'm in !

src/families/cosmos/types.js
henri-ly: we received this data from libcore in this kind of shape : 
`{
delegator: string,
sourceValidator: string,
destValidator: string,
entries: [{
 amount: number,
 completionTime: date
}]
}`

I choose to flatten it to simplify the UI part, so we will have to flatten when we received it from libcore

src/families/cosmos/types.js
henri-ly: Sorry it was a try to test all the pending rewards, it will be removed !

src/families/cosmos/types.js
henri-ly: I will come with something this afternoon

src/families/cosmos/types.js
gre: ok thanks! let's keep this discussion in mind for future iteration. i'll take a safe path for now which is to not introduce yet the redelegations array until we are sure about the data model and also how design planned to address it

we'll come back to it in next days/weeks

src/families/cosmos/types.js
gre: what is the best wording for such "vote" that goes into a delegation transaction? curious about COSMOS' semantic here 🤔

PR #621, opened 2020-04-21
Add delegated resources list

fbsobreira:
- Add fetch account delegated resources call
- Add delegated resources to Tron account info

### DIFF
--- a/src/api/Tron.js
+++ b/src/api/Tron.js
@@ -285,7 +285,11 @@ export const getAccountDelegatedResourcesDetails = async (
   );
 
   const resources: DelegatedResources[] = [];
-  if (result && result.delegatedResource &&  result.delegatedResource.length === 1) {
+  if (
+    result &&
+    result.delegatedResource &&
+    result.delegatedResource.length === 1
+  ) {
     const {
       frozen_balance_for_bandwidth,
       frozen_balance_for_energy,
--- a/src/api/Tron.js
+++ b/src/api/Tron.js
@@ -342,7 +342,7 @@ export const getAccountDelegatedResources = async (
           })
       );
     });
-    await Promise.all(promises);
+    await promiseAllBatched(promises);
   }
   return resources;
 };
--- a/src/api/Tron.js
+++ b/src/api/Tron.js
@@ -283,7 +283,7 @@ export const getAccountDelegatedResourcesDetails = async (
   toAddress: string
 ): Promise<DelegatedResources[]> => {
   const result = await fetch(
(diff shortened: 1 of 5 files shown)


src/families/tron/types.js
```diff
@@ -196,6 +205,7 @@ export type TronResourcesRaw = {|
     bandwidth: ?DelegatedFrozenInfoRaw,
     energy: ?DelegatedFrozenInfoRaw
   },
+  delegatedResources: DelegatedResources[],
```
gre, line 208: `DelegatedResources` is not serializable (meaning `JSON.parse(JSON.stringify(delegatedResources))` is not equivalent to `delegatedResources`) so we need to have a `DelegatedResourcesRaw` that is covered in the "{from/to}TronResourceRaw" functions.

src/api/Tron.js
```diff
@@ -275,6 +276,77 @@ export const getContractUserEnergyRatioConsumption = async (
...
+          .catch(e => {
+            log(
+              "tron-error",
+              "fetch account delegated resources fails with " + e.message,
+              { fromAddress }
+            );
+          })
+      );
+    });
+    await Promise.all(promises);
```
gre, line 345: prefer usage of our internal `promiseAllBatched` , we had bad surprise when doing too much Promise.all with tron api.

also i'm a bit scared about the performance of this, how much "toAccounts" would there be in usual accounts?

fbsobreira: Usually 1-2, some may have about 10. Maybe a dApp creator which may need to delegate energy to several contracts...

src/families/tron/types.js
fbsobreira: Ok! I will do. Thanks!

fbsobreira:
> Thanks for the contribution!
> I'm not exactly sure what this is for and can't give ETA before internal discussion, but i'm giving a code review here anyway :)

On Tron we have a delegate freeze transaction. Where one can send energy/bandwidth to another address/contract. Without have to send them the TRX.

It happens the same way of usual freeze, where user will have to wait for minimum 3 days to unfreeze. The resource is credit in the delegated account. The user delegating still keep the TronPower (used to vote)

gre:
Thank you for your help. 
I've just synchronized with my team. We thought about implementing it in the past in fact, but so far we consider it an expert feature. A lot of UX work would be needed. I'm going to close this PR for now, but thanks a lot for the work. We will use it if and when we do decide to implement freezing operation to delegate resources to other accounts.

🤝

PR #623, opened 2020-04-23
Stellar: Fixes false negative broadcast error

gre:
## The problem

A Stellar transaction can fail at broadcast time but in fact still eventually passed. This is a high risk of double sending funds in current status of Ledger Live.

More precisely, you can have a long period of `Broadcasting...` which eventually falls into a HTTP 504 error which makes user think you have to try to do it again. In fact, if you go back to account and wait for a few minutes, it's very likely the operation eventually will appear.

## The current proposal solution

If we see a broadcast error and that the broadcasting time was long enough (20s threshold), we assume it's a success.
This should make Ledger Live correctly creating the pending operations and replacing it as soon as a network operation appear thanks to `transactionSequenceNumber`.

As it was also discussed, we MUST also prevent a successive transaction to be done so we introduce an ERROR in case you do a send again too quickly.

### The tradeoff of that solution

- on Mobile, we can't know the error status code (which is another task in our sprint to fix in future) so we can't refine the error case to specifically 504, for now we catch all errors, including potential case it would take forever to broadcast like VERY slow network. it's a compromise here. By default, Ledger Live keep an operation pending only for 30mn.
- on Desktop, when you do a Send, stay on the modal and do a `< BACK` **it seems we do not refresh the account model** you will not see the new error. it's an potential problem we'll need to track internally. Also we must not be able to `< BACK` during broadcast.
- there is an imprecision in the represensation of `transactionSequenceNumber` of about +- 10 ledger at the moment. this is not a blocker and will be improved in future.

> A great master once said, “ça plante mais en fait ça plante pas”.

### DIFF
--- a/src/families/stellar/libcore-buildOperation.js
+++ b/src/families/stellar/libcore-buildOperation.js
@@ -1,6 +1,7 @@
 // @flow
 
 import type { CoreOperation } from "../../libcore/types";
+import { libcoreBigIntToBigNumber } from "../../libcore/buildBigNumber";
 
 async function stellarBuildOperation({
   coreOperation
@@ -8,9 +9,17 @@ async function stellarBuildOperation({
   coreOperation: CoreOperation
 }) {
   const stellarLikeOperation = await coreOperation.asStellarLikeOperation();
-  const stellarLikeTransaction = await stellarLikeOperation.getRecord();
-  const hash = await stellarLikeTransaction.getTransactionHash();
-  return { hash };
+  const stellarLikeRecord = await stellarLikeOperation.getRecord();
+  const stellarLikeTransaction = await stellarLikeOperation.getTransaction();
+  const hash = await stellarLikeRecord.getTransactionHash();
+  const transactionSequenceNumberRaw = await stellarLikeTransaction.getSourceAccountSequence();
+  const transactionSequenceNumber = await libcoreBigIntToBigNumber(
+    transactionSequenceNumberRaw
+  );
+  return {
+    hash,
+    transactionSequenceNumber: transactionSequenceNumber.toNumber()
+  };
 }
 
 export default stellarBuildOperation;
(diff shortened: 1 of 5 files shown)


juan-cortes:
While this might prevent the double-spending of funds if the transaction takes a little bit of time to be broadcasted I believe the proper way of doing this is by specifying the time bounds on tx submission. I see you are setting a limit of 20s but what if it takes longer? There's still a risk of double-spending.

https://www.stellar.org/developers/guides/concepts/transactions.html#time-bounds

As long as the submitted tx is specifying the time bounds to be now + 20s it should be ok. Otherwise, it can still fail.

src/families/stellar/libcore-buildOperation.js
```diff
@@ -1,16 +1,25 @@
...
   coreOperation: CoreOperation
 }) {
   const stellarLikeOperation = await coreOperation.asStellarLikeOperation();
-  const stellarLikeTransaction = await stellarLikeOperation.getRecord();
-  const hash = await stellarLikeTransaction.getTransactionHash();
-  return { hash };
+  const stellarLikeRecord = await stellarLikeOperation.getRecord();
+  const stellarLikeTransaction = await stellarLikeOperation.getTransaction();
+  const hash = await stellarLikeRecord.getTransactionHash();
+  const transactionSequenceNumberRaw = await stellarLikeTransaction.getSourceAccountSequence();
```
henri-ly, line 15: Do you a different value for each operation ? 
I'm not sure if this one return the correct Sequence number or the Account Sequence.
If it's not we probably need to ask some dev to libcore to get the right sequence number

gre: checked with @pollastri-pierre , getSourceAccountSequence correctly returns the transaction ledger and not the account's. phew.

PR #625, opened 2020-04-27
[Stellar] use modeAccept for network fees

henri-ly:
We were used an hard-coded fees because Daemon Block fees wasn't working properly so we had to hotfix it.

Now that libcore got the right value it should be okay !


gre:
is there a test that covers is already? i guess as part of the dataset it's going to go through it right? weird that codecov did not triggered here

PR #626, opened 2020-04-27
Mock - Add a migratable account for testing

juan-cortes:
When scanning accounts for ethereum in mock mode we will now get the third account as migratable. We are using the `version` of the id as the discriminatory field. This allows us to complete the account migration flow on desktop with spectron.


gre:
it looks like someone did a ETH tx that makes the CI fail :ohno:
cc @Arnaud97234

PR #627, opened 2020-04-28
LL-2407

LFBarreto:
Add in base operation types for :atom: and extra information needed on specific operations
need validation before starting working on mock operations


LFBarreto:
@gre i think so yes its just in order to agrre on a structure to advance on mocks

PR #630, opened 2020-04-29
LL-2407

LFBarreto:
Add in mock generation of specific :atom: operations:

- [x] delegate

- [x] redelegate

- [x] undelegate

- [x] claim rewards


src/families/cosmos/mock.js
```diff
@@ -0,0 +1,170 @@
...
+  const opIndex = rng.next(0, 10);
+
+  const lastOp = ops[opIndex];
+  const date = new Date(
+    (lastOp ? lastOp.date.valueOf() : Date.now()) -
+      rng.nextInt(0, 100000000 * rng.next() * rng.next())
+  );
+
+  /** select between 3 to 5 validators and split the amount evenly */
+  const delegatedValidators = Array(rng.nextInt(3, 5))
```
LukeSugiura, line 43: This can be `Array.from({ length: rng.nextInt(3, 5) })` then remove `fill(null)`

src/families/cosmos/mock.js
```diff
@@ -0,0 +1,170 @@
...
+    recipients: [],
+    blockHash: genHex(64, rng),
+    blockHeight: account.blockHeight - Math.floor((Date.now() - date) / 900000),
+    accountId: account.id,
+    date,
+    extra
+  };
+
+  /** update account cosmosResources, spendableBalance and operations */
+  account.cosmosResources = cosmosResources;
```
LukeSugiura, line 112: I prefer immutable way
```
return {
  ...account,
  cosmosResources,
  spendableBalance,
  ...
}
```

src/families/cosmos/mock.js
```diff
@@ -0,0 +1,170 @@
...
+}
+
+/**
+ * add in specific cosmos operations
+ * @memberof cosmos/mock
+ * @param {Account} account
+ * @param {Prando} rng
+ */
+function enhanceOperations(account: Account, rng: Prando): Account {
+  account = genDelegationOperation(account, rng);
```
LukeSugiura, line 127: can you simply write like this? in this case, what's the meaning of wrapping genDelegationOperation tho
```
return genDelegationOperation(account, rng)
```

LFBarreto: we will chain several operations here i guess we can figure a cleaner way once we have each operations set :thinking:

src/families/cosmos/mock.js
LFBarreto: weidly the operations aren't updated that way :shrug:

src/bridge/mockHelpers.js
```diff
@@ -150,6 +161,12 @@ export const scanAccounts: $PropertyType<CurrencyBridge, "scanAccounts"> = ({
         account.name = validateNameEdition(account);
         if (isLast) {
           account.spendableBalance = account.balance = BigNumber(0);
+
+          const perFamilyOperation = perFamilyMock[currency.id];
+          const clearAccount =
+            perFamilyOperation && perFamilyOperation.clearAccount;
+
+          if (clearAccount) clearAccount(account);
```
gre, line 169: In term of semantic I don't really understand why it's called clearAccount.

I think maybe we could have a general "enhancer" for all accounts that get scanned. or even more generally a way for you to overide what is getting generated.

to start with it could just be a `postAccountScanned: (Account, { isEmpty }) => ?Account` which you only apply for the case of an empty?

src/bridge/mockHelpers.js
```diff
@@ -44,6 +47,14 @@ export const sync: $PropertyType<AccountBridge<*>, "sync"> = initialAccount =>
           balance,
           spendableBalance: balance
         };
+
+        const perFamilyOperation = perFamilyMock[acc.currency.id];
+        const enhanceAccount =
+          perFamilyOperation && perFamilyOperation.enhanceAccount;
+
+        if (enhanceAccount) return enhanceAccount(nextAcc);
```
gre, line 55: suggestion of naming: `postSyncAccount`

src/mock/account.js
```diff
@@ -407,5 +385,12 @@ export function genAccount(
   account.spendableBalance = account.balance = ensureNoNegative(
     account.operations
   );
+
+  const perFamilyOperation = perFamilyMock[currency.id];
+  const enhanceOperations =
+    perFamilyOperation && perFamilyOperation.enhanceOperations;
+
+  if (enhanceOperations) enhanceOperations(account, rng);
```
gre, line 393: genAccountEnhanceOperations ?

src/families/cosmos/types.js
```diff
@@ -115,15 +115,15 @@ export type CosmosDelegateTxInfo = {|
 |};
 
 export type CosmosUndelegateTxInfo = {|
-  validators: CosmosDelegationInfo[]
+  validator: CosmosDelegationInfo[]
```
gre, line 118: it's still an array here 🤔

src/mock/helpers.js
```diff
@@ -0,0 +1,39 @@
...
+  rng: Prando
+) {
+  if (
+    currency.type === "CryptoCurrency"
+      ? currency.family === "ethereum" // all eth family
+      : currency.id.startsWith("ethereum") // erc20 case
+  ) {
+    return `0x${genHex(40, rng)}`;
+  }
+  return genBitcoinAddressLike(rng);
```
gre, line 38: thoughts for future. each family should define their account address generator, it should be a function with things like `(currency, derivationMode) => string`

PR #633, opened 2020-05-04
enable native segwit on LTC and DGB

valpinkman:


src/libcore/scanAccounts.js
```diff
@@ -176,6 +179,15 @@ export const scanAccounts = ({
 
             let result;
 
+            if (derivationMode === "native_segwit") {
+              const { name, version } = await getAppAndVersion(transport);
```
gre, line 183: can you move this logic inside the if below?

valpinkman: done

PR #634, opened 2020-05-04
Countervalues Rework

gre:
first iteration of a new PoC of the countervalues logic. It is not yet ready for production as iteration is ongoing.


src/countervalues/react.js
```diff
@@ -0,0 +1,457 @@
...
+  );
+
+  // trigger poll but not every time poll callback is changed
+  const [triggerPoll, setTriggerPoll] = useState(false);
+
+  useEffect(() => {
+    if (pending || !triggerPoll) return;
+    dispatch({ type: "pending" });
+    loadCountervalues(state, userSettings).then(
+      (state) => {
```
gre, line 104: ⚠️  there are race condition here if a second effect runs before the end of the first

src/countervalues/react.js
```diff
@@ -0,0 +1,457 @@
...
+  }, [autopollInterval, pollInitDelay, isPolling]);
+
+  const polling = useMemo<Polling>(
+    () => ({
+      wipe: () => {
+        dispatch({ type: "wipe" });
+      },
+      poll: () => setTriggerPoll(true),
+      start: () => setIsPolling(true),
+      stop: () => setIsPolling(false),
```
gre, line 162: I don't understand the distinction of the `isPolling` and `pending`

src/countervalues/react.js
```diff
@@ -0,0 +1,457 @@
...
+      if (
+        JSON.stringify(userSettings) === JSON.stringify(userSettingsRef.current)
+      )
+        return;
+    }
+    setTriggerPoll(true);
+    userSettingsRef.current = userSettings;
+  }, [userSettings]);
+
+  const [isPolling, setIsPolling] = useState(true);
```
gre, line 141: I don't understand the difference between `isPolling` and `pending`

JunichiSugiura: `isPolling` state is used to trigger polling effect where `pending` is the state.
https://github.com/LedgerHQ/ledger-live-common/blob/countervalues-poc/src/countervalues/react.js#L160-L161

JunichiSugiura: `triggerPollling` would be more appropriate name

gre: but there is already a triggerPoll 🤔

JunichiSugiura: oh, i've noticed that `triggerPolling` also exists.

JunichiSugiura: `triggerPolling` for one shot fetching and `isPolling` for continuous syncing.

PR #636, opened 2020-05-05
Move device transaction logic to live-common

juan-cortes:
Brings to live-common the logic that determines whether a field is displayed or not on the device confirmation steps. This is renderer agnostic and needs a companion component with matching names on the desktop/mobile implementation for it to work.

### Context
https://ledgerhq.atlassian.net/browse/LL-2423


gre:
let's have potential tests implemented as a discussed in another future PR ;)

PR #639, opened 2020-05-07
Fixes #629 – add Account#creationDate

### LINKED ISSUE #629: Addition of "creationDate" in an account that correspond to oldest tx received on account
we need that for countervalues and the idea to see the graph "full history"

gre:


gre:
@juan-cortes this is not a "user data" because it's actually not the _time the account was added into Ledger Live_ but instead it's the _time the account started to exists in the **blockchain**_. So it's something we restore at a sync from the blockchain.

gre:
finally 😆

PR #655, opened 2020-05-19
LL-2548 Extract Cosmos utils from LLD

JunichiSugiura:


src/families/cosmos/react.js
```diff
@@ -14,3 +24,143 @@ export function useCosmosPreloadData() {
...
+): CosmosFormattedDelegation[] {
+  const { validators } = useCosmosPreloadData();
+  const delegations = account.cosmosResources?.delegations;
+  invariant(delegations, "cosmos: delegations is required");
+
+  const formattedDelegations = useMemo(
+    () => formatDelegations(delegations, validators),
+    [delegations, validators]
+  );
+  const unit = useMemo(() => getAccountUnit(account), [account]);
```
gre, line 40: this does not need `useMemo`

src/families/cosmos/react.js
```diff
@@ -14,3 +24,143 @@ export function useCosmosPreloadData() {
...
+export function useCosmosFormattedDelegations(
+  account: Account,
+  option?: FormatOption
+): CosmosFormattedDelegation[] {
+  const { validators } = useCosmosPreloadData();
+  const delegations = account.cosmosResources?.delegations;
+  invariant(delegations, "cosmos: delegations is required");
+
+  const formattedDelegations = useMemo(
+    () => formatDelegations(delegations, validators),
```
gre, line 37: shouldn't the switch be INSIDE this? the value of useMemo could only be useful if we move the switch **inside** the function too and that we also accept an `option` as useMemo dependency.

src/families/cosmos/utils.js
```diff
@@ -0,0 +1,60 @@
...
+import { BigNumber } from "bignumber.js";
+import type {
+  CosmosDelegation,
+  CosmosDelegationInfo,
+  CosmosValidatorItem,
+  CosmosFormattedDelegation,
+} from "./types";
+import type { Unit } from "../../types";
+
+export function formatDelegations(
```
gre, line 11: naming: I wouldn't call this `format` as I wouldn't call the type `CosmosFormatted...`. `format*` function naming we have in live-common is usually for function that return texts.

could we find a better wording?

possible prefix/suffix: Detail, Full, Row, 🤔

src/families/cosmos/utils.js
```diff
@@ -0,0 +1,60 @@
...
+}
+
+export const formatDelegationsInfo = (
+  delegations: CosmosDelegationInfo[],
+  validators: CosmosValidatorItem[]
+): {
+  validator: ?CosmosValidatorItem,
+  address: string,
+  amount: BigNumber,
+}[] => {
```
gre, line 33: I don't understand what this is VS the other. we have a lot of types that are a bit hard to understand at first glance.

src/families/cosmos/types.js
```diff
@@ -161,3 +161,18 @@ export type TransactionRaw = {|
...
+  address: string,
+  pendingRewards: BigNumber,
+  amount: BigNumber,
+  status: CosmosDelegationStatus,
+};
+
+export type CosmosFormattedDelegation =
+  | CosmosFormattedDelegationRaw
+  | (CosmosFormattedDelegationRaw & { formattedAmount: string })
+  | (CosmosFormattedDelegationRaw & { reward: string });
```
gre, line 176: this is a bit too complex type structure to me. what is the reason behind this? if we don't have a _field discriminator_ then it's probably not a good idea because you are opening a lot of problems with FlowType as it's not going to be able to differienciate.

**I would just add these as `?: string`**

src/families/cosmos/react.js
```diff
@@ -14,3 +24,143 @@ export function useCosmosPreloadData() {
...
+) {
+  const [query, setQuery] = useState<string>("");
+  const delegations = useCosmosFormattedDelegations(account, option);
+
+  const options = useMemo<CosmosFormattedDelegation[]>(
+    () =>
+      delegations.filter(
+        // [TODO] better query test
+        ({ validator }) =>
+          !query || !validator || new RegExp(query, "gi").test(validator.name)
```
gre, line 83: if we don't know exactly what this is, we better should expose it as a function (filtering function) rather than accepting obscuring regexp the userland can set 🤔 

i wonder if there is existing engine we have somewhere (fuzzy search, etc..). what did we do for tron?

src/families/cosmos/react.js
```diff
@@ -14,3 +24,143 @@ export function useCosmosPreloadData() {
...
+      ),
+    [query, delegations]
+  );
+
+  const selectedValidator = useMemo(
```
gre, line 88: there is no need for `useMemo`

src/families/cosmos/react.js
```diff
@@ -14,3 +24,143 @@ export function useCosmosPreloadData() {
...
+  search: string,
+  validators: CosmosValidatorItem[],
+  delegations: CosmosDelegationInfo[]
+): {
+  validator: CosmosValidatorItem,
+  name: ?string,
+  address: string,
+  rank: number,
+}[] {
+  const { current: initialVotes } = useRef(
```
gre, line 132: the use of `useRef` here is suboptimal, you are always calculating `delegations.map(({ address }) => address)` here and pass it to useRef. personally i would just have use `const [initialVotes] = useState(() => delegations.map(({ address }) => address))` that (even if it's a read only state) is clearer to read and also it benefit from fact the init-er is called once.

JunichiSugiura: @LFBarreto

LFBarreto: i'm not sure the useRef recalculates but sure we can do this with useState instead might be clearer to read :+1:

JunichiSugiura: @gre In this case, useMemo with empty dependencies is more appropriate imo.
@LFBarreto I checked whether useRef recalculates and yes, it does.

JunichiSugiura: since useMemo doesn't guarantee to prevent re-computation, I would suggest to create useMemoOnce.

```
function useMemoOnce<T>(callback: () => T): T {
  const ref = useRef()
 
  if (!ref.current) {
    ref.current = callback()
  }

  return ref.current
}
```

src/families/cosmos/react.js
LFBarreto: on tron we used something like this:
`const terms = `${name || ""} ${address}`;
  return terms.toLowerCase().includes(query.toLowerCase().trim());`

PR #659, opened 2020-05-20
Add fee to processTransaction

teams2ua:
We need to have fees in swap, to user validation.


juan-cortes:
@gre I'm integrating this directly in the swap branch, it's easier like this

juan-cortes:
Code was directly added to https://github.com/LedgerHQ/ledger-live-common/tree/swap

PR #661, opened 2020-05-22
Cosmos

henri-ly:


cli/yarn.lock
```diff
@@ -1033,6 +1033,7 @@
     ethereumjs-tx "^1.3.7"
     invariant "^2.2.2"
     isomorphic-ws "^4.0.1"
+    ledger-cosmos-js "^2.1.7"
```
gagbo, line 1036: We use the bundled one in cli too technically. That should clean up the rest of the diff in the lock file

cli/yarn.lock
```diff
@@ -3367,6 +3378,24 @@ kuler@1.0.x:
...
+  version "2.1.7"
+  resolved "https://registry.yarnpkg.com/ledger-cosmos-js/-/ledger-cosmos-js-2.1.7.tgz#362d1139ac2504ccb56029abda053b2c5b290e8d"
+  integrity sha512-RyaP+6lRhllQTgk5X14PiZtsUE8bYP7mOiFkOMAzPQvUDdy8IZr17mDaH9whqHtE0wZd95YPN89VAhbfV+Re8w==
+  dependencies:
+    "@babel/runtime" "^7.7.4"
+    "@ledgerhq/hw-transport" "^4.77.0"
+    bech32 "^1.1.3"
+    ripemd160 "^2.0.2"
+
+"ledger-cosmos-js@file:.yalc/ledger-cosmos-js":
```
gagbo, line 3391: This yalc leftover can be cleaned with yalc remove I suppose ?

package.json
```diff
@@ -30,7 +30,8 @@
     "flow",
     "react.js",
     "reactNative.js",
-    ".flowconfig"
+    ".flowconfig",
+    ".yalc"
```
gagbo, line 34: We don't need the `.yalc` trick anymore, we can ask Live on another PR

package.json
```diff
@@ -85,6 +86,7 @@
     "ethereumjs-tx": "^1.3.7",
     "invariant": "^2.2.2",
     "isomorphic-ws": "^4.0.1",
+    "ledger-cosmos-js": "^2.1.7",
```
gagbo, line 89: Ledger cosmos js has been bundled for the time being

src/env.js
```diff
@@ -70,6 +75,16 @@ const envDefinitions = {
...
   },
+  COSMOS_GAS_AMPLIFIER: {
+    def: 2,
+    parser: intParser,
+    desc: "estimate gas multiplier",
+  },
+  COSMOS_GAS_PRICE: {
+    def: 0.025,
+    parser: floatParser,
+    desc: "gasLimit * gasPrice to determine the fees price",
```
gagbo, line 86: "A too low GAS_PRICE will get rejected before the transaction is broadcast" _may_ be added

src/errors.js
```diff
@@ -66,3 +66,34 @@ export const AccountAwaitingSendPendingOperations = createCustomErrorClass(
...
+  "CosmosTooMuchRedelegations"
+);
+export const CosmosTooMuchUnboundings = createCustomErrorClass(
+  "CosmosTooMuchUnboundings"
+);
+export const CosmosRedelegationInProgress = createCustomErrorClass(
+  "CosmosRedelegationInProgress"
+);
+
+export const CosmosBroadcastError = {
```
gagbo, line 80: Add the source for error codes so that if something weird happens we know where to look : https://github.com/cosmos/cosmos-sdk/blob/v0.37.9/types/errors.go#L28

I'm afraid that these values will have to be shifted by 1 when cosmos pushes the `protobuf` changes

src/families/cosmos/app/helperV1.js
```diff
@@ -0,0 +1,76 @@
+import { CLA, errorCodeToString, INS, processErrorResponse } from "./common";
```
gagbo, line 1: We can remove the whole file and all references to helperV1 I think.

Live isn't going to support v1 of cosmos app anyway

yarn.lock
```diff
@@ -4827,6 +4860,16 @@ lcid@^2.0.0:
   dependencies:
     invert-kv "^2.0.0"
 
+ledger-cosmos-js@^2.1.7:
```
gagbo, line 4863: Same here, as we bundle ledger-cosmos-js for the time being, there shouldn't be mentions of ledger-cosmos-js in the lock file right ?

package.json
gre: yeah please let's not ship .yalc in NPM ^^

src/families/cosmos/app/index.js
```diff
@@ -0,0 +1,298 @@
...
+          // eslint-disable-next-line no-await-in-loop
+          result = await this.signSendChunk(1 + i, chunks.length, chunks[i]);
+          if (result.return_code !== 0x9000) {
+            break;
+          }
+        }
+
+        return {
+          return_code: result.return_code,
+          error_message: result.error_message,
```
gre, line 272: the fact there is this "weird" object with error_message makes me thing it's not going to be very well remapped as other coins. we should just "let it throw" like other implementations.

meaning when you refuse on device it's correctly identified as UserRefusedOnDevice and so on. (we based all our work on the fact we have "TransportStatusError".

yarn.lock
```diff
@@ -1172,11 +1172,25 @@
...
+    "@ledgerhq/errors" "^4.78.0"
+    "@ledgerhq/logs" "^4.72.0"
+    rxjs "^6.5.3"
+
 "@ledgerhq/errors@5.15.0", "@ledgerhq/errors@^5.15.0":
   version "5.15.0"
   resolved "https://registry.yarnpkg.com/@ledgerhq/errors/-/errors-5.15.0.tgz#e5d5b5ad48fc07f6308b78b065242e158bc044a2"
   integrity sha512-ZlLhR7qaChPgEbvcqOptRepWGm8VhhwOM6kC1gx3WErutbtaOjUX8lLA4ButWFU2f+xTl2rS/5c86wC7qGqGXQ==
 
+"@ledgerhq/errors@^4.78.0":
```
gre: this shouldn't appear. this means there is a dup of package. please run `yarn-deduplicate`.

src/families/cosmos/bridge/libcore.js
```diff
@@ -0,0 +1,237 @@
...
+      account: a,
+      transaction: t,
+    });
+  },
+  (a, t) =>
+    `${a.id}_${t.amount.toString()}_${t.recipient}_${
+      t.gasLimit ? t.gasLimit.toString() : ""
+    }_${t.fees ? t.fees.toString() : ""}
+    _${String(t.useAllAmount)}_${t.mode}_${
+      t.validators ? t.validators.map((v) => v.address).join("-") : ""
```
gre, line 41: doesn't `memo` should make the fee recalculate?
same question for cosmosSourceValidator

src/families/cosmos/bridge/libcore.js
```diff
@@ -0,0 +1,237 @@
...
+});
+
+const updateTransaction = (t, patch) => ({ ...t, ...patch });
+
+const redelegationStatusError = (a, t) => {
+  if (a.cosmosResources) {
+    const redelegations = a.cosmosResources.redelegations;
+
+    if (redelegations.length >= 7) {
+      return new CosmosTooMuchRedelegations();
```
gre, line 66: It shouldn't be a new error. please switch to `invariant` but expose us this `7` via an exported constant so our UI can takes this into account.

(UI should prevent it to happen before user see the error, so it's an invariant)

src/families/cosmos/bridge/libcore.js
```diff
@@ -0,0 +1,237 @@
...
+  const warnings = {};
+
+  if (t.mode === "redelegate") {
+    const redelegationError = redelegationStatusError(a, t);
+    if (redelegationError) {
+      errors.redelegation = redelegationError;
+    }
+  } else if (t.mode === "undelegate") {
+    if (a.cosmosResources && a.cosmosResources.unbondings.length >= 7) {
+      errors.unbondings = new CosmosTooMuchUnboundings();
```
gre, line 105: same to me.

src/families/cosmos/bridge/libcore.js
```diff
@@ -0,0 +1,237 @@
...
+    errors,
+    warnings,
+    estimatedFees,
+    amount,
+    totalSpent,
+  });
+};
+
+const prepareTransaction = async (a, t) => {
+  return t;
```
gre, line 186: is there anything i miss in the reason why there is no filling up of `networkInfo` in here?

src/families/cosmos/bridge/libcore.js
```diff
@@ -0,0 +1,237 @@
...
+};
+
+const currencyBridge: CurrencyBridge = {
+  preload: async () => {
+    const validators = await getValidators();
+    setCosmosPreloadData({ validators });
+    return { validators };
```
gre, line 193: please make it explicit that preload function returns a Promise of CosmosPreloadData (as in mock)

src/families/cosmos/app/index.js
```diff
@@ -0,0 +1,298 @@
...
+        const returnCode = errorCodeData[0] * 256 + errorCodeData[1];
+
+        const compressedPk = Buffer.from(response.slice(0, 33));
+        const bech32Address = Buffer.from(response.slice(33, -2)).toString();
+
+        return {
+          bech32_address: bech32Address,
+          compressed_pk: compressedPk,
+          return_code: returnCode,
+          error_message: errorCodeToString(returnCode),
```
gre, line 232: same problem with the error.

src/families/cosmos/libcore-buildTransaction.js
```diff
@@ -0,0 +1,98 @@
...
+    gas = await libcoreBigIntToBigNumber(
+      await cosmosLikeAccount.estimateGas(gasRequest)
+    );
+  }
+  const gasAmount = await bigNumberToLibcoreAmount(core, coreCurrency, gas);
+  if (isCancelled()) return;
+
+  await transactionBuilder.setGas(gasAmount);
+
+  const gasPrice = getEnv("COSMOS_GAS_PRICE");
```
gre, line 78: this should be part of the transaction object (as the other env)

src/families/cosmos/libcore-buildTransaction.js
```diff
@@ -0,0 +1,98 @@
...
+    core
+  );
+
+  promiseAllBatched(
+    3,
+    messages,
+    async (message) => await transactionBuilder.addMessage(message)
+  );
+
+  const memoTransaction = memo ? memo : "";
```
gre, line 55: `memo || ""`

src/families/cosmos/libcore-postBuildAccount.js
```diff
@@ -0,0 +1,143 @@
...
+  const validatorInfo = await cosmosAccount.getValidatorInfo(address);
+  return status[await validatorInfo.getActiveStatus()];
+};
+
+const getFlattenDelegation = async (cosmosAccount) => {
+  const delegations = await cosmosAccount.getDelegations();
+
+  return await promiseAllBatched(10, delegations, async (delegation) => {
+    const validatorAddress = await delegation.getValidatorAddress();
+    const pendingRewards = await cosmosAccount.getPendingRewards();
```
gre, line 27: is there any reason why the pendingRewards fetching and the loop was not BEFORE the promiseAll? we are doing this again and again for each delegation, could it be a performance issue?

src/families/cosmos/libcore-postBuildAccount.js
```diff
@@ -0,0 +1,143 @@
...
+        : BigNumber(0),
+      status: await getValidatorStatus(cosmosAccount, validatorAddress),
+    };
+  });
+};
+
+const getFlattenRedelegations = async (cosmosAccount) => {
+  const redelegations = await cosmosAccount.getRedelegations();
+
+  return redelegations.reduce(async (old, current) => {
```
gre, line 55: is there a reason to not use promiseAllBatched here?

src/families/cosmos/libcore-postBuildAccount.js
```diff
@@ -0,0 +1,143 @@
...
+      }
+    );
+    return [...collection, ...redelegationMapped];
+  }, Promise.resolve([]));
+};
+
+const getFlattenUnbonding = async (cosmosAccount) => {
+  const unbondings = await cosmosAccount.getUnbondings();
+
+  return unbondings.reduce(async (old, current) => {
```
gre, line 79: same

src/families/cosmos/message.js
```diff
@@ -0,0 +1,118 @@
...
+          return await core.CosmosLikeMessage.wrapMsgDelegate({
+            delegatorAddress: freshAddress,
+            validatorAddress: validator.address,
+            amount: { amount: validator.amount.toString(), denom: "uatom" },
+          });
+        })),
+      ];
+    }
+  }
+  throw new Error("unknown message");
```
gre, line 117: includes the .mode maybe?

src/families/cosmos/validators.js
```diff
@@ -0,0 +1,197 @@
...
+      (1 - minPoint / 2) * rewardsState.inflationMinRate +
+      (minPoint / 2) * rewardsState.currentValueInflation;
+    return averageInflation;
+  }
+
+  throw new Error("Unreachable code");
+};
+
+export const getRewardsState = async () => {
+  return await cacheRewardsState();
```
gre, line 168: maybe just rename `cacheRewardsState` to `getRewardsState`

src/env.js
```diff
@@ -32,6 +32,11 @@ const stringParser = (v: mixed): ?string =>
   typeof v === "string" ? v : undefined;
 
 const envDefinitions = {
+  API_COSMOS_BLOCKCHAIN_EXPLORER_API_ENDPOINT: {
+    def: "https://cosmos.coin-proxy.dev.aws.ledger.fr",
```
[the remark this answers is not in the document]
henri-ly: Not yet, we still don't have a dev one, officially... This one was from Arnaud that have leave Ledger two weeks ago :/

src/errors.js
gagbo: Those can be seen in a traditional user path
```
CodeOK                CodeType = 0
	CodeUnknownAddress    CodeType = 9
	CodeOutOfGas          CodeType = 12
	CodeMemoTooLarge      CodeType = 13
```

These are (probably) happening if we have a bug on Live side
```
	CodeTxDecode          CodeType = 2
	CodeInvalidSequence   CodeType = 3
	CodeUnauthorized      CodeType = 4
	CodeInsufficientFunds CodeType = 5
	CodeInvalidAddress    CodeType = 7
	CodeInsufficientCoins CodeType = 10
	CodeInsufficientFee   CodeType = 14
	CodeTooManySignatures CodeType = 15
	CodeGasOverflow       CodeType = 16
	CodeNoSignatures      CodeType = 17
```

These probably happen if we have an issue on Core side
```
	CodeUnknownRequest    CodeType = 6
	CodeInvalidPubKey     CodeType = 8
	CodeInvalidCoins      CodeType = 11
```

This one means we have a serious issue (in that, we don't know what happens)
```
CodeInternal          CodeType = 1
```

Basically we have this information coming directly from the underlying node we use for broadcasting, and if users have issues sending transaction, having the exact value in the "Screenshot" that users send is going to be a _huge_ help to debug the issue. Maybe we can "skip" traduction by saying it's an "internal error" with the english name for the code ?

src/__tests__/__snapshots__/all.libcore.js.snap
```diff
@@ -13766,6 +13766,1868 @@ Array [
...
+          "pendingRewards": "18",
+          "status": "bonded",
+          "validatorAddress": "cosmosvaloper1clpqr4nrk4khgkxj78fcwwh6dl3uw4epsluffn",
+        },
+      ],
+      "pendingRewardsBalance": "12151",
+      "redelegations": Array [
+        Object {
+          "amount": "50000",
+          "completionDate": "Thu Jun 04 2020 04:40:52 GMT-0400 (Eastern Daylight Time)",
```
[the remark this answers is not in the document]
gagbo: The completion time is stored in the state of the chain, it doesn't depend on the node

src/env.js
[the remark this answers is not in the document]
gagbo: Not yet, still working on this

src/env.js
```diff
@@ -70,6 +75,16 @@ const envDefinitions = {
     desc:
       "force implementation for ALL currency bridges (affects scanning accounts)",
   },
+  COSMOS_GAS_AMPLIFIER: {
+    def: 2,
+    parser: intParser,
+    desc: "estimate gas multiplier",
+  },
+  COSMOS_GAS_PRICE: {
```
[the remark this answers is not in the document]
gagbo: The user might want to customize this because this is a spam protection mechanism only : if the user wants to use their own API node with a lower COSMOS_GAS_PRICE, we should allow them to modify it here so that the automatic fee estimation can directly use this value to make the computation

src/families/cosmos/bridge/libcore.js
gagbo: technically it should, but hopefully this is covered by the multiplier on the initial gas estimate. A "send" typically uses 30/40k Gas, and memo costs 10 gas per byte

the cosmosSourceValidator might have effects I don't know of, but theoretically this shouldn't change the cost much

src/families/cosmos/bridge/libcore.js
gagbo: what should networkInfo contain ?

src/families/cosmos/libcore-buildTransaction.js
gagbo: This is only related to the actual instance of the Gaia backend we're using, not to the transaction theoretically

src/families/cosmos/libcore-postBuildAccount.js
henri-ly: Totally perf issue. I could've done in a smarter way 👍

src/env.js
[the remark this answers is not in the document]
gre: ok for now, but for long term we need to consider the concept of per family/account/currency settings that we expose to UI.

src/families/cosmos/bridge/libcore.js
gre: I don't know but i see it in the current Transaction, so if we don't need them, we should remove.
usually it's things like the current network fees. it's often in the cycle of a Transaction creation so it doesn't "resolve" on the fly but only once at the "initial load" of one send flow.

src/errors.js
gre: ok. i guess we can keep these error here, and this comment will be good to guide to what worth a translation.

PR #663, opened 2020-05-25
LL-2549

LFBarreto:
:atom: mocks updated with unbondings and redelegations + fees operations
also added utils for mapping unbondings and redelegations to given validators data.

### DIFF
--- a/src/families/cosmos/mock.js
+++ b/src/families/cosmos/mock.js
@@ -2,7 +2,12 @@
 import Prando from "prando";
 import { BigNumber } from "bignumber.js";
 import type { Account, Operation, OperationType } from "../../types";
-import type { CosmosResources, CosmosDelegation } from "./types";
+import type {
+  CosmosResources,
+  CosmosDelegation,
+  CosmosUnbonding,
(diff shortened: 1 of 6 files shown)


src/families/cosmos/mock.js
```diff
@@ -30,8 +37,16 @@ function setCosmosResources(
       ? account.cosmosResources.unbondingBalance.plus(unbondingBalance)
       : unbondingBalance,
     withdrawAddress: account.id,
-    unbondings: [],
-    redelegations: [],
+    unbondings: unbondings
```
JunichiSugiura, line 40: with optional chaining + nullish coalescing operator
```
unboundings ?? account.cosmosResources?.unbounding ?? []
```

src/families/cosmos/mock.js
```diff
@@ -13,7 +18,9 @@ const { validators } = preloadedData;
 function setCosmosResources(
```
JunichiSugiura, line 18: it's better using one object arg when there're more than two arguments especially, some of them are optional

PR #672, opened 2020-05-27
add a way to filter on one or more coin families for the bot

valpinkman:


gre:
i think we should do it for a given coin (by looking at the spec.currency and leveraging on the "findCurrencyByTicker") but it's good enough for now

PR #677, opened 2020-05-28
tron freeze Unfreeze

Arnaud97234:
### Tron test cases

- Send certain amount
- Send Max
- Freeze Energy OR Bandwidth
- Unfreeze Energy OR Bandwidth
- Submit vote

### DIFF
--- a/src/families/tron/specs.js
+++ b/src/families/tron/specs.js
@@ -1,5 +1,6 @@
 // @flow
 import { BigNumber } from "bignumber.js";
+import sample from "lodash/sample";
 import invariant from "invariant";
 import expect from "expect";
 import sortBy from "lodash/sortBy";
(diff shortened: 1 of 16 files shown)


src/families/tron/specs.js
```diff
@@ -0,0 +1,152 @@
...
+  appQuery: {
+    model: "nanoS",
+    appName: "Tron",
+    appVersion: "0.2.x",
+  },
+  mutations: [
+    {
+      name: "move 50% to another account",
+      transaction: ({ account, siblings, bridge }) => {
+        invariant(account.balance.gt(minimalAmount), "balance is too low");
```
gre, line 27: most usage of `account.balance` shall be replaced by `maxSpendable` because not all your balance is spendable.

src/families/tron/specs.js
```diff
@@ -0,0 +1,152 @@
...
+      test: ({
+        account,
+        accountBeforeTransaction,
+        transaction,
+        optimisticOperation,
+        operation,
+      }) => {
+        const resourceType = transaction.resource;
+        const expectedAmount = BigNumber(transaction.amount)
+          .times(10e6)
```
gre, line 94: i don't understand why we multiply by 10^6

src/families/tron/specs.js
```diff
@@ -0,0 +1,152 @@
...
+      },
+    },
+    {
+      name: "unfreeze bandwith / energy",
+      transaction: ({ account, bridge, amount }) => {
+        invariant(account.tronResources.tronPower.gt(0), "no frozen assets");
+        const currentDate = new Date();
+        invariant(
+          account.tronResources.frozen.bandwidth.expiredAt < currentDate &&
+            currentDate > account.tronResources.frozen.energy.expiredAt,
```
gre, line 117: to be checked what we have in our live-common but i suspect we have helpers function for these that could be at use.

src/families/tron/specs.js
Arnaud97234: Because the units are not the same between `amount` entered for the transaction and the value of `tronResources.frozen...` value that we use in the test.

gre: 🤯  this is pretty weird we have this in live-common. transaction.amount is supposed to be in TRON, otherwise we must introduce another field / put in extra.
cc @nukky94  @LFBarreto

gre:
i merged master in, but it doesn't seem the spec are working right now.

PR #678, opened 2020-05-29
Inflate extraInfo correctly in cosmos transactions

gagbo:


henri-ly:
I had to add Core to buildAccount, can it bring any issue like performance ?

gre:
Does it need a libcore bump? CI shows this is not mergeable. JS error

henri-ly:
No I don't think we need to bump libcore, i had these function with the reflect I'll investigate on why it doesn't work

PR #685, opened 2020-06-01
[Cosmos] Change spendable balance

henri-ly:


gre:
in the test we can see the spendableBalance already have moved. if this value is always moving, maybe we should ignore it. what is your theory on this @henri-ly  ?

PR #698, opened 2020-06-04
bot: Add bitcoin_testnet spec + handle inconsistency w/ device units

onyb:
### Note on handling device units

Since the device can only display ASCII characters, the screen shows `TEST` instead of `𝚝BTC`.

~This PR is about handling the above case, by introducing a new field to `CryptoCurrency` type called `deviceUnits`. If in future, the device starts displaying `TEST BTC` instead of just `TEST`, a simple change to `makeDeviceTestnetUnit` should be enough.~

This PR is about handling the above case, by introducing a new field to `CryptoCurrency` type called `deviceTicker`.

### DIFF
--- a/src/families/bitcoin/speculos-deviceActions.js
+++ b/src/families/bitcoin/speculos-deviceActions.js
@@ -42,10 +42,17 @@ const acceptTransaction: DeviceAction<Transaction, {}> = ({
   }
 
   if (s.screen === "fees") {
-    let expectedText = formatCurrencyUnit(account.unit, status.estimatedFees, {
-      showCode: true,
-      disableRounding: true,
-    }).replace(/\s/g, " ");
+    let expectedText = formatCurrencyUnit(
+      {
+        ...account.unit,
+        code: account.currency.deviceTicker || account.unit.code,
+      },
+      status.estimatedFees,
+      {
+        showCode: true,
+        disableRounding: true,
+      }
+    ).replace(/\s/g, " ");
     let text = event.text;
 
     if (account.currency.id === "pivx" && text.startsWith("PIV ")) {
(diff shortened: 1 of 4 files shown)


src/currencies/getDeviceUnit.js
```diff
@@ -0,0 +1,20 @@
...
+import type { Account, Unit } from "../types";
+
+/**
+ * getDeviceUnit returns the unit that appears on the device, based on the
+ * user preferred unit for the account.
+ *
+ * Normally, it is the same as account.unit, except for testnet accounts,
+ * where the unit on device is just "TEST".
+ */
+export function getDeviceUnit(account: Account): Unit {
```
gre, line 13: I suggest a small simplification, we need need to have `deviceUnits?: Unit[]` but just a `deviceTicker?: string`

I think especially we can:
- in CryptoCurrency type a `deviceTicker?: string` 
- define for bitcoin_testnet `deviceTicker: "TEST"`
- in the deviceAction, you can `account.currency.deviceTicker || account.unit`

src/families/bitcoin/speculos-deviceActions.js
```diff
@@ -42,10 +42,17 @@ const acceptTransaction: DeviceAction<Transaction, {}> = ({
...
 
   if (s.screen === "fees") {
-    let expectedText = formatCurrencyUnit(account.unit, status.estimatedFees, {
-      showCode: true,
-      disableRounding: true,
-    }).replace(/\s/g, " ");
+    let expectedText = formatCurrencyUnit(
+      {
+        ...account.unit,
+        code: account.currency.deviceTicker || account.unit.code,
```
gre, line 48: nice catch. i wonder if the formatter hsouldn't have a flag to say you want the deviceTicker, but this will definitely do it for now! in future bunch of this code will move somewhere else because we're not yet done on this "deviceActions" here, as you may see the code of this file is pretty spaghetti but we'll figure out a higher level utility for this & more reusable for other apps ;)

gre:
👏

PR #711, opened 2020-06-09
Add mustUpgrade mecanism for apps + prepare shouldRecommendUpgrade

gre:
Capture d’écran 2020-06-09 à 15 59 16

### DIFF
--- a/src/apps/versionRequirements.js
+++ b/src/apps/versionRequirements.js
@@ -1,0 +1,37 @@
+// @flow
+
+import semver from "semver";
+import type { DeviceModelId } from "@ledgerhq/devices";
+import { getDependencies } from "./polyfill";
+import { getEnv } from "../env";
+
+export function shouldRecommendUpgrade(
+  deviceModel: DeviceModelId,
+  appName: string,
+  appVersion: string
+) {
+  if (getEnv("DISABLE_APP_VERSION_REQUIREMENTS")) return false;
+  const deps = getDependencies(appName);
+  if (deps.includes("Bitcoin") && deviceModel !== "blue") {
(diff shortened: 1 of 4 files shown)


src/apps/versionRequirements.js
```diff
@@ -0,0 +1,37 @@
...
+  deviceModel: DeviceModelId,
+  appName: string,
+  appVersion: string
+) {
+  if (getEnv("DISABLE_APP_VERSION_REQUIREMENTS")) return false;
+  const deps = getDependencies(appName);
+  if (deps.includes("Bitcoin") && deviceModel !== "blue") {
+    // https://donjon.ledger.com/lsb/010/
+    return !semver.satisfies(appVersion, ">= 1.4.0");
+  }
```
gagbo, line 18: Would it be possible to add a shouldUpgrade for cosmos to be `>=2.13` ? (regarding the FW app bug we saw recently)

src/hw/connectApp.js
```diff
@@ -7,14 +7,17 @@ import {
   FirmwareOrAppUpdateRequired,
   UserRefusedOnDevice,
   BtcUnmatchedApp,
+  UpdateYourApp,
```
gagbo, line 10: The error was always in the "errors" package or does the PR need a package.json and/or package.lock bump ?

src/apps/versionRequirements.js
gre: We can yeah. I think let's wait we see how we use this as contract might change in future. Today it might be highly related to recommending upgrade because security requirement.

src/hw/connectApp.js
gre: It's already there yeah

PR #712, opened 2020-06-09
[Cosmos] Claim compound warning and add error for rejected

henri-ly:


src/errors.js
```diff
@@ -113,3 +113,7 @@ export const CosmosBroadcastError = {
   "16": createCustomErrorClass(" CosmosBroadcastCodeGasOverflow"),
   "17": createCustomErrorClass(" CosmosBroadcastCodeNoSignatures"),
 };
+
+export const TransactionWasRejected = createCustomErrorClass(
+  "TransactionWasRejected"
```
gre, line 118: what do you mean by Reject? If it's user rejecting on the device, please reuse UserRefusedOnDevice

henri-ly: My bad I didn't know that errors exists. Fixed !

PR #713, opened 2020-06-09
Add support for v1 of cosmos FW app

gagbo:
The only code that is not common and that is used in this layer is the
serializePath function


gre:
ok, I think let's discuss internally because as we just have made an error that requires app >=2, i think it doesn't worth it to be backward compatible especially it makes things slower (asking app version each time).
let's focus on other important parts for now, it's not a bad idea that we require users to be up to date.

gagbo:
No worries, I had already started to prepare this just in case, so I'm just sharing the would-be changes

gre:
closing in favor of https://github.com/LedgerHQ/ledger-live-common/pull/711

PR #715, opened 2020-06-10
Cosmos unbondings limit and sort by date

LFBarreto:
+ (csms): unbondings limit raised to 7 as for redelegations 
+ sort unbondings by completion date
+ add redelegation getter helper

spreadSheet point #50 #51

### DIFF
--- a/src/families/cosmos/logic.js
+++ b/src/families/cosmos/logic.js
@@ -20,8 +20,8 @@ import type { Transaction } from "./types";
 import { getFeesForTransaction } from "../../libcore/getFeesForTransaction";
 import { makeLRUCache } from "../../cache";
 
-export const COSMOS_MAX_REDELEGATIONS = 6;
-export const COSMOS_MAX_UNBONDINGS = 6;
+export const COSMOS_MAX_REDELEGATIONS = 7;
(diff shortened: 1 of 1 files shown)


src/families/cosmos/logic.js
```diff
@@ -196,7 +198,7 @@ export function canUndelegate(account: Account): boolean {
   invariant(cosmosResources, "cosmosResources should exist");
   return (
     cosmosResources.unbondings &&
-    cosmosResources.unbondings.length < COSMOS_MAX_UNBONDINGS
+    cosmosResources.unbondings.length <= COSMOS_MAX_UNBONDINGS
```
gre, line 201: you are both increasing 6 to 7 and also moving to `<=` which means something was off by 2 ?

LFBarreto: We can have 7 active unbondings and 7 redelegations it indeed was off by 2 i guess

henri-ly: Yes it was my bad i tought it was 5, but I missmatched it with the number of validators

LFBarreto: i changed also the <= to < after testing i saw my error there sorry :sweat_smile:

PR #721, opened 2020-06-11
[Cosmos] Memo

henri-ly:


src/families/cosmos/bridge/libcore.js
```diff
@@ -300,6 +300,16 @@ const getTransactionStatus = async (a, t) => {
 };
 
 const prepareTransaction = async (a, t) => {
+  let memo = t.memo;
+
+  if (t.mode !== "send" && !memo) {
+    memo = "Ledger Live";
```
gre, line 306: with this solution I can't opt out from "Ledger Live" by setting the memo value; I wonder if we want memo to be a `?string` to support that, but as it's not exposed at all on the UI, let's go with this for now!

henri-ly: Yeah I dislike the idea to not allow the user to change what he want to write in the memo for this, but I follow the company directives

PR #722, opened 2020-06-11
Merge Fees in other Operations

gagbo:
The Fees aren't included in the operations list given by Ledger Live
Common. A fold is accomplished over the list of transactions given by
LibCore that includes the fees in all the other operations that match
the transaction Hash, and therefore the fees operations can be ignored
in the end result.

### DIFF
--- a/src/families/cosmos/libcore-postBuildAccount.js
+++ b/src/families/cosmos/libcore-postBuildAccount.js
@@ -1,5 +1,5 @@
 // @flow
-import type { Account } from "../../types";
+import type { Account, Operation } from "../../types";
 import type { CoreAccount } from "../../libcore/types";
 import type { CosmosResources, CoreCosmosLikeAccount } from "./types";
 import { BigNumber } from "bignumber.js";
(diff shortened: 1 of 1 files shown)


src/families/cosmos/libcore-postBuildAccount.js
```diff
@@ -11,6 +11,39 @@ import {
...
+  }>,
+  newCoreOp: Operation
+): Promise<{
+  ops: { [id: string]: [Operation] },
+  fees: { [tx_hash: string]: BigNumber },
+}> => {
+  let result = await accountOpsFeeStore;
+  const id = newCoreOp.id;
+  // This split comes from the code in src/families/cosmos/libcore-buildOperation.js
+  const txHash = id.split("-")[1];
```
henri-ly, line 27: I think you can get the hash from the operation no ?

henri-ly: `newCoreOp.hash` should work I think

gagbo: Didn't think of that <_< but yeah I really just want the txHash

gre:
essentially it needs to be fixed around these : https://github.com/LedgerHQ/ledger-live-common/blob/master/src/libcore/buildAccount/index.js#L152-L163

gre:
I'll try to tackle this tomorrow.

gagbo:
This is also the "base work" on which to iterate if we want to set the fees in the operations that are _not_ fees

PR #724, opened 2020-06-11
Cosmos: send all amount with delegations warning

LFBarreto:
(csms): add warning in case of useAllAmount with a delegating account

### DIFF
--- a/src/families/cosmos/bridge/libcore.js
+++ b/src/families/cosmos/bridge/libcore.js
@@ -14,6 +14,7 @@ import {
   InvalidAddressBecauseDestinationIsAlsoSource,
   InvalidAddress,
   AmountRequired,
+  RecommendUndelegation,
 } from "@ledgerhq/errors";
 import {
   CosmosRedelegationInProgress,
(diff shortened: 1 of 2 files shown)


src/families/cosmos/bridge/libcore.js
```diff
@@ -147,6 +148,14 @@ const getSendTransactionStatus = async (a, t) => {
     errors.amount = new NotEnoughBalance();
   }
 
+  if (
+    a.cosmosResources &&
+    a.cosmosResources.delegations.length > 0 &&
+    t.useAllAmount
+  ) {
+    warnings.emptyBalance = new CosmosEmptyBalanceWithDelegationWarning();
```
gre, line 156: I agree with this to be fixed, but i'm not sure about the proposal implementation

- emptyBalance is **NOT** an existing fields, warnings and errors are supposed to be attached to field. 

- in Tezos, we do this:

```
        warnings.amount = new RecommendUndelegation();
```

can you please assess if "RecommendUndelegation" wouldn't be convenient to reuse? (prefer reuse of existing errors)
can we please make sure

PR #731, opened 2020-06-12
[Cosmos] precalc fees in prepare transaction

henri-ly:


src/families/cosmos/deviceTransactionConfig.js
```diff
@@ -23,85 +86,123 @@ function getDeviceTransactionConfig({
...
-    });
-  }
-
-  if (estimatedFees && !estimatedFees.isZero()) {
-    fields.push({
-      type: "fees",
-      label: "Fees",
-    });
-  }
+  fields = [...fields, ...postCommon(account, memo, estimatedFees, gas)];
```
gre, line 205: i'm not sure why we need to split into two functions

src/families/cosmos/libcore-buildTransaction.js
```diff
@@ -65,28 +65,32 @@ export async function cosmosBuildTransaction({
...
+    estimatedGas = BigNumber(gas);
   } else {
     const gasRequest = await core.CosmosGasLimitRequest.init(
       memoTransaction,
       messages,
       getEnv("COSMOS_GAS_AMPLIFIER")
     );
-    gas = await libcoreBigIntToBigNumber(
+    estimatedGas = await libcoreBigIntToBigNumber(
       await cosmosLikeAccount.estimateGas(gasRequest)
```
gre, line 79: why are we estimating again the gas? i think it can still mismatch if we do this?

src/families/cosmos/bridge/libcore.js
```diff
@@ -191,7 +196,7 @@ const getDelegateTransactionStatus = async (a, t) => {
     errors.amount = new AmountRequired();
   }
 
-  const estimatedFees = await getEstimatedFees(a, t, errors);
+  const { estimatedFees } = await getEstimatedFees(a, t, errors);
```
gre, line 199: mmh we shouldn't estimate again if it was already done in prepareTransaction

src/families/cosmos/libcore-buildTransaction.js
henri-ly: Well we estimated here on the first time we call the buildTransaction, then on the second time the gas should be set in the Transaction object if it's go by prepareTransaction.

gre: but then, wouldn't this produce the actual bug we are currently having where the device does not match the Live display?

gre: OH i see it now!

PR #756, opened 2020-06-22
LL-2545 NatSeg/Seg sort creatable accounts

LFBarreto:
Bitcoin creatable accounts sorted with native segwit first.


gre:
isn't this going to cause click jacking during the add accounts? i wonder how it looks like, could be good to do a small video

LFBarreto:
@gre on desktop it doesn't however on mobile the whole add accounts list takes some extra time to show up so the ordering changes afterwards there :shrug:

gre:
can we try to just reverse the ifs in the derivation.js that settle the native_segwit and segwit options?

PR #759, opened 2020-06-24
Poc/modularized core

phaazon:


src/families/bitcoin/libcore-broadcast.js
```diff
@@ -2,12 +2,17 @@
...
 import { makeBroadcast } from "../../libcore/broadcast";
+import invariant from "invariant";
 
 async function broadcast({
+  core,
   coreAccount,
   signedOperation: { signature, operation }
 }) {
-  const bitcoinLikeAccount = await coreAccount.asBitcoinLikeAccount();
+  const bitcoinLikeAccount = await new core.Bitcoin().fromCoreAccount(
```
gre, line 12: i don't think this will work on mobile. there is no instantiation possible over react native. (unless you generated some JS somewhere?)

```
core.Bitcoin.fromCoreAccount
```

doesn't work?

phaazon: IIRC, nah, it doesn’t work, because `Bitcoin` is never created / inserted as an empty object in `core`. But maybe there’s something simple to create the object?

gre: i don't understand why you can do `core.CoreEthereumLikeAccount.fromCoreAccount` in that same PR

phaazon: I think it’s a mistake. I’m patching all this up.

PR #763, opened 2020-06-29
LL-2645 (Manager): add firmware unsupported helper

LFBarreto:
Add function to check if firmware version is unsupported
 
currently applies to nanoS only on versions < 1.3.0

### DIFF
--- a/src/manager/index.js
+++ b/src/manager/index.js
@@ -66,6 +66,12 @@ const CacheAPI = {
   ): boolean =>
     deviceModel === "nanoS" && semver.lte(deviceInfo.version, "1.4.2"),
 
+  firmwareUnsupported: (
+    deviceModel: DeviceModelId,
+    deviceInfo: DeviceInfo
+  ): boolean =>
+    deviceModel === "nanoS" && semver.lt(deviceInfo.version, "1.3.0"),
+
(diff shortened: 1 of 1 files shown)


src/apps/logic.js
```diff
@@ -424,6 +425,19 @@ export const isOutOfMemoryState = (state: State): boolean => {
...
 };
 
+/** checks if the device firmware is deprecated or not */
+export const isFirmwareDeprecated = (
+  deviceModel: DeviceModel,
+  deviceInfo: DeviceInfo
+): boolean => {
+  switch (deviceModel.id) {
+    case "nanoS":
+      return deviceInfo.version < "1.3.0";
```
gre, line 435: dangerous (`"1.11.0" < "1.3.0"`) . use semver.

src/apps/logic.js
```diff
@@ -424,6 +425,19 @@ export const isOutOfMemoryState = (state: State): boolean => {
   return totalAppsBlocks > appsSpaceBlocks;
 };
 
+/** checks if the device firmware is deprecated or not */
+export const isFirmwareDeprecated = (
```
gre, line 429: deprecated => unsupported

PR #768, opened 2020-07-01
first draft for the ethereum transaction data serializer

IAmMorrow:


src/families/ethereum/transactionData.js
```diff
@@ -0,0 +1,145 @@
...
+}
+
+const erc20Transfer: EthereumOperationModeDataSerializer = (a, s, t) => {
+  let amount;
+
+  // determining the token amount to transfer
+  if (t.useAllAmount) {
+    amount = s.balance;
+  } else {
+    if (t.amount.gt(s.balance)) {
```
gre, line 71: it's quite curious this part was done here. i'll investigate. serializer shouldn't have to validate^^

gre: i think it's a relicat of old code i moved into this generic function, we'll have to split it out back into some other place. anyway, not blocker for now

gre:
I think we can try to merge this soon but:

- please rebase.
- mode can be set as "send" by default. which would make it a non optional 🤔 would be consistent with other coin impls like tron,tezos,cosmos.

PR #774, opened 2020-07-06
LL-2762 – re-unifies cosmos op messages into one operation

gre:
**What does it fixes?**

- libcore implementation of cosmos has a "flattened" version of operations, which i will call here "cosmos operation messages".
- introduce `libcore-mergeOperations` hook to allow to reconciliate this.
- fixes optimistic operation to now matches the final operation (modulo the fee estimation). proven by bot.
- fixes `operation.extra` to contains BigNumber in OperationRaw by introducing a fromOperationExtraRaw/toOperationExtraRaw that can be implemented by family.

**Implementation notes**

- This closes #722
- basically it implements re-unification of libcore's operation messages into one when they match by hash.
- It assumes the operation messages of a same hash are consecutive. 
- The way it is implemented should remain performant and paginated: it stops as soon as a "same" operation is found (which was not the case in #722)

**Tests**

From first try with the desktop app, everything seems to work.

Capture d’écran 2020-07-06 à 21 10 42
Capture d’écran 2020-07-06 à 21 10 37

(nb screenshot above is outdated as the delegation amount was incorrect, now properly reflect the fees consumed)


src/families/cosmos/libcore-signOperation.js
```diff
@@ -61,6 +61,14 @@ async function signTransaction({
       ? "REWARD"
       : "OUT";
 
+  const extra = {};
+  if (transaction.mode === "REDELETE") {
```
gagbo, line 65: s/redelete/redelegate ?

gre: good catch! i'll try to add test in bot too.

PR #775, opened 2020-07-07
Ctrl+C to kill process if done after 3s again

gre:
some transports fails to properly terminate which makes the CLI hanging forever even if you do Ctrl+C.

this allows to send a second Ctrl+C 3 seconds after to close the process regardless of if it was gracefully exited.

Capture d’écran 2020-07-07 à 09 39 35


valpinkman:
Lgtm

PR #778, opened 2020-07-08
LL-2645 (Manager) Unsupported firmware

LFBarreto:
issue on 1.2 nanoS 1.2 with non semantically correct version while using semver.
we now corce and clean the version before comparing it. "1.2" -> "1.2.0"

### DIFF
--- a/src/manager/index.js
+++ b/src/manager/index.js
@@ -46,14 +46,18 @@ const CacheAPI = {
   firmwareUpdateNeedsLegacyBlueResetInstructions: (
     deviceInfo: DeviceInfo,
     deviceModelId: DeviceModelId
-  ) => deviceModelId === "blue" && semver.lt(deviceInfo.version, "2.1.1"),
+  ) =>
+    deviceModelId === "blue" &&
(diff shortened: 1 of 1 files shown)


gre:
deviceInfo.version is 1.2 ?! 😳

I wonder if we need to solve it actually on the getDeviceInfo mmh

gre:
makes me think we need this to be covered by test. (we have some data in here https://github.com/LedgerHQ/ledger-live-common/blob/master/src/__tests__/hw/getDeviceInfo.js#L8-L28 )

gre:
forget about my idea to coerce on getDeviceInfo side, it would not be great because of this:

```
> semver.coerce("1.2.7-1")
SemVer {
  options: {},
  loose: false,
  includePrerelease: false,
  raw: '1.2.7',
  major: 1,
  minor: 2,
  patch: 7,
  prerelease: [],
  build: [],
  version: '1.2.7'
}
```

also Nano X firmware are currently versioned like this, we need to keep this in mind in future: coerce works but we have to be careful about it.

PR #780, opened 2020-07-10
experimental polling of actions/app.js

gre:
- Introduces two different implementation for actions/app : one that is event based, the second that is polling based. at the moment we do the switch GLOBALLY, but in practice event based is for node-hid et polling based is for when we don't have usb events (like on bluetooth or mobile hid)
- Makes the actions/app no longer active when it have landed once into the "opened" state. This might fixes situation when we rarely saw a "double transaction" problem, but also will help ongoing development.

---

to be used with the device actions mobile poc.

```
import { setDeviceMode } from "@ledgerhq/live-common/lib/hw/actions/app";
setDeviceMode("polling");
```

---

**NB: we will need to do the same for actions/manager.js** and there is no easy factorization.

### DIFF
--- a/src/hw/actions/app.js
+++ b/src/hw/actions/app.js
@@ -8,6 +8,8 @@ import {
   catchError,
   switchMap,
   tap,
+  first,
+  delay,
 } from "rxjs/operators";
 import { useEffect, useCallback, useState, useMemo } from "react";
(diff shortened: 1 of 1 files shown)


gre:
HODL: manager was not addressed..

PR #786, opened 2020-07-20
Token support and further swap development

juan-cortes:
~This pr breaks jest tests because of the outdated snapshots~ Snapshots updated


juan-cortes:
This is where the snapshots are failing @gre, I'll need a new bumped live-common to address the issues from the sheet on LLD when you have time

gre:
@juan-cortes CI complains about lint issues

cli/src/commands/swap.js
```diff
@@ -21,20 +27,58 @@ type SwapJobOpts = ScanCommonOpts & {
...
+    });
+    // We have a token account, keep track of both now;
+    fromParentAccount = fromAccount;
+    fromAccount = subAccount;
+    invariant(fromAccount, "No account found");
+  }
+  if (fromParentAccount) {
+    console.log("fromParentAccount:", fromParentAccount.id);
+  }
+  console.log("fromAccount:", fromAccount.id);
```
gre, line 80: Can you remove the console logs?

src/families/tron/bridge/js.js
```diff
@@ -375,6 +375,7 @@ const getAccountShape = async (info, syncConfig) => {
           operations.length > 0
             ? operations[operations.length - 1].date
             : new Date(),
+        swapHistory: [],
       };
```
gre, line 379: this should be preserving the potential `maybeExistingSubAccount` the same way we preserve pending operations

src/cross.js
```diff
@@ -306,6 +332,7 @@ export const accountDataToAccount = ({
     index,
     freshAddress,
     freshAddressPath,
+    swapHistory: swapHistory.map(fromSwapOperationRaw),
```
gre, line 335: Can you run a small study on how much time it do to import accounts if you have let's say 50 swap history? (You can fake data just to look at how much qr code it generates)

src/cross.js
```diff
@@ -132,6 +139,20 @@ const asResultAccount = (unsafe: mixed): AccountData => {
...
     throw new Error("invalid account.balance");
   }
+
+  if (swapHistory && Array.isArray(swapHistory)) {
+    for (let s of swapHistory) {
+      if (s && typeof s === "object") {
+        for (let k in s) {
+          const v = s[k];
+          if (v && typeof v !== "string") {
+            throw new Error(`invalid account.swapHistory.${k}`);
```
gre, line 149: Instead of throwing I would prefer we just ignore swapHistory to not prevent us to change the format in future.

src/families/tron/bridge/js.js
juan-cortes: 😕 I dont really get it

gre: See the code

```js
        pendingOperations: maybeExistingSubAccount
          ? maybeExistingSubAccount.pendingOperations
          : [],
```

we didn't do `[]` because in fact we always recreate the sub account reference in Tron case,

this is why there is this comment too

```
  // FIXME: this is bad for perf: we should reconciliate with potential existing data
  // we need to KEEP REF as much as possible & use minimalOperationsBuilderSync
```

so it's really similar for your swapHistory usage, if you set it to [] and do not take into account the `maybeExistingSubAccount` you are going to lose the swaps history.

src/cross.js
juan-cortes: At first invalid, we now set a flag and default to `[]`

src/families/tron/bridge/js.js
juan-cortes: Gotcha, I think
```
swapHistory: maybeExistingSubAccount
   ? maybeExistingSubAccount.swapHistory
   : [],
```

PR #794, opened 2020-07-28
kickoff compound data models

gre:


src/types/operation.js
```diff
@@ -15,7 +15,12 @@ export type OperationType =
   | "UNFREEZE"
   | "VOTE"
   | "REWARD"
-  | "FEES";
+  | "FEES"
+
+  // COMPOUND TYPE OPERATIONS
+  | "SUPPLY"
+  | "REDEEM"
+  | "APPROVE";
```
gre, line 23: I would suggest ALLOW or ALLOW_CONTRACT instead of APPROVE
also, i'm not really sure if we need that concept at all. I think we only could have SUPPLY and REDEEM for now.
The Approvals counter will have to be in a ethereumResources which we forgot to kickoff!

PR #798, opened 2020-07-31
Swap mobile compatibility and export operations

juan-cortes:
Introduces the changes needed to the `initSwap` device action and implements the export operations feature that will be triggered from desktop. This is the result of the peer programming done with @gre that identified a mismatch between the action and command signatures. It also introduces the fix to make cli swap command work again.


juan-cortes:
> linter issue

Fixedd

PR #799, opened 2020-07-31
Algorand

henri-ly:
The PR is still not mergeable, we are still missing :

- the libcore PR to be merged
https://github.com/LedgerHQ/lib-ledger-core/pull/616

- Then, update his bindings for desktop: 
https://github.com/LedgerHQ/lib-ledger-core-node-bindings/pull/82
and mobile: 
https://github.com/LedgerHQ/lib-ledger-core-react-native-bindings/pull/61

- add a new nano for the test-dataset
- enable bot


src/families/algorand/specs.js
```diff
@@ -0,0 +1,237 @@
...
+        // Send the full spendable balance
+        const amount = spendableBalance;
+
+        checkSendableToEmptyAccount(amount, sibling);
+
+        return {
+          transaction: bridge.createTransaction(account),
+          updates: [
+            {
+              recipient: pickSiblings(siblings, 30).freshAddress,
```
gre, line 144: shouldn't it be `sibling.freshAddress` ?

src/families/algorand/specs.js
```diff
@@ -0,0 +1,237 @@
...
+            },
+            { useAllAmount: true },
+          ],
+        };
+      },
+      test: ({ account }) => {
+        const minBalance = getMinBalance(account);
```
gre, line 151: correct me if i'm wrong but we shouldn't have to do this "min balance", that logic that is included in getMinBalance should be implemented as part of the account.spendableBalance. is it not the case?

src/families/algorand/specs.js
```diff
@@ -0,0 +1,237 @@
...
+  }
+};
+
+// Extract asset ID from asset-type subaccount
+let extractAssetId = (subaccount) => {
+  if (subaccount.type !== "TokenAccount") {
+    return null;
+  }
+
+  const assetIdRegex = /algorand\/asa\/(\d+)/g; // e.g, 'algorand/asa/312769'
```
gre, line 52: this is yet another logic to extract out the "id". this really must be factorized into a function

src/families/algorand/tokens.js
```diff
@@ -0,0 +1,43 @@
+// @flow
+
+import type { TokenCurrency } from "../../types";
+import { getCryptoCurrencyById } from "../../data/cryptocurrencies";
+import { addTokens } from "../../data/tokens";
+
+type TokenType = "asa";
+
```
gre, line 8: this file sounds the proper place to add a `extractTokenId` helper

src/libcore/platforms/nodejs.js
```diff
@@ -381,11 +385,17 @@ export default (arg: {
               if (process.env.VERBOSE) {
                 log("libcore-call", id + "." + method, args);
               }
-              const constructorArgs =
-                typeof njsBuggyMethodIsNotStatic === "function"
-                  ? njsBuggyMethodIsNotStatic(args)
-                  : args;
-              const value = new m(...constructorArgs)[method](...args);
+              let value;
```
gre, line 388: mmh why is this now changing?

henri-ly: Static doesn't seems to unwrap hex, so when i was sending hex string to libcore it wasn't converted to binary.
I couldn't find any example in types that needed to convert hex string in static

src/families/algorand/ledger-app/Algorand.js
```diff
@@ -0,0 +1,151 @@
+/********************************************************************************
```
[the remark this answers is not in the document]
henri-ly: I don't think so, but it can be, but I don't know what the requirement to be in ledgerjs repo ? what the minimum functionalities we do need ?

src/env.js
```diff
@@ -33,7 +33,7 @@ const stringParser = (v: mixed): ?string =>
 
 const envDefinitions = {
   API_ALGORAND_BLOCKCHAIN_EXPLORER_API_ENDPOINT: {
-    def: "https://algorand.coin.staging.aws.ledger.com",
+    def: "https://algorand.coin.ledger.com/",
```
gre, line 36: extra slash?

henri-ly: nice catch, thanks !

src/families/algorand/ledger-app/Algorand.js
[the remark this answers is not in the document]
gre: i can deal with it after it's merged, but basically we should make a `@ledgerhq/hw-app-algorand` with what you have so the community can reuse / contribute.

src/families/algorand/libcore-buildSubAccounts.js
```diff
@@ -0,0 +1,132 @@
...
+        if (!blacklistedTokenIds.includes(id)) {
+          existingAccountTickers.push(ticker);
+          existingAccountByTicker[ticker] = existingSubAccount;
+        }
+      }
+    }
+  }
+
+  // filter by token existence
+  promiseAllBatched(3, accountASA, async (asa) => {
```
gre, line 103: misses a `await` to not have race condition and/or finishing earlier than expected the `withLibcore` scope which can have bad consequences on mobile.

i highly suspect this will not work on mobile and was tested only on cli/desktop which is not actually asynchronous.

henri-ly: We do not retest it on mobile will redo it

PR #804, opened 2020-08-03
Failsafe for swaphistory cross export/import

juan-cortes:


src/cross.js
```diff
@@ -164,7 +164,7 @@ const asResultAccount = (unsafe: mixed): AccountData => {
     index,
     balance,
     // $FlowFixMe No idea how to make flow like this
```
gre, line 166: Can you remove this flowfixme then? I imagine it was why it was complaining

src/cross.js
```diff
@@ -164,7 +164,7 @@ const asResultAccount = (unsafe: mixed): AccountData => {
...
-    swapHistory: invalidSwapHistory ? [] : swapHistory,
+    swapHistory: swapHistory || invalidSwapHistory ? [] : swapHistory,
```
gre, line 167: I don't understand the logic. Did you mean `!swapHistory` ?

gre: Let's review tomorrow. I would prefer we satisfy flow, maybe by building up an array, and instead of using a flag, I feel we can simply use that same array to set it to [] or something

src/cross.js
```diff
@@ -334,7 +334,7 @@ export const accountDataToAccount = ({
     index,
     freshAddress,
     freshAddressPath,
-    swapHistory: swapHistory.map(fromSwapOperationRaw),
+    swapHistory: (swapHistory || []).map(fromSwapOperationRaw),
```
gre, line 337: This part makes no sense to me. Why would it be falsy? To me it means the error is somewhere else where we miss to set it as clearly we have defined our type to not be falsy

src/cross.js
juan-cortes: I'll check

src/cross.js
juan-cortes: Indeed

juan-cortes: Older LLD exports without swap types, newer LLM imports with swap types and crashes trying to iterate undefined

src/cross.js
juan-cortes: This is indeed meant to be `!swapHistory` and it does not fix flow, because flow is complaining about the array being mixed and not valid 🤷‍♂️

gre: ok yeah, that's why we need to build up an array

PR #816, opened 2020-08-14
change currency display format

valpinkman:
### Context 

https://ledgerhq.atlassian.net/browse/LL-2855

### DIFF
--- a/src/__tests__/currencies.js
+++ b/src/__tests__/currencies.js
@@ -201,25 +201,25 @@ test("can format a currency unit", () => {
     formatCurrencyUnit(btc, BigNumber(1000000), {
       showCode: true,
     })
-  ).toBe("BTC 0.01");
+  ).toBe("0.01 BTC");
   expect(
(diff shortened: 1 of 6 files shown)


gre:
it seems to fail here

```
FAIL src/families/cosmos/react.test.js (41.424 s)
  ● cosmos/react › useCosmosFormattedDelegations › should return formatted delegations

info Visit https://yarnpkg.com/en/docs/cli/run for documentation about this command.
    expect(received).toBe(expected) // Object.is equality

    Expected: "uatom"
    Received: "478,371"

      46 | 
      47 |       const { code } = getAccountUnit(account);
    > 48 |       expect(result.current[0].formattedAmount.split(" ")[0]).toBe(code);
         |                                                               ^
      49 |       expect(result.current[0].formattedPendingRewards.split(" ")[0]).toBe(
      50 |         code
      51 |       );

```

looks like there is something fishy in the code involved here, i mean `split(" ")[0])` seems to assume too much things. it probably should do `result.current[0].formattedAmount.includes(code)`

PR #819, opened 2020-08-17
Skip the deviceId when on hid

juan-cortes:
Probably not the right way to do this but when connecting via `live-desktop` we get no `deviceId` so this skips the check if hid.


juan-cortes:
Closed in favour of https://github.com/LedgerHQ/ledger-live-desktop/pull/3123

PR #826, opened 2020-08-18
Algorand - Add dataset test cases

Arnaud97234:
Add tests in Algorand dataset:
- Send ASA if Algo balance too low
- OptIn if Algo balance too low

Use accounts from frozen QA seed


src/families/algorand/test-dataset.js
```diff
@@ -5,7 +5,10 @@ import type { DatasetTest } from "../../__tests__/test-helpers/bridge";
 import {
   InvalidAddressBecauseDestinationIsAlsoSource,
   NotEnoughBalance,
+  NotEnoughBalanceInParentAccount,
+  NotEnoughBalanceBecauseDestinationNotCreated,
```
gre, line 9: some lint errors

PR #839, opened 2020-08-27
Algo device transaction opt-in token name issue

LFBarreto:
(Algo): device transaction config optin for asa token name should now be shown on confirmation step


haammar-ledger (approved):
LGTM code-wise.

PR #842, opened 2020-08-28
split fimware hash

valpinkman:
LL-3066

adds a new way to display the hash format during firmware update, to look like what the nano S|X displays

image

### DIFF
--- a/src/manager/index.js
+++ b/src/manager/index.js
@@ -3,6 +3,7 @@
 // Higher level cache on top of Manager
 
 import semver from "semver";
+import chunk from "lodash/chunk";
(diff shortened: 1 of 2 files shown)


IAmMorrow (approved):

src/manager/index.js
```diff
@@ -83,17 +84,32 @@ const CacheAPI = {
     // FIXME these will be made mandatory
     deviceModel?: DeviceModelId,
     deviceInfo?: DeviceInfo
-  ): string => {
+  ): string[] => {
```
gre, line 87: strictly speaking a change of interface be a MAJOR change, not a PATCH bump
and i see you only made a patch semver change mmh

PR #843, opened 2020-08-31
LL-3036 ETH bridge tx status error lag issue

LFBarreto:
(ETH): transaction status bridge patch to ensure not enough balance  is always triggered when needed

(in some cases when estimated fees are triggered too many times the error on the amount isn't updated correctly or may be skipped)

### DIFF
--- a/src/families/ethereum/bridge/libcore.js
+++ b/src/families/ethereum/bridge/libcore.js
@@ -7,6 +7,7 @@ import {
   FeeRequired,
   FeeTooHigh,
   GasLessThanEstimate,
+  NotEnoughBalance,
 } from "@ledgerhq/errors";
 import type { AccountLike } from "../../../types";
 import type { AccountBridge, CurrencyBridge } from "../../../types/bridge";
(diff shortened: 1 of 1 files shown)


src/families/ethereum/bridge/libcore.js
```diff
@@ -135,8 +136,16 @@ const getTransactionStatus = async (a, t) => {
...
 
-  if (!errors.amount && amount.eq(0)) {
-    errors.amount = new AmountRequired();
+  if (!errors.amount) {
+    if (amount.eq(0)) {
+      errors.amount = new AmountRequired();
+    } else if (
+      !tokenAccount
+        ? amount.gt(account.balance.minus(estimatedFees))
+        : amount.gt(tokenAccount.balance) || estimatedFees.gt(a.balance)
```
juan-cortes, line 145: `estimatedFees.gt(a.balance)` should trigger `NotEnoughBalanceInParentAccount` instead of `NotEnoughBalance` cc @LFBarreto

juan-cortes (changes requested):

gre (approved):

PR #854, opened 2020-09-04
Swap - Fix bad BCH check + serialization bug (Object clone)

juan-cortes:
A bad serialization combination between the `icpRenderer` commands and the actions was triggering (albeit inconsistently) an obscure error about not being able to clone an object. This was actually well documented in the breaking changes for Electron 9 https://www.electronjs.org/docs/breaking-changes#behavior-changed-sending-non-js-objects-over-ipc-now-throws-an-exception and has now been addressed. In addition to this, there was a bad check on the `getSerializedAddressParameters` for the bitcoin family introduced by the recent support for BCH, this is also addressed in this PR.

### DIFF
--- a/src/swap/initSwap.js
+++ b/src/swap/initSwap.js
@@ -15,8 +15,7 @@ import network from "../network";
 import { getAccountBridge } from "../bridge";
 import { BigNumber } from "bignumber.js";
 import { SwapGenericAPIError, TransactionRefusedOnDevice } from "../errors";
-import type { Exchange, ExchangeRate, SwapRequestEvent } from "./types";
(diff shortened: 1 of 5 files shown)


src/swap/initSwap.js
```diff
@@ -36,6 +42,13 @@ export type InitSwapInput = {
   deviceId: string,
 };
 
+export type InitSwapInputRaw = {
```
LFBarreto, line 45: Why not export that type in ./types as well? :shrug:

LFBarreto (approved):

LFBarreto (approved):
LGTM :dog:

PR #857, opened 2020-09-07
bump-libcore cli v6.8.10

LFBarreto:
bump libcore cli v6.8.10


LFBarreto:
will be included in #856

PR #865, opened 2020-09-09
Override recipient for estimateMaxSpendable calculation when it's empty

juan-cortes:
Estimate max spendable was broken for ETH because the empty recipient was overriding the one used for the calculation, resulting in an error added to the recipient field which was silenced because it's not a field we have on the swap flow. Reordering the built transaction used to get that transaction status in a way that prioritises the `transaction.recipient` value but defaults to the abandonseed or equivalent recipient makes the estimation work again.


gre (approved):
good catch

PR #873, opened 2020-09-11
Attempt to bump axios

gre:


gre:
maybe something to do around this line https://github.com/axios/axios/issues/1754#issuecomment-562836049

gre:
now appears to work 🎊

PR #874, opened 2020-09-11
LL-3152 (ConnectApp): list apps logic added on regular flow

LFBarreto:
(ConnectApp): list apps even if no experimental flow is enabled and detect if app requested is installed


gre:
internal discussion. to be clarify what is solved here.

PR #877, opened 2020-09-14
Update asa.js

jtmarkley1:
Updating with new assets such as USDC


FabriceDautriat (commented):
LGTM (need to confirm the ASA details by fetching information from node)

gre (approved):

FabriceDautriat:
Version 1.2.9 of the Algorand app is mandatory, please make sure to update 
See https://github.com/LedgerHQ/ledger-live-common/pull/878

gre:
Thanks for the PR,

as the device also includes the tokens data and old version would be incompatible with this new list we also need to make sure this is used with the latest app.

to do so, in src/apps/support.js, we have a appVersionsRequired map that will express what required version is needed, can you update it to `>= 1.2.9` please?

thanks

gre:
oh i didn't see it was another PR. ok we can merge both.

PR #883, opened 2020-09-18
Swap - alias horizen to zencash ids

juan-cortes:


gre (changes requested):
to be fixed in crypto-assets where we need to agree on the same id. there are many consequence on top of swap if we start diverging on ids.

juan-cortes:
Will be fixed in a less hacky way 🚀

PR #888, opened 2020-09-24
@ledgerhq/cryptoassets

gre:
Goes with https://github.com/LedgerHQ/ledgerjs/pull/529

This moves many of the crypto assets data and importers into a dedicated ledgerjs library.

## Context

LL-3336

### as a user of live-common

- All imports from `@ledgerhq/live-common/lib/data/*` can be safely changed to `@ledgerhq/live-common/lib/currencies` (it should have been the case). If it's a type, it can be from `@ledgerhq/live-common/lib/types`.
- All imports from `@ledgerhq/live-common/lib/load/*` can be safely removed. This make our library a bit more stateless and pure.

### as a developer of live-common

- any crypto assets data (addition, edit) no longer exists in live-common and will happen directly in @ledgerhq/cryptoassets library and will progressively be moved to crypto-assets repository itself with ledgerjs lib acting as an "importer" of crypto-assets. This is also a future possibility that crypto-assets would be fetched dynamically, but we would still need the "initial state" that @ledgerhq/cryptoassets would offer


LFBarreto (approved):

LFBarreto (approved):
Set a reminder to remove all the tokens load imports on LLD and LLM once this is bumped there :+1:

gre:
@LFBarreto there are already all the PR created on LLM and LLD

valpinkman (approved):

henri-ly (approved):

PR #891, opened 2020-09-28
(AppHW): fix loading after connect app polling

LFBarreto:
After a successfull app connect device becomes unresponsive we now show a loading between this state and getting the derivation ﻿for the accounts.

### DIFF
--- a/src/hw/actions/app.js
+++ b/src/hw/actions/app.js
@@ -81,7 +81,8 @@ type Event =
   | { type: "error", error: Error, device?: ?Device }
   | { type: "deviceChange", device: ?Device }
   | ConnectAppEvent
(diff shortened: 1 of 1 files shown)


gre:
(i'm setting this on hodl until next sprint, to focus our QA on the essential parts)

gre:
@LFBarreto what do you think is the status of this PR? should this land on master, does it require some extensive QA?

LFBarreto:
> @LFBarreto what do you think is the status of this PR? should this land on master, does it require some extensive QA?

I think we need to QA this and check if that doesn't break anything. I'm pretty sure the impact is minor but helps on the UX of the app flow

gre:
to be revisited.

PR #893, opened 2020-09-28
LL-3397 Prevent fees from being the same

juan-cortes:
This code will prevent us from offering feePerByte values that are the same but have misleadingly different labels.
It will work as follows, given we don't set a different value for `speedStepIfEqual`. 

```[0,1,2] => [0,1,2]
[0,0,1] => [0,1,2]
[0,0,3] => [0,1,3]
[1,1,1] => [1,2,3]
```


gre:
❓

PR #894, opened 2020-09-28
LL-3397 Prevent fees from being the same

juan-cortes:
This code will prevent us from offering feePerByte values that are the same but have misleadingly different labels.
It will work as follows.

```
[1,1,1] => [3,2,1]
[3,2,2] => [4,3,2]
[4,4,3] => [5,4,3]
```


gre:
Description says it's doing

[3,2,2] => [4,3,2]

Looking at the code, I don't see that, maybe I missed something

gre:
it's good enough anyway, it's ok to be +2 fee per byte pessimistic to avoid the confusion of having same fee everywhere

PR #897, opened 2020-09-30
add descriptor logic

gre:
add helpers to iterate on xpubs and descriptors.

also add `ledger-live scanDescriptors`

### DIFF
--- a/src/derivation.js
+++ b/src/derivation.js
@@ -1,8 +1,19 @@
 // @flow
 import invariant from "invariant";
+import { Observable, defer, of, range, empty } from "rxjs";
+import {
+  catchError,
+  switchMap,
+  concatMap,
+  takeWhile,
+  map,
+} from "rxjs/operators";
+import { log } from "@ledgerhq/logs";
+import { TransportStatusError, UserRefusedAddress } from "@ledgerhq/errors";
 import type { CryptoCurrency, CryptoCurrencyConfig } from "./types";
 import { getCryptoCurrencyById } from "./currencies";
 import { getEnv } from "./env";
+import type { GetAddressOptions, Result } from "./hw/getAddress/types";
 
 type LibcoreConfig = {
   [_: string]: mixed,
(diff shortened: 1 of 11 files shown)


src/families/bitcoin/descriptor.test.js
```diff
@@ -0,0 +1,67 @@
...
+import invariant from "invariant";
+import { getCryptoCurrencyById } from "../../currencies";
+import {
+  inferDescriptorFromDeviceInfo,
+  inferDescriptorFromAccount,
+} from "./descriptor";
+import { fromAccountRaw } from "../../account";
+import bitcoinDatasets from "./datasets/bitcoin";
+
+describe("inferDescriptorFromAccount", () => {
```
onyb, line 11: Can you add some tests for:
* wrapped segwit (`49'/0'/x'`)
* testnet accounts (`x'/1'/y'`)

src/derivation.js
```diff
@@ -335,6 +346,20 @@ export const runDerivationScheme = (
...
+export const runAccountDerivationScheme = (
+  scheme: string,
+  currency: { coinType: number },
+  opts: {
+    account?: number | string,
+  } = {}
+) =>
+  runDerivationScheme(scheme, currency, {
+    ...opts,
+    address: "_",
```
juan-cortes, line 359: What's the logic for passing an underscore here 🤔

juan-cortes (approved):

src/derivation.js
gre: We only want the account part of the derivation path, so I used _ as a way to identify which part is not account based. It's a bit overkill but it works

src/families/bitcoin/descriptor.test.js
gre: duly noted. At the moment however, we don't have segwit account in datasets/bitcoin and i'll wait @Arnaud97234  to provide one for future testing this.
we also don't have testnet datasets

as we try to keep these dataset static and from same QA see we'll add more test a bit later. this was tested on CLI and should work. There is also some chance that more functions will be needed when we will do the LLD integration, so this PR should be good to go as first kickoff of the logic.

PR #899, opened 2020-10-02
Swap - Split the out of range error for rates into two

juan-cortes:
image

This would allow us to do what's shown on the screenshot after a bump, so the error message is shorted and we don't have UI issues on the LLM of overflowing messages. Also if the user is out of bounds it's going to be from one of the sides, not going to be too high first and too low after

### DIFF
--- a/src/swap/getExchangeRates.js
+++ b/src/swap/getExchangeRates.js
@@ -8,7 +8,10 @@ import network from "../network";
 import { getSwapAPIBaseURL } from "./";
 import { getEnv } from "../env";
 import { BigNumber } from "bignumber.js";
-import { SwapExchangeRateOutOfBounds } from "../errors";
+import {
+  SwapExchangeRateAmountTooLow,
+  SwapExchangeRateAmountTooHigh,
+} from "../errors";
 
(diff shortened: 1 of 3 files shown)


src/swap/getExchangeRates.js
```diff
@@ -40,11 +43,17 @@ const getExchangeRates: GetExchangeRates = async (
...
+
+        throw isTooSmall
+          ? new SwapExchangeRateAmountTooLow(null, {
+              unit: unitFrom.code,
+              minAmountFrom,
+            })
+          : new SwapExchangeRateAmountTooHigh(null, {
+              unit: unitFrom.code,
+              maxAmountFrom,
+            });
```
gre, line 56: how can we be sure the else case is a too high case? i'm a bit lost by this because it's the `!rate` case, there is no other possible issue?

juan-cortes: If you look at the api definition, it's either a rate or a bad amount response on this endpoint https://github.com/LedgerHQ/swap/blob/master/doc/openapi-swap.yaml#L49 so that's what I'm basing the code logic on.

gre:
lint issue

PR #907, opened 2020-10-12
Added sell crypto device commands and logic

IAmMorrow:
This PR include all the required logic for the sell crypto feature. It also feature an update on the swap code since both of those logics have been factorized together.


src/hw/actions/initSell.js
```diff
@@ -0,0 +1,246 @@
...
+                });
+
+                bridge.prepareTransaction(mainAccount, transaction).then(() => {
+                  bridge
+                    .getTransactionStatus(mainAccount, transaction)
+                    .then((status) => {
+                      setCoinifyContext({ context, transaction, status });
+                    });
+                });
+              });
```
gre, line 171: The promise that goes out on a tap asynchronous, if there is an error would never get seen by the user land, so this would need to be turned to a mergeMap or similar

gre (commented):
One feedback in the observable

PR #908, opened 2020-10-16
migrate swapHistory for keeping the history after migrating to ETHJS

gre:
- [x] implement `implicitMigration: Account[] => Account[]` that needs to be called by LLD and LLM just after loading the data from store.
- [x] implement a few tests with real dataset (that includes at least these swaps: TOKEN to bitcoin, bitcoin to TOKEN, ethereum to bitcoin, ethereum to TOKEN, bitcoin to litecoin (to ensure it doesn't change))
- [x] run the function in all existing dataset tests so we ensure it doesn't break somewhere.


gre:
@juan-cortes i got this implemented by did not test at all. We need to write some tests for it, do you have a bunch of swap history accumulated so we can try to hardcode accounts in a test and make sure the function is doing what is expected?

juan-cortes (approved):

juan-cortes:
Should have enough tests for the swap History part imho

gre:
- [ ] test that verify the "stability" of running twice the migrate function (to make sure it doesn't "leak" some duplications)
- [ ] since we saw that the decodeTokenAccountId can throw, need to review all usecase of it and protect the code to not crash (typical usecase is a token that gets removed, the tokenId will not resolved so the function might fail... i think that means we probably should make this function returning an option instead of crashing)

gre:
✅  USDT -> BTC
❌  USDT -> USDC

seems like the "to token" does not work.

gre:
ok, this was false positive. i didn't reboot my Live before migrating.

PR #914, opened 2020-10-21
[INFRA-2261] Add Digibyte to available explorers v3

lgillot-lamure-ledger:
Url would be `https://explorers.api-01.live.ledger-stg.com/blockchain/v3/dgb/` in staging (not available in prod yet).

### DIFF
--- a/src/api/Ledger.js
+++ b/src/api/Ledger.js
@@ -73,6 +73,7 @@ const ledgerExplorersV3 = {
   ethereum_ropsten: "eth_ropsten",
   ethereum_classic: "etc",
   litecoin: "ltc",
+  digibyte: "dgb",
   // tezos: "xtz"
 };
 


lgillot-lamure-ledger:
@gre does that look correct, are the ticker and the URL ok ?

gre:
- by putting `digibyte: "dgb"` in `ledgerExplorersV3`, it enables the possibility for Live (via libcore) to use `explorers.api.live.ledger.com/blockchain/v3/dgb/*` endpoints.
- the fact it's still in `ledgerExplorersV2` also means `explorers.api.live.ledger.com/blockchain/v2/dgb/*` is used by default but if user enables in Ledger Live the "Experimental Explorers" settings, it will switch to `explorers.api.live.ledger.com/blockchain/v3/dgb/*`

This is currently the only granularity for our end user to test an experimental explorer, Live don't support natively the idea of "staging", unless if we want to start having a "per coin" URL at some point in the future, but today we work on this simple take that the URL is just `explorers.api.live.ledger.com/blockchain/${version}/${explorerId}/*`.

So in short, I can't merge this PR until this goes in production OR if we reconsider the way we work toward testing explorers.

lgillot-lamure-ledger:
Hi. Ah… well then it means you need to remove LTC from the V3 list, as it doesn’t have a prod endpoint either (yet).

So how can you guys test a new v3 coin ? Because we’re supposed to wait for your testing before promoting it to prod, but you need us to promote it to prod, at this point.

Of course if the solution is "QA uses a branch of LL", then it’s fine and we’ll just keep this PR laying around as a Draft for this purpose :).

gre:
I guess it is because it used to be at some point.

It doesn't matter if it is not defined because it's acceptable that coins in Experimental explorers are broken. However not having prepared coins like dgb make it impossible to test in current live. So I almost tend to think we would need them all. And maybe we can make the checkbox also using staging / have another checkbox to switch to staging...

 I think we need to make the bot/dataset tests run in whatever is the most Experimental and available (even staging 🤔).

need to think more.. will see if we can set up an easy solution

lgillot-lamure-ledger:
Ok, thanks for handling it !

gre:
https://github.com/LedgerHQ/ledger-live-common/pull/918 will trigger the CI with the new config

PR #920, opened 2020-10-23
LL-2672 update ripple lib

henri-ly:


gre:
we will HODL this until more QA on CI team side and will consider including it not in our next release but potentially more in LL 2.17

PR #922, opened 2020-10-26
LL-3765 (useBridgeTransaction): fix tx mode reseting to send

LFBarreto:
(useBridgeTransaction): fix tx mode reseting to send when setAccount is used﻿


gre (approved):
good for now. 
note for future: we might consider that a setAccount will never reset the transaction, unless the account currency change from prev tx. not easy to do for now, will come back later.

PR #925, opened 2020-10-27
fix(bridge): complete scan when unsubscribed

rjarasson-ledger:
When a scan is started, it can be aborted prematurely (unsubscribed).
To avoid scan to run additional steps, i added loop breaks to ensure scan is stopped as soon as possible.

It fixes a bug in CLI when using `-i` argument, where scan sends an additionnal ADPU to device before finishing, which conflicts with other calls to device.

### DIFF
--- a/src/bridge/jsHelpers.js
+++ b/src/bridge/jsHelpers.js
@@ -184,6 +184,8 @@ export const makeScanAccounts = (
       derivationMode,
       seedIdentifier
     ): Promise<?Account> {
+      if (finished) return;
+
       const accountId = `js:2:${currency.id}:${address}:${derivationMode}`;
       const accountShape: Account = await getAccountShape(
         {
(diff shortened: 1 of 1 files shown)


src/bridge/jsHelpers.js
```diff
@@ -193,7 +195,6 @@ export const makeScanAccounts = (
         },
         syncConfig
       );
-      if (finished) return;
```
gre, line 196: this line is still worth keeping imo

src/bridge/jsHelpers.js
```diff
@@ -329,7 +334,7 @@ export const makeScanAccounts = (
                   : "no account"
               }`
             );
-            if (!account) return;
+            if (!account) break;
```
gre, line 337: imo we shouldn't change that. we should still return because the only case it's `!account` is when finished happened

gre (commented):
quick feedback (tell me if i'm wrong)

src/bridge/jsHelpers.js
rjarasson-ledger: Ok, so it's ok if o.complete() is never called then ?

gre: if the consumer has unsubscribed, all observable events (next, error, complete) are ignored completely by rxjs.

PR #927, opened 2020-10-28
add amount to result in fillTransactionStatus

valpinkman:


src/families/ethereum/modules/erc20.js
```diff
@@ -38,6 +38,7 @@ const erc20approve: ModeModule = {
     if (!t.useAllAmount && t.amount.eq(0)) {
       result.errors.amount = new AmountRequired();
     }
+    result.amount = t.useAllAmount ? BigNumber(2).pow(256).minus(1) : t.amount;
```
LFBarreto, line 41: just use t.amount here the UI handles the no limit

src/families/ethereum/modules/compound.js
```diff
@@ -70,6 +70,7 @@ const compoundSupply: ModeModule = {
     } else if (t.amount.gt(subAccount.spendableBalance)) {
       result.errors.amount = new NotEnoughBalance();
     }
+    result.amount = t.useAllAmount ? a.balance : t.amount;
```
LFBarreto, line 73: i also think using t.amount here is also safe as is for withdraw and enable

src/families/ethereum/modules/erc20.js
gre: `BigNumber(2).pow(256).minus(1)` isn't it a constant somewhere?

valpinkman: I think we haven't set it up but I will

LFBarreto (approved):
:+1:

gre (approved):

PR #933, opened 2020-10-30
Ll 3715 - wallet connect cli/poc

machard:
can try on https://example.walletconnect.org/

`ledger-live walletconnect -i 0 -c ethereum --walletConnectURI "QRCODEDATA"`
or 
`Ledger-live walletconnect -i 0 -c ethereum --walletConnectSession "sessionJSONString"`
sessionJSONString is provided in the log when using QRCODEDATA just under the line `connected`


cli/src/commands/walletconnect.js
```diff
@@ -0,0 +1,286 @@
...
+      desc: "verbose mode",
+    },
+    {
+      name: "silent",
+      type: Boolean,
+      desc: "do not output the proxy logs",
+    },
+  ],
+  job: (opts: Opts) =>
+    Observable.create(async (o) => {
```
gre, line 53: minor note: Observable.create can't take an async function. the unsub returned value will not be used.
you could eventually put the logic out in an async function outside, which is going to be the case when things get stabilized back into live-common

cli/src/commands/walletconnect.js
```diff
@@ -0,0 +1,286 @@
...
+                derivationMode: account.derivationMode,
+              };
+
+              log("walletconnect", "message to sign");
+              log("walletconnect", (message: any));
+
+              result = await withDevice(opts.device || "")((t) =>
+                from(signMessage(t, message))
+              )
+                .pipe(take(1))
```
gre, line 139: this is not needed. toPromise() implicitly is taking the first element already.

cli/src/commands/walletconnect.js
```diff
@@ -0,0 +1,286 @@
...
+                .toPromise();
+              result = result.signature;
+
+              log("walletconnect", "message signature");
+              log("walletconnect", result);
+
+              break;
+            case "eth_signTransaction":
+            case "eth_sendTransaction":
+              transactionRaw = payload.params[0];
```
gre, line 149: wording choice: probably use another name than "transactionRaw" to not confuse with the `TransactionRaw` type in live-common.

cli/src/commands/walletconnect.js
```diff
@@ -0,0 +1,286 @@
...
+          log("walletconnect", approval);
+
+          connector.approveRequest(approval);
+        } else {
+          let message = "JSON RPC method not supported";
+          if (errorMsg) {
+            message = errorMsg;
+          }
+
+          const rejection: any = {
```
gre, line 255: casting something to `any` hides type error. to avoid.

gre (commented):
nice progress

machard:
I'm not sure I can put more stuff in common

PR #935, opened 2020-11-02
Expose useSupplyMax and export const useSupplyMaxChoiceButtons

gre:
- This should allow to address https://ledgerhq.atlassian.net/browse/LL-3810 on LLD and LLM
- This also fixes a bug that could have created issue: the value must be rounded to integer.


valpinkman (approved):
YES

PR #945, opened 2020-11-06
sats stack business logic functions

gre:
Implementation of helpers for integration of https://github.com/LedgerHQ/satstack in Ledger Live.


gre:
- [ ] introduce new error for when SATSTACK but not success status in a Bitcoin sync/scan accounts

PR #892, opened 2020-09-28
LL-2135 New Countervalues module

JunichiSugiura:

PR #946, opened 2020-11-17
LL-3885 Integrate quitApp into manager device action

juan-cortes:
Introduce support (when applicable) for the quitApp command that will automatically exit the current app if it's not the one expected in the current flow. This was already in use in the `connectApp` device action, this pr introduces it to flows that use the `connectManager` device action, in particular it affects direct manager access, swap feature and sell feature/

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -195,11 +195,6 @@ const envDefinitions = {
     parser: boolParser,
     desc: "enable an experimental implementation of USB support",
   },
-  EXPERIMENTAL_QUIT_APP: {
-    def: false,
-    parser: boolParser,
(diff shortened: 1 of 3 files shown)


juan-cortes:
@gre I tested on both mobile and desktop, should be fine for QA from my side

gre:
Ok thanks! We'll merge this and this will go in our next major. potentially as part of #892 and #920 which will all drive more QA involvement and a decent "major" bump update.

we will need to remember we need to drop the experimental on LLD/LLM that had this quit app when we bump there. in that sense we can consider it a "breaking change" and therefore consolidate this "major" requirement.

gre:
I'm HODLing the PR until countervalues gets in first.

PR #947, opened 2020-11-18
LL-3497 : add algorand subOperation

henri-ly:


gre:
Wouldn't the top level type need to be a "FEE" operation? I'm not sure I read the diff correctly but it sounds like you have twice the same type on parent and child opération, does that respect the non-optomistic version of this tx anyway? 🤔

Reminds me we would deserve to add such tests in our bot generic test.. I'll plan it

PR #953, opened 2020-11-18
Add missing `hold` op status

juan-cortes:
This status (and `overdue` although it doesn't seem to be a real status) as defined in the api definition https://github.com/LedgerHQ/swap/blob/master/doc/openapi-swap.yaml#L179-L189 was missing and would make users facing AML/KYC procedures not have an icon on their swap history/details which adds insult to injury. It's a fairly rare operation status to fall into but that doesn't justify ignoring it.


gre (changes requested):
Prettier

PR #955, opened 2020-11-19
Fixes sat-stack naming

gre:
> “i dont mind, but it the app is called satstack, we should respect that”
– someone


juan-cortes (approved):

onyb (approved):
Putting stamp of approval on a surface ( "OK" )

PR #958, opened 2020-11-20
fix(tests): hydrate after preload with serialized data

rjarasson-ledger:
`hydrate` is always called after a `preload`, but with a JSON serialized version.
Next preload should result in the same payloads (after deserialization of the hydrated version).

If further preload makes an iterative update from hydrated data, it should not alter data (or reuse a cached version).

### DIFF
--- a/src/__tests__/test-helpers/bridge.js
+++ b/src/__tests__/test-helpers/bridge.js
@@ -291,14 +291,16 @@ export function testBridge<T>(family: string, data: DatasetTest<T>) {
 
       test("preload and rehydrate", async () => {
         const data1 = await bridge.preload(currency);
-        const serialized = JSON.stringify(data1);
         if (data1) {
-          expect(serialized).toBeDefined();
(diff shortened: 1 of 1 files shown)


rjarasson-ledger:
I did not tested it against all currencies, so it might report undetected behaviours (that would need fix).

src/__tests__/test-helpers/bridge.js
```diff
@@ -291,14 +291,15 @@ export function testBridge<T>(family: string, data: DatasetTest<T>) {
 
       test("preload and rehydrate", async () => {
         const data1 = await bridge.preload(currency);
-        const serialized = JSON.stringify(data1);
+        const serialized1 = JSON.parse(JSON.stringify(data1));
```
gre, line 294: data1 can be undefined which will make this line crash.

gre (changes requested):
see comment.

gre:
thanks Rémi!

PR #959, opened 2020-11-23
(do not merge) Test LTC v3 staging

gre:


gre:
all good. just a permutation of some addresses, not an issue.

PR #961, opened 2020-11-23
walletconnect tests

machard:


gre (changes requested):
CI reveals flow issues

Issue #734, opened 2020-06-15
unify speculos-deviceActions, getTransactionConfig and formatTransaction


gre:
something good can be done to join these 3 topics.

PR #963, opened 2020-11-25
[Coin-integration] Polkadot

rjarasson-ledger:
Full integration of Polkadot.

TODO: description and remaining tasks


src/families/polkadot/bridge/js.js
```diff
@@ -0,0 +1,102 @@
...
+const sameFees = (a, b) => (!a || !b ? a === b : a.eq(b));
+
+const prepareTransaction = async (a, t) => {
+  let fees = t.fees;
+
+  fees = await calculateFees({
+    a,
+    t: {
+      ...t,
+      recipient: "1Z4QdzRrpVbggYoGK5pfbeMyzpVVDK7WxheVjWFxfv6sxjV", //Empty account to calculate fees
```
gre, line 62: - first of all, i'm curious why we do this here. I expect fees to not be calculated if you don't have set a recipient (consistent with other impl), it's ok tho to keep it (because in any case, you need it to probably estimate max) 
- just a good internal practice, but we prefer this to use a "public seed" so we can "prove" that we don't own this address. we usually will just use the `abandon ..` test seed to make it clear.
- also you can put it in here => https://github.com/LedgerHQ/ledgerjs/blob/master/packages/cryptoassets/src/abandonseed.js

src/families/polkadot/deviceTransactionConfig.js
```diff
@@ -0,0 +1,211 @@
...
+
+    case "nominate":
+      fields.push({
+        type: "text",
+        label: "Staking",
+        value: "Nominate",
+      });
+
+      fields.push({
+        type: "polkadot.validators",
```
gre, line 90: just FYI: we will work forward to the idea that we might no longer will allow such type. it will likely be only "text" in the future so we could reuse it for the bot logic etc.. (context of #734 )
but in any case, it's ok to keep this for now. i just wonder: is there a specific UI on LLD that would requires to not do a list of "text" ?

src/families/polkadot/js-estimateMaxSpendable.js
```diff
@@ -0,0 +1,71 @@
...
+  parentAccount: ?Account,
+  transaction: ?Transaction,
+}): Promise<BigNumber> => {
+  const mainAccount = getMainAccount(account, parentAccount);
+  const t = {
+    ...createTransaction(),
+    ...transaction,
+    recipient:
+      transaction?.recipient ||
+      "1Z4QdzRrpVbggYoGK5pfbeMyzpVVDK7WxheVjWFxfv6sxjV", // need abandon seed and being empty
```
gre, line 28: same comment as before. need to move in ledgerjs.

src/families/polkadot/js-signOperation.js
```diff
@@ -0,0 +1,174 @@
...
+): Operation => {
+  const type = MODE_TO_TYPE[transaction.mode] ?? MODE_TO_TYPE.default;
+
+  const value =
+    type === "OUT" ? BigNumber(transaction.amount).plus(fee) : BigNumber(fee);
+
+  const extra = getExtra(type, account, transaction);
+
+  const operation: $Exact<Operation> = {
+    id: `${account.id}--${type}`,
```
gre, line 72: can use encodeOperationId helper too.

gre (changes requested):
giving some first 👀  feedbacks.
overall good job.
BTW i didn't really looked at some of the complex business logic as i'm not qualified on the polkadot logic itself.

src/api/polkadot/bisontrails.js
```diff
@@ -0,0 +1,211 @@
...
+    blockHeight: reward.blockNumber,
+    date: new Date(reward.timestamp),
+    extra: getExtra(type, reward),
+    senders: [reward.validatorStash].filter(Boolean),
+    recipients: [reward.accountId].filter(Boolean),
+  };
+};
+
+const slashToOperation = (addr, accountId, slash) => {
+  const hash = `${slash.blockNumber}-0`; // to be compatible with explorer
```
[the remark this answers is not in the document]
rjarasson-ledger: Yes, i suggest using "+" plus sign

src/api/polkadot/bisontrails.js
```diff
@@ -0,0 +1,211 @@
...
+import network from "../../network";
+import https from "https";
+
+import { BigNumber } from "bignumber.js";
+
+import { getEnv } from "../../env";
+import { getOperationType } from "./common";
+
+// TMP
+const agent = new https.Agent({
```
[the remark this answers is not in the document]
rjarasson-ledger: Yes, and also because it's only because of BisonTrails having a self-signed SSL certificate on Staging. This is blocking on our side for mobile, so we are using subscan.

src/api/polkadot/bisontrails.js
```diff
@@ -0,0 +1,211 @@
...
+  const operations = data.extrinsics.map(
+    extrinsicToOperation.bind(null, addr, accountId)
+  );
+
+  const rewards = data.rewards.map(
+    rewardToOperation.bind(null, addr, accountId)
+  );
+
+  const slashes = data.slashes.map(
+    slashToOperation.bind(null, addr, accountId)
```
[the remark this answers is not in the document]
rjarasson-ledger: Oh, sorry, i suggested this way of doing things... my bad @henri-ly

src/api/polkadot/bisontrails.js
```diff
@@ -0,0 +1,211 @@
...
+  );
+
+  const mergedOp = [...prevOperations, ...operations, ...rewards, ...slashes];
+
+  if (
+    operations.length < LIMIT &&
+    rewards.length < LIMIT &&
+    slashes.length < LIMIT
+  ) {
+    return mergedOp.sort((a, b) => b.date - a.date);
```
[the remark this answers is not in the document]
rjarasson-ledger: mergedOp is a mix of different types, so they are not sorted once concatenated.
I think this is necessary, though, this call should be incremental, so i don't think there is a real performance issue there.

src/families/polkadot/bridge/js.js
rjarasson-ledger: I suggested @henri-ly to explicitely have a non-existing account and visually understandable as such.
Good idea to put this in an abandon address.

This is because sometimes, user fills an erroneous address as recipient... this breaks fees estimation.
Also, if not filled, this can lead to shorter transactions, giving wrong fees.
I suggested to hardcode address in this case, so that it could also benefit fees estimation cache (since changeing recipient won't affect fees).

src/families/polkadot/deviceTransactionConfig.js
rjarasson-ledger: There is a specific component yes on LLD and LLM.
This is an array of addresses, so we will need a way to override UI element for this field if this is Generically interpreted as StringArray.
I think this would be easy though to find a solution, by changing the way TransactionConfirm component works on front-end (use something like "key" instead of type maybe for rendering, with "type" being a default rendering).

For instance, field definition for polkadot validators could be:
```
{ type: "StringArray", key: "validators", "value": [....], "label": "Targets" }
```

src/api/polkadot/bisontrails.js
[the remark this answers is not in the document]
henri-ly: The sign "-" is what the explorer need to find the right extrinsic for the event reward / slash : 
We wouldn't be able to redirect with view in details
https://polkadot.subscan.io/extrinsic/2737842-1

src/families/polkadot/deviceTransactionConfig.js
gre: for now, let's keep just what you had (polkadot.validators) and this is a topic that I need to work on for all coins (because many coins integration did their own type) and see how we want it to be done. (because we could also just go with doing as .push as there are validators, as they all end up with screens on the device at the end)

valpinkman (approved):
Reviewed the PR
Everything LGTM

src/__tests__/all.libcore.js
```diff
@@ -14,6 +14,12 @@ import specifics from "../generated/test-specifics";
 import type { DatasetTest } from "../types";
 import { disconnectAll } from "../api";
 
+// Polkadot Api is very verbose...
+jest.mock("@polkadot/util/logger.js", () =>
```
rjarasson-ledger, line 18: unnecessary anymore i think

src/env.js
```diff
@@ -57,6 +57,22 @@ const envDefinitions = {
     def: "wss://xrpl.ws",
     desc: "websockets api for ripple",
   },
+  API_POLKADOT_INDEXER: {
+    parser: stringParser,
+    def: "https://polkadot.coin.ledger.com",
+    // def: "https://polkadot.subscan.io",
+    desc: "Explorer API for polkadot",
+  },
+  API_POLKADOT_NODE: {
```
rjarasson-ledger, line 66: unnecessary anymore

PR #964, opened 2020-11-26
Allow to disable auto quit app when needed on manager access

juan-cortes:
When a user accesses the manager flow and is already inside an app, the new auto quit app logic will exit the given app (if possible) and continue the flow, listing apps, etc and entering the manager. There is one case where the user might have accessed the app while inside the manager. 

The current logic will exit the manager, quit the app, and access the manager again. This pr provides a way to flag this second entry as to not trigger an automatic exit of the app and merely prompt the user to exit the app to access the manager again.

### DIFF
--- a/src/hw/actions/manager.js
+++ b/src/hw/actions/manager.js
@@ -59,6 +59,10 @@ type ManagerState = {|
   closeRepairModal: () => void,
 |};
 
+export type ManagerRequest = ?{
+  autoQuitAppDisabled?: boolean,
+};
+
 type Result = {|
   device: Device,
   deviceInfo: DeviceInfo,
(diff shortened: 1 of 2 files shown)


src/hw/actions/manager.js
```diff
@@ -59,6 +59,10 @@ type ManagerState = {|
   closeRepairModal: () => void,
 |};
 
+export type ManagerRequest = {
+  autoQuitAppDisabled?: boolean,
```
gre, line 63: Can you try to
Make it a ?{
In order to not be breaking changes

gre (changes requested):

PR #970, opened 2020-11-30
LL-3501 fix force provider default value

machard:
https://ledgerhq.atlassian.net/browse/LL-3501

cf
https://github.com/LedgerHQ/ledger-live-mobile/pull/1486

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -216,7 +216,7 @@ const envDefinitions = {
       "force an old application version to be accepted regardless of its version",
   },
   FORCE_PROVIDER: {
-    def: 1,
+    def: false,
     parser: intParser,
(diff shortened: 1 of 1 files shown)


gre (changes requested):
I don't think we should change this.

1 is the default.

to know that the checkbox is not enabled, you actually compare the default and the value.

we managed to solve it on LLD so i don't think we need a change in live-common to solve this on LLM as well.

https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/screens/settings/sections/Experimental/ExperimentalInteger.js

https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/screens/settings/sections/Experimental/index.js#L43

PR #973, opened 2020-12-04
LL-3916 allow bridge to be able to define preload() refresh time

gre:
- allow bridge to be able to define `getPreloadConfig`
- define an helper `makeBridgeCacheSystem` that we can reuse on LLD/LLM to generate prepareCurrency and hydrateCurrency concepts


gre:
@henri-ly @rjarasson-ledger @juan-cortes any of you can review?

PR #974, opened 2020-12-04
LL-4001 check LSS version

gre:
LL-4001 check LSS version

+ add "satstack-outdated" status
+ throw SatStackVersionTooOld in syncs

LL-4002 add check that descriptor are imported in sync
- add error SatStackDescriptorNotImported

### DIFF
--- a/src/families/bitcoin/satstack.js
+++ b/src/families/bitcoin/satstack.js
@@ -1,8 +1,10 @@
 // @flow
 import { Observable, interval, from } from "rxjs";
+import semver from "semver";
 import url from "url";
 import { share, switchMap } from "rxjs/operators";
 import {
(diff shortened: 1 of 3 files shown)


src/families/bitcoin/satstack.js
```diff
@@ -16,6 +18,12 @@ import type { AccountDescriptor } from "./descriptor";
 import { getEnv } from "../../env";
 import { getCurrencyExplorer } from "../../api/Ledger";
 
+const minVersionMatch = ">=0.10.0";
```
onyb, line 21: Can we make sure the minimum version is v0.11.1-beta, since it fixes the bug with auto detection of config file?

juan-cortes (approved):

PR #975, opened 2020-12-08
Fetch countervalues at least for the past one year

JunichiSugiura:
Load countervalues for a year instead of account.creationDate if the creationDate is less than a year ago.


juan-cortes (approved):
lgtm

PR #979, opened 2020-12-10
xrp: fix queued transaction

henri-ly:


gre:
is it easy to identify an operation that is still "in queue"? to me it means it's not yet in a block, or is it? because the way we could solve it is by setting null to blockHeight

that would fix the fact that Live says "confirmed" even when it's actually still "in queue".

wdyt?

src/families/ripple/bridge/js.js
```diff
@@ -183,10 +183,16 @@ const broadcast = async ({ signedOperation: { signature, operation } }) => {
   const api = apiForEndpointConfig(RippleAPI);
   try {
     await api.connect();
-    const submittedPayment = await api.submit(signature);
+    const submittedPayment = await api.request("submit", {
+      tx_blob: signature,
+      fail_hard: false,
```
juan-cortes, line 188: Isn't that the default? https://xrpl.org/submit.html#main_content_body

src/families/ripple/bridge/js.js
```diff
@@ -183,10 +183,16 @@ const broadcast = async ({ signedOperation: { signature, operation } }) => {
...
+    });
 
-    if (submittedPayment.resultCode !== "tesSUCCESS") {
-      throw new Error(submittedPayment.resultMessage);
+    if (
+      submittedPayment.engine_result !== "tesSUCCESS" &&
+      submittedPayment.queued !== true
+    ) {
+      throw new Error(submittedPayment.engine_result_message);
```
juan-cortes, line 195: Maybe not for now, but we could potentially map the errors here for more granularity 🤔 
cc @gre

juan-cortes (approved):

src/families/ripple/bridge/js.js
henri-ly: Yes is it

henri-ly:
> is it easy to identify an operation that is still "in queue"? to me it means it's not yet in a block, or is it? because the way we could solve it is by setting null to blockHeight
> 
> that would fix the fact that Live says "confirmed" even when it's actually still "in queue".
> 
> wdyt?

I don't think that operations will be in queue if it was synced it should be just as a normal operation.

The `in queue` only apply to pendingOperations, well I don't think we have a `blockHeight` at this moment on the flow (before a broadcast and neither after is it's in queue).
In pending operation it's just `blockHeight` and `blockHash` are `null`

or may be i misunderstood something ?

henri-ly:
I'll do a another PR with websocket refacto

PR #982, opened 2020-12-11
Add appUninstallAll command

juan-cortes:


cli/src/commands/appUninstallAll.js
```diff
@@ -0,0 +1,40 @@
...
+        mergeMap((deviceInfo) =>
+          listApps(t, deviceInfo).pipe(
+            filter((e) => e.type === "result"),
+            map((e) => {
+              const { appByName, installed } = e.result;
+              // NB will uninstall those with dependecies first
+              const sortedInstalledApps = installed.sort(
+                (a, b) =>
+                  appByName[b.name].dependencies.length -
+                  appByName[a.name].dependencies.length
```
gre, line 28: it's good enough but i think a better would be to use app's logic called "wipe".

gre (approved):

cli/src/commands/appUninstallAll.js
juan-cortes: Fair enough, didn't know it existed

gre (approved):

PR #983, opened 2020-12-11
fix play console error

machard:
Capture d’écran 2020-12-11 à 15 23 52

### DIFF
--- a/src/hw/actions/manager.js
+++ b/src/hw/actions/manager.js
@@ -214,8 +214,14 @@ const implementations = {
           .pipe(
             timeout(DEVICE_POLLING_TIMEOUT),
             catchError((err) => {
-              const productName = getDeviceModel(pollingOnDevice.modelId)
-                .productName;
+              let productName;
+              try {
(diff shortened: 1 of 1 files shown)


src/hw/actions/manager.js
```diff
@@ -214,8 +214,14 @@ const implementations = {
...
             catchError((err) => {
-              const productName = getDeviceModel(pollingOnDevice.modelId)
-                .productName;
+              let productName;
+              try {
+                // this one throws if a wrong modelId is passed
+                const deviceModel = getDeviceModel(pollingOnDevice?.modelId);
+                productName = deviceModel.productName;
+              } catch {
+                productName = "product name unknown";
```
valpinkman, line 223: I think we'd rather have an empty string here but other than that lgtm

gre:
to me this is not to be fixed here. The problem is not here, a modelId is supposed to always be defined on user land. it's a bug in LLM to not provide one, fallbacks need to happen on LLM side.

PR #987, opened 2020-12-15
LL-3897 close loan order

machard:
https://ledgerhq.atlassian.net/browse/LL-3897

### DIFF
--- a/src/compound/logic.js
+++ b/src/compound/logic.js
@@ -157,7 +157,7 @@ export function makeCompoundSummaryForAccount(
         while (amountToClose.gt(0) && opened.length > 0) {
           const closingOperation = opened.shift();
           if (amountToClose.gte(closingOperation.amountSupplied)) {
-            closed.push({
+            closed.unshift({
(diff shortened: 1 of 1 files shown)


gre (approved):

valpinkman:
Ok so I think it's safer not to touch this function,

I'd rather changed the sort in the `makeClosedHistoryForAccounts`
https://github.com/LedgerHQ/ledger-live-common/blob/2734a985fbad8fe9de7303d45a20c0b09ab1784d/src/compound/logic.js#L264-L266

I think we just need to invert the date checking in the sort (b - a)

Need some test to be sure it works

valpinkman (changes requested):
Ok so I think it's safer not to touch this function,

I'd rather changed the sort in the `makeClosedHistoryForAccounts`
https://github.com/LedgerHQ/ledger-live-common/blob/2734a985fbad8fe9de7303d45a20c0b09ab1784d/src/compound/logic.js#L264-L266

I think we just need to invert the date checking in the sort (b - a)

Need some test to be sure it works

PR #991, opened 2020-12-18
add eip-681

machard:
So,
I am not sure if this is full support.
Do we want full support ?
Also I am not sure where to test this ?
https://ledgerhq.atlassian.net/browse/LL-4000

### DIFF
--- a/src/__tests__/currencies.js
+++ b/src/__tests__/currencies.js
@@ -434,6 +434,31 @@ test("decodeURIScheme", () => {
     address: "1gre1noAY9HiK2qxoW8FzSdjdFBcoZ5fV",
     amount: BigNumber("1234567000000"),
   });
+
+  expect(
+    decodeURIScheme(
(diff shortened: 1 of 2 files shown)


gre (changes requested):
Check again but I think it is not accepting wei value

machard:
"The amount is specified in value parameter, in the atomic unit (i.e. Wei). The use of scientific notation is strongly encouraged. For example, requesting 2.014 ETH to address 0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359 would look as follows: ethereum:0xfb6916095ca1df60bb79Ce92ce3ea74c37c5d359?value=2.014e18"
Indeed it seems you're right, no need to do the multiplication

gre:
Can you add tests to cover the different syntaxes? Thanks

gre (dismissed):
please add tests 👍

gre (approved):

PR #993, opened 2020-12-22
Add support for floating rates on Swap.

juan-cortes:
This is a breaking change that should **not be merged** until the current v2 version of the swap backend is in production.
Backend will maintain a legacy version for users running live-common prior to this commit, and a v2 which will start from here on. The changes include dropping unneeded wrapping arrays for some endpoints and more importantly the support for floating rates from our providers.

The CLI command has also been updated accordingly to allow testing the feature.


gre:
@juan-cortes you'll probably want to merge live-common master in as well as on LLD/LLM, because desktop & mobile have latest live-common version now.

juan-cortes:
OK, I'll take note, cheers

gre (approved):

PR #995, opened 2020-12-28
LL-4176 LL-4159 update tezos bakers

machard:
https://ledgerhq.atlassian.net/browse/LL-4159
https://ledgerhq.atlassian.net/browse/LL-4176


gre:
Capture d’écran 2021-01-04 à 14 07 09

it seems like we must remove these for CI to pass. this usually means these validators are no longer active.

PR #997, opened 2021-01-05
add isConfirmed logic in live-common

henri-ly:
We weren't able to know which logic is the right one for confirmed an operation status

The logic was defined in several places :

- [in ConfirmationCell](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/components/OperationsList/ConfirmationCell.js#L61)

- [in OperationDetails](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/modals/OperationDetails/index.js#L148)

- [in mobile OperationDetails](https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/screens/OperationDetails/Content.js#L114)

So the main idea of this PR is to centralised the logic and apply it on mobile and desktop to have the same condition everywhere.

### DIFF
--- a/src/operation.js
+++ b/src/operation.js
@@ -122,3 +122,12 @@ export const getOperationConfirmationDisplayableNumber = (
   account.blockHeight && operation.blockHeight && account.currency.blockAvgTime
     ? String(account.blockHeight - operation.blockHeight + 1)
     : "";
+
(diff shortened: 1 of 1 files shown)


gre (approved):
thanks.

PR #999, opened 2021-01-05
fix(network): handle error response as parsed json

rjarasson-ledger:
Currently, `network.js` is used as a generic http client by many coins, with the advantage of handling errors and logging, and casting to error.

But it does it strangely, not taking into account that in most cases, body is already parsed as json.
I added the handling of error responses as Object, and use an helper to get the message.
I wildly added the "cause" property as a possible error message (that is used in polkadot's sidecar), I don't think it's the best idea though.

I really think that generic client should be more generic and provide a clean way to handle error / transform response etc... without forcing to use plain axios, it's a good idea to wrap it with common behaviour.

### DIFF
--- a/src/network.js
+++ b/src/network.js
@@ -17,11 +17,18 @@ const makeError = (msg, status, url, method) => {
     : new LedgerAPI5xx(msg, obj);
 };
 
+const getErrorMessage = (data: Object): ?string => {
+  return (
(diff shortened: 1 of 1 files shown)


gre (approved):
I agree with you regarding the hack-y-ness of this approach.

PR #1002, opened 2021-01-06
Prevent loadCountervalues to be executed 3 times initially

JunichiSugiura:
Current impl triggers initial polling 3 times.

1. when polling loop initiated
1. when pairs changed
1. when pairs order changed

This fix will prevent 2 and 3 above.


gre (approved):

src/countervalues/react.js
```diff
@@ -119,11 +120,18 @@ export function Countervalues({
     // eslint-disable-next-line react-hooks/exhaustive-deps
   }, [savedState]);
 
-  // trigger poll only when trackingParis is not exactly the same
-  const pairs = useMemo(() => userSettings.trackingPairs.map(pairId).join(), [
-    userSettings.trackingPairs,
-  ]);
+  // trigger poll only when trackingPairs is not exactly the same
```
juan-cortes, line 123: I liked _Paris_ better

JunichiSugiura: 🤫

gre:
https://github.com/LedgerHQ/ledger-live-common/pull/1004 is a proposal that address this problem more generally: basically debouncing the userSettings + making sure it is only changing reference when necessary.

PR #1003, opened 2021-01-06
fix(bridge): no debounce first transaction update + avoid double call

rjarasson-ledger:
This changes the debounce strategy used in useBridgeTransaction.

**Note: This change WILL impact the behaviour of all flow using this hook, so we must test it widely.**

## Context

useBridgeTransaction is a hook handling all interaction with a transaction in an AccountBrige.
It is responsible for preparing transaction (prepareTransaction in bridge can update transaction in order for instance, to add properties like fees), and getting transaction status (including warnings, errors, amount value and fees for a transaction)

## Current behaviour
When a transaction is updated, the resulting status (not the request) is debounced, with a minimum delay of 300ms before updating the UI.
Any transaction occuring during this delay will discard the previous results from previous updates, and trigger a new status retrieval (prepareTransaction + getTransactionStatus).
It has a some flaws:
* There is always a delay of 300ms that acts like a rendering throttle and visual latency.
* The request is not debounced, so if prepareTransaction or getTransactionStatus do HTTP requests, or heavy calculation, they will be executed anyway
* If prepareTransaction modifies the transaction, the effect is called twice:
  * First effect - will update status and update transaction with changed fields (dispatch)
  * Second effect - due to dispatch of a different transaction while in fact, transaction and statusOnTransaction are now the same
  
 
### No double call
If prepareTransaction changes the transaction, this update will be ignored in the next effect execution, by using the `bridgePending` boolean.
If the transaction and the statusOnTransaction are the same, the state is synced, and UI is not loading anymore. Thus we don't need to call getTransactionStatus again.

### First update in a row is not debounced
If user trigger an transaction update (e.g. filling an input with a single char), the effect is called Immediately.

If this effect is almost immediate (no HTTP request or heavy computation), the resulting status is dispatched immediately.

**This will also re-render UI immediately. If next updates occurs soon, re-renders can occur too often and put heavy load on CPU.**

### Next updates in a row are debounced
If first effect (in a row) takes more time than the second update happens, any update occuring while transaction status is being calculated will be debounced.
First effect results will be discarded as soon as possible (generally when result is resolved), and *only the last update in a row* will trigger computation and then will be handled and dispatched.

This is the current behaviour (minus the double call that has been removed).

So if user enters an amount, like "0.11111111111", and status takes some time to be obtained:
* first update effect (value = "0") will immediately execute (for instance, fee calculation)
* second and next updates (values "0." ... "0.1111111111") will be debounced of 300ms and ignored
* last update (value = ""0.11111111111") will be debounced of 300ms and then be handled and dispatched
* status will be rendered in UI only once after user finished typing + 300ms + XXXms required to calculate status

## Possible issue
Since first update in a row is not debounced, if status is ready faster than the updates rate, multiple renders will occur since it would be immediately available and dispatched. It's more reactive but could also be heavier.

If this issue is problematic, *an additionnal throttle feature must be added* by debouncing also updates that happens too soon after the last onStatus dispatch.

Here is my idea if we want to also add throttle:
```
const statusIsPending = useRef(false);
const lastStatusUpdate = useRef(0);

// [...]

const shouldDebounce = statusIsPending.current || lastStatusUpdate.current - Date.now() < DEBOUNCE_STATUS_DELAY
statusIsPending.current = true

// When status is resolved or rejected
statusIsPending.current = false
lastStatusUpdate = Date.now()
```

### DIFF
--- a/src/bridge/useBridgeTransaction.js
+++ b/src/bridge/useBridgeTransaction.js
@@ -143,7 +143,7 @@ const reducer = (s: State, a): State => {
 
 const INITIAL_ERROR_RETRY_DELAY = 1000;
 const ERROR_RETRY_DELAY_MULTIPLIER = 1.5;
-const DEBOUNCED_STATUS = 300;
+const DEBOUNCE_STATUS_DELAY = 300;
 
 const useBridgeTransaction = (optionalInit?: ?() => $Shape<State>): Result => {
   const [
@@ -179,16 +179,32 @@ const useBridgeTransaction = (optionalInit?: ?() => $Shape<State>): Result => {
   const mainAccount = account ? getMainAccount(account, parentAccount) : null;
 
   const errorDelay = useRef(INITIAL_ERROR_RETRY_DELAY);
+  const statusIsPending = useRef(false); // Stores if status already being processed
+
+  const bridgePending = transaction !== statusOnTransaction;
 
   // when transaction changes, prepare the transaction
   useEffect(() => {
     let ignore = false;
     let errorTimeout;
+
+    // If bridge is not pending, transaction change is due to
+    // the last onStatus dispatch (prepareTransaction changed original transaction) and must be ignored
+    if (!bridgePending) return;
+
     if (mainAccount && transaction) {
-      Promise.resolve()
+      // We don't debounce first status refresh, but any subsequent to avoid multiple calls
+      // First call is immediate
+      const debounce = statusIsPending.current
+        ? delay(DEBOUNCE_STATUS_DELAY)
+        : null;
+
+      statusIsPending.current = true; // consider pending until status is resolved (error or success)
+
+      Promise.resolve(debounce)
         .then(() => getAccountBridge(mainAccount, null))
         .then(async (bridge) => {
-          const start = Date.now();
+          if (ignore) return;
           const preparedTransaction = await bridge.prepareTransaction(
             mainAccount,
             transaction
@@ -199,10 +215,6 @@ const useBridgeTransaction = (optionalInit?: ?() => $Shape<State>): Result => {
             preparedTransaction
           );
           if (ignore) return;
-          const delta = Date.now() - start;
-          if (delta < DEBOUNCED_STATUS) {
-            await delay(DEBOUNCED_STATUS - delta);
-          }
 
           return {
             preparedTransaction,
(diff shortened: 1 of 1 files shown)


src/bridge/useBridgeTransaction.js
```diff
@@ -179,16 +179,32 @@ const useBridgeTransaction = (optionalInit?: ?() => $Shape<State>): Result => {
...
-      Promise.resolve()
+      // We don't debounce first status refresh, but any subsequent to avoid multiple calls
+      // First call is immediate
+      const debounce = statusIsPending.current
+        ? delay(DEBOUNCE_STATUS_DELAY)
+        : null;
+
+      statusIsPending.current = true; // consider pending until status is resolved (error or success)
+
+      Promise.resolve(debounce)
```
gre, line 204: I'm not sure that `Promise.resolve(promise)` actually awaits the promise.

gre (approved):
I'm ok with the approach but I think the delay is actually not taken into account.

src/bridge/useBridgeTransaction.js
rjarasson-ledger: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Promise/resolve

```The Promise.resolve() method returns a Promise object that is resolved with a given value. If the value is a promise, that promise is returned; if the value is a thenable (i.e. has a "then" method), the returned promise will "follow" that thenable, adopting its eventual state; otherwise the returned promise will be fulfilled with the value.```

It permits passing a promise or a raw value directly (here null) and cast it as a Promise in the next steps.
For what i have tested, this is working as said on Polkadot.

But i did not tested this with immediate status on other coins.
On polkadot, i have almost immediate value once fee estimation is in cache, and i tried triggering multiple updates as fast as possible but clicking fast on a checkbox:
* if fees are not cached, it estimate fees only twice (first click and last)
* if fees are cache, the status is updated quickly, but if i'm fast enough (quicker than getting a response) the status starts to debounce and only renders at the end.

File: src/bridge/useBridgeTransaction.js [hunk 2]
```diff
@@ -179,16 +179,32 @@ const useBridgeTransaction = (optionalInit?: ?() => $Shape<State>): Result => {
...
+  const bridgePending = transaction !== statusOnTransaction;
 
   // when transaction changes, prepare the transaction
   useEffect(() => {
     let ignore = false;
     let errorTimeout;
+
+    // If bridge is not pending, transaction change is due to
+    // the last onStatus dispatch (prepareTransaction changed original transaction) and must be ignored
+    if (!bridgePending) return;
```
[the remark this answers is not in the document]
gre: i was about to say this too (regarding mainAccount) but actually in the reducer, we will always re-set a new Transaction when a new Account changes. so it should be good.

gre:
I'm setting a "HODL" on this PR because I prefer we merge it beginning of next week, after the end of this sprint to have the sprint-time to do full QA on all impacts this could have.

gre (approved):

PR #1004, opened 2021-01-06
useTrackingPairForAccounts for better userSettings ref change

gre:
- This adds `useTrackingPairForAccounts` that makes sure the reference stay the same as long as the tracking pair are the same.
- When used in LLM/LLD this will ensure that the userSettings really change ref when necessary. That way, it simplifies the actual logic of countervalues.
- We also debounce userSettings itself in order to not be spammed of initial reloads. In practice, this should not occur very often, tracking pair will rarely change twice in a row. The possibility that this happens however is if you come back a few days later on our ethereum/tron accounts and let's say you receive more tokens => each sync will add more tracking pairs.

---

Let's test a bit this works fine and then plan to merge this.

It will requires a change in desktop and mobile to correctly work.

---

To use with this diff on desktop:

```diff
--- a/src/renderer/actions/general.js
+++ b/src/renderer/actions/general.js
@@ -14,8 +14,8 @@ import type { FlattenAccountsOptions } from "@ledgerhq/live-common/lib/account";
 import {
   useDistribution as useDistributionCommon,
   useCalculateCountervalueCallback as useCalculateCountervalueCallbackCommon,
+  useTrackingPairForAccounts,
 } from "@ledgerhq/live-common/lib/countervalues/react";
-import { inferTrackingPairForAccounts } from "@ledgerhq/live-common/lib/countervalues/logic";
 import { pairId } from "@ledgerhq/live-common/lib/countervalues/helpers";
 import type { State } from "~/renderer/reducers";
 import { accountsSelector, activeAccountsSelector } from "~/renderer/reducers/accounts";
@@ -152,8 +152,5 @@ export function useTrackingPairIds(): string[] {
 export function useTrackingPairs() {
   const accounts = useSelector(accountsSelector);
   const countervalue = useSelector(counterValueCurrencySelector);
-  return useMemo(() => inferTrackingPairForAccounts(accounts, countervalue), [
-    accounts,
-    countervalue,
-  ]);
+  return useTrackingPairForAccounts(accounts, countervalue);
 }
```

### DIFF
--- a/src/countervalues/react.js
+++ b/src/countervalues/react.js
@@ -2,6 +2,7 @@
 import { BigNumber } from "bignumber.js";
 import React, {
   createContext,
+  useRef,
   useMemo,
   useContext,
   useEffect,
@@ -35,13 +36,15 @@ import {
   loadCountervalues,
   exportCountervalues,
   importCountervalues,
+  inferTrackingPairForAccounts,
 } from "./logic";
 import type {
   CounterValuesState,
   CounterValuesStateRaw,
   CountervaluesSettings,
+  TrackingPair,
 } from "./types";
-import { pairId } from "./helpers";
+import { useDebounce } from "../hooks/useDebounce";
 
 // Polling is the control object you get from the high level <PollingConsumer>{ polling => ...
 export type Polling = {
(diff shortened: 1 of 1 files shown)


JunichiSugiura (approved):
LGTM

`triggerLoad` and `isPolling` could be included inside a reducer but another time!

PR #1006, opened 2021-01-08
Fix/walletconnect tests

machard:
it fixes a walletconnect test randomly failing.


gre:
maybe you need to regen snapshot?

PR #1010, opened 2021-01-12
LL-4373 Discard existing ops on sync if blacklist has changed

juan-cortes:
Since the js bridge rewrite blacklisted tokens could no longer be shown without a clear cache, since we wo uld only fetch from a given block height. This addresses that by setting a cache id at the account level and dismissing the existing ops if it doesnt match the blacklist settings.

https://ledgerhq.atlassian.net/browse/LL-4373


gre:
LGTM to me. 2 things to polish as we discussed:
- no need to do this for cross.js
- add a test, similar ideas the the file compound.resilience.test.js

juan-cortes:
Should be fine now, I added test coverage for the blacklistedTokenids functionality

gre:
we'll merge #963  first and then we'll regen the snapshots here.

gre:
@juan-cortes we can actually move on at putting the latest snapshots of the CI. the other PR is not ready yet.

PR #1012, opened 2021-01-18
walletconnect - include decoded data

machard:


src/families/ethereum/types.js
```diff
@@ -79,6 +79,10 @@ export type TypedMessage = {
   primaryType: string,
   domain: any,
   message: any,
+  hahes: {
```
gre, line 82: Typo

src/families/ethereum/types.js
```diff
@@ -87,6 +91,9 @@ export type TypedMessageData = {
   verify?: boolean,
   derivationMode: DerivationMode,
   message: TypedMessage,
+  hahes: {
```
gre, line 94: Typo

src/walletconnect/index.js
```diff
@@ -2,20 +2,21 @@
...
 } from "../families/ethereum/types";
-import {
-  stringHash,
-  domainHash,
-  messageHash,
-} from "../families/ethereum/hw-signMessage";
+import { domainHash, messageHash } from "../families/ethereum/hw-signMessage";
 import type { MessageData } from "../hw/signMessage/types";
 
+const decoder = new InputDataDecoder(`${__dirname}/erc20.abi.json`);
```
gre, line 18: I'm pretty sure will not work in React native.
I'm afraid we need to make the erc20 abi a .js file that we import

gre (changes requested):
Mmh I'm still not sure to understand how it works.

To me the goal is that, if we detect it's an approve, we make it a Transaction object of .mode==erc20.approve

So we need some kind of mapper. I don't see this in current code.

PR #1029, opened 2021-01-26
LL-4428 Fix ghost pending tx when dropped

juan-cortes:
We were only clearing the pending operations when an `OUT` operation was added with a nonce higher or equal to the pending one, this resulted in the edge case where we could replace a tx with another one sending to a different account but it would remain there for a self tx. The reason is this self transaction is currently marked as an `IN` not a `SELF` or similar (could be an improvement cc product according to @gre).

### DIFF
--- a/src/families/ethereum/postSyncPatch.js
+++ b/src/families/ethereum/postSyncPatch.js
@@ -33,11 +33,10 @@ const postSyncPatchGen = <T: AccountLike>(
 };
 
 const postSyncPatch = (initial: Account, synced: Account): Account => {
-  const sendingOps = (synced.operations || []).filter(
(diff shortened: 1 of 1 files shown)


gre:
it would be great but a bit challenging to cover the "replace by fee" in future for 🤖 . but it's good for now.

gre (approved):

juan-cortes:
> it would be great but a bit challenging to cover the "replace by fee" in future for 🤖 . but it's good for now.

couldn't we do exactly the same scenario? send a tx with low gas and another one with higher gas, wait for sync, expect the first one to not exist and the second to exist 🤔

PR #1033, opened 2021-01-26
Polkadot: update api lib to 3.6.4

henri-ly:
Previously revert because of the bug of polkadot api lib

It will be needed to merged this commit after i fixed the signing problem

https://github.com/LedgerHQ/ledger-live-desktop/pull/3515


rjarasson-ledger:
I tested successfully with https://github.com/LedgerHQ/ledger-live-desktop/pull/3515

rjarasson-ledger (approved):

PR #1038, opened 2021-01-28
Ethereum tokens refresh

gre:
When updating Ledger Live, if new tokens were added, Ethereum accounts were not correctly loading the new token accounts (user need to Clear Cache). This fixes this.


juan-cortes (approved):
lgtm

PR #1040, opened 2021-01-29
feat(Polkadot): improve fees cache by using tx bytesize - COIN-1249

rjarasson-ledger:
Fees estimation depends on 2 factors:
* weight: which depends on the method called by the extrinsic
* bytesize (or length): which varies when changing amounts.

Different amounts don't always change bytesize so we can use bytesize instead of exact amount - avoiding many useless calls.

NOTE: This should be heavily tested since any bad implementation can lead to wrong fee estimation, but I have no specific test plan for this.

### DIFF
--- a/src/families/polkadot/cache.js
+++ b/src/families/polkadot/cache.js
@@ -1,5 +1,4 @@
 // @flow
-import { BigNumber } from "bignumber.js";
 import { TypeRegistry, ModulesWithCalls } from "@polkadot/types";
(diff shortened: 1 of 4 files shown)


gre (approved):

haammar-ledger (approved):

gre:
test fail because ALGO & BCH. probably infra situation, no problem on dot. => go

PR #1041, opened 2021-01-29
Use correct unit for swap rate errors

juan-cortes:
Fixes a regression in the error handling of swap exchange rate bounds where we would treat 1BTC (or base unit for chose currency) as 1 satoshi of the same currency, resulting in error that displayed the wrong amount. 
image

>_Prettier brought some unrelated styling changes here, perhaps we need to fix that on master first_


gre:
CI is go, just algorand is :x: because ongoing infra issue, all other tests pass.

PR #1042, opened 2021-01-30
Test account sync resilience when no network

gre:
add a test to ensure a sync fails when no network. For instance, a networking error should not yield empty accounts but instead should fail the sync (it was a regression)


gre:
This works locally but not on CI.. on CI it seems to just times out..

juan-cortes:
I'm taking a look locally and it also passes, I'll update if I find out more about what's happening

gre:
Ok, i just recalled what the issue is:

two **Jest** tests that uses **libcore** makes one of the test **hanging forever**. I never managed to understand why, but it sounds related to junky behavior of Jest when forking test engines (diff processes) and/or probably the behavior of libcore with concurrency.

gre:
I don't know an easy fix. will create a ticket later.

src/countervalues/logic.test.js
```diff
@@ -28,7 +40,39 @@ describe("API sanity", () => {
...
+      calculate(state, { from: bitcoin, to: usd, value: 10000000 })
+    ).toBeGreaterThan(1000);
   });
+  test("resilience when no network with a cache", async () => {
+    const userSettings = {
+      trackingPairs: [
+        {
+          from: bitcoin,
+          to: usd,
+          startDate: new Date(now - 200 * 24 * 60 * 60 * 1000),
```
juan-cortes, line 53: Is this arbitrary?

juan-cortes (approved):

src/countervalues/logic.test.js
gre: yeah. like many tests. it's going to ask 200 days of data like if account was 200 days old.

gre:
lol the test is showing to me that if you do a sync with libcore and you did it in the past, it actually succeed 🤯

juan-cortes:
> lol the test is showing to me that if you do a sync with libcore and you did it in the past, it actually succeed 🤯
> because the test taken individually, proving that no network=sync fails, actually succeed. but fails if there were sync before.

image

valpinkman (approved):

PR #1048, opened 2021-02-01
LL-3936 (AddAcounts): add preferred new account scheme to the add account flow

LFBarreto:
(AddAcounts): add preferred new account scheme to the add account flow + unlock legacy derivation mode for newly created accounts﻿


src/derivation.js
```diff
@@ -418,21 +418,42 @@ export const getDerivationModesForCurrency = (
...
+const preferredList = ["native_segwit", "segwit", ""];
+// null => no settings
+// [ .. ]
+export const getPreferredNewAccountScheme = (
+  currency: CryptoCurrency
+): ?(DerivationMode[]) => {
+  if (currency.family !== "bitcoin") return null;
+  const list = getDerivationModesForCurrency(currency)
+    .filter((s) => preferredList.includes(s))
+    .sort((a, b) => preferredList.indexOf(a) - preferredList.indexOf(b));
```
gre, line 445: preferredlist.filter(..)

Would be a bit more concise

src/derivation.js
```diff
@@ -418,21 +418,42 @@ export const getDerivationModesForCurrency = (
...
-  }
   if (currency.supportsNativeSegwit) {
     all.push("native_segwit");
   }
   if (currency.supportsSegwit) {
     all.push("segwit");
   }
+  if (!disableBIP44[currency.id]) {
+    all.push("");
+  }
```
gre, line 429: I don't think we should reorder this. Or maybe we need to reorder many of them, but right now the idea was to scan the most legacy first and least legacy at the end.

gre: Maybe it's ok to reverse it all because indeed native segwit and segwit ordering was already not really consistant.

gre: In short my question is: is this change necessary / why was it needed? Let's discuss

LFBarreto: without this when we first start the scan of the accounts we get an error on the front end side while we retrieve the other types of accounts + the newly created accounts ordering is messed up at the end and we show the legacy first which is not what we want :shrug: i remember we already switched the native and segwit order once i though this might be ok

juan-cortes (approved):

src/derivation.js
gre: ok. i'm ok with it.
**on a more long term discussion, i genuinely wonder if we should put all the "advanced/custom" searches actually AFTER these, so we truely have a reversed order** (from most preferable to less likely).

I actually like the other way (most likely to less likely), because it will make the UX perceived as faster anyway... let's keep this, but we need to keep the question for the future.

gre:
@LFBarreto are we are going to keep this swap, please make the CI pass:

- first of all, the Lint is :x: right now
- secondly, the CI will likely fail because we changed the derivation ordering... this is actually not that trivial to fix because of the dataset hardcodes APDUs, we can change the ordering of all the scan accounts APDUs to fix it and regen snapshot.. alternatively we can ask arnaud to regen them for us, but well, it's very time consuming too. let's discuss

src/derivation.js
```diff
@@ -418,21 +418,43 @@ export const getDerivationModesForCurrency = (
...
```
gre, line 429: NB: due to this change of order that we agreed on, all the CI tests are going to break and will need to be fixed as well.

gre (changes requested):
Please also add some unit tests for some currencies that would test the two new functions added.

gre:
@LFBarreto you can merge master and do a checkout --theirs on the .snap file because in any case we would need to regen from the CI snapshot.

gre (changes requested):
current CI find bugs in the recent changes of ordering..

PR #1049, opened 2021-02-01
re-enable dataset tests on Algorand

gre:


juan-cortes (approved):

gre:
this appears ok now. but we'll regen snapshots on master.

PR #1053, opened 2021-02-04
DASH to v3 with V3->V2 rollback system

gre:
This implements a remote config mechanism to be able to do incremental switch on our explorers, remotely.

In this PR, we also will include:
- LL-4552 rollback system in place
- LL-5359 #1146  dash to go v3
- LL-5360 #1149 Fixes libcore to support config changes without clear cache to be necessary

---

notes regarding plan.

With this PR. DASH v3 is enabled but with the rollback system and the current configuration (https://cdn.live.ledger.com/config/explorerConfig.v1.json) only 20% of us will be on V3. Rest will fallback on V2.

To test this, you can simply change your USER_ID to have chance to be one of the 20%. We can assert it's indeed the case. We can also assert that the Live is able to correctly move to V3 or back to V2 whenever you change back between 2 USER_IDs. it does not require a clear cache (fixed in this PR)


gre:
Todo: add 5mn cache

valpinkman (approved):

PR #1055, opened 2021-02-09
LL-4454 (csvExport): ignore FEES op type duplicate amount

LFBarreto:
(csvExport): ignore FEES op type duplicate amount﻿


gre (changes requested):
i have a comment on LL-4454.

PR #1057, opened 2021-02-09
LL-4545 start announcements logic

valpinkman:
cc: @IAmMorrow


src/announcements/react.js
```diff
@@ -0,0 +1,249 @@
...
+      return {
+        ...state,
+        seenIds: uniq([...state.seenIds, ...action.seenIds]),
+      };
+
+    case "loadCache": {
+      const cache = action.announcements.reduce((acc, curr) => {
+        acc[curr.uuid] = curr;
+        return acc;
+      }, {});
```
gre, line 102: reduce to build a `{}` is imo an anti-pattern (fact we mutate the accumulator), we could just forEach and with a `const cache = {};` that we fill from inside forEach.

src/announcements/react.js
```diff
@@ -0,0 +1,249 @@
...
+      const seenIds = intersection(allIds, state.seenIds);
+      return {
+        ...state,
+        cache,
+        seenIds,
+        allIds,
+        isLoading: false,
+        error: null,
+      };
+    }
```
gre, line 130: there is almost twice the same code.

src/announcements/react.js
```diff
@@ -0,0 +1,249 @@
...
+  createContext,
+  useContext,
+  useEffect,
+  useMemo,
+  useReducer,
+} from "react";
+import type { Announcement } from "./types";
+import { fetchAnnouncements } from "./logic";
+
+type Context = {
```
gre, line 14: I think we really want to find another name. "Context" collides with react's Context naming, and here it's not even the "context" provided by your AnnouncementsContext.

I usually call this "user settings". like in CountervaluesUserSettings.

src/announcements/react.js
```diff
@@ -0,0 +1,249 @@
...
+
+    if (
+      announcement.currencies &&
+      !announcement.currencies.some((c) => currencies.includes(c))
+    ) {
+      return false;
+    }
+
+    const publishedAt = new Date(announcement.published_at);
+    if (publishedAt.getTime() > date.getTime()) {
```
gre, line 171: you don't need to `getTime()` to compare dates

src/announcements/react.js
```diff
@@ -0,0 +1,249 @@
...
+      return false;
+    }
+
+    if (
+      announcement.expired_at &&
+      new Date(announcement.expired_at).getTime() < date.getTime()
```
gre, line 177: same

src/announcements/react.js
```diff
@@ -0,0 +1,249 @@
...
+  };
+
+  return (
+    <AnnoucementsContext.Provider value={value}>
+      {children}
+    </AnnoucementsContext.Provider>
+  );
+};
+
+export const useAnnouncements = (): Context => useContext(AnnoucementsContext);
```
LFBarreto, line 249: I think there's an issue on the type returned here not the correct one

LFBarreto (changes requested):

package.json
```diff
@@ -84,6 +84,7 @@
     "@ledgerhq/logs": "5.43.0",
     "@polkadot/types": "3.6.4",
     "@walletconnect/client": "1.3.6",
+    "@xstate/react": "^1.3.1",
```
gre, line 87: will this be a problem if we depends on live-common from CLI project?

src/announcements/react.js
```diff
@@ -0,0 +1,162 @@
...
+  localizeAnnouncements,
+  filterAnnouncements,
+} from "./logic";
+import fetchApi from "./api";
+import { useMachine } from "@xstate/react";
+import { announcementMachine } from "./machine";
+
+type Props = {
+  children: React$Node,
+  onLoad: () => Promise<{
```
gre, line 23: I don't like the wording onLoad/onSave because it makes think it's just an event (like onClick) but actually it's the actual load/save

src/announcements/react.js
```diff
@@ -0,0 +1,162 @@
...
+    announcements: Announcement[],
+    seenIds: string[],
+    lastUpdateTime: number,
+  }>,
+  onSave: ({
+    announcements: Announcement[],
+    seenIds: string[],
+    lastUpdateTime: number,
+  }) => Promise<void>,
+  context: AnnoucementsUserSettings,
```
gre, line 33: Is it possible to also rename the prop?

src/announcements/machine.js
```diff
@@ -0,0 +1,106 @@
...
+    },
+  },
+  {
+    actions: {
+      setAsSeen: assign((context, event) => ({
+        seenIds: uniq([...context.seenIds, event.seenId]),
+      })),
+    },
+  }
+);
```
gre, line 106: Maybe it's just me but the complexity brought by the xstate approach is a bit overwhelming and I'm not confident reviewing that part myself.
I'm not sure who's in the team have enough knowledge of this.

gre (commented):
Not super convinced by the usage of xstate here. We managed to do more complex things that this without xstate and it's sometimes easier to read.
Personally I can't understand what's the machine is about and why it's needed.

I think if we can afford to drop it we should

gre:
ok let's keep xstate & schedule internal prez about this.

IAmMorrow:
We are seeing the end of the tunnel on this.

gre:
I suggest to rename `providers` folder to `announcements`

PR #1058, opened 2021-02-10
Few updates on countervalues imports & doc

meriadec:
main issue I'm facing right now in web-integration is caused by import pulling the whole npm registry, and ending up with incompatible modules e.g:

This PR does not bring much change than just importing directly from helper file rather than index file.

+ minor doc adjustment

### DIFF
--- a/docs/countervalues.md
+++ b/docs/countervalues.md
@@ -32,14 +32,14 @@ import {
 
 function App() {
   // get these in the client side
-  const account = ...
-  const trackingPairs = ...
+  const accounts = ...
 
   const trackingPairs = useTrackingPairForAccounts(accounts, countervalueCurrency)
+
   return (
     <Countervalues
       userSettings={{
-        autoFillGaps: true,
+        autofillGaps: true,
         trackingPairs
       }}
       // use this prop to restore local countervalues cache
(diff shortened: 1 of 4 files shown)


src/countervalues/logic.js
```diff
@@ -2,8 +2,7 @@
 
 import { log } from "@ledgerhq/logs";
 import type { Currency, Account } from "../types";
-import { flattenAccounts } from "../account";
-import { getAccountCurrency } from "../account";
+import { flattenAccounts, getAccountCurrency } from "../account/helpers";
```
JunichiSugiura, line 5: 👍

docs/countervalues.md
```diff
@@ -32,14 +32,14 @@ import {
...
-  const trackingPairs = ...
+  const accounts = ...
 
   const trackingPairs = useTrackingPairForAccounts(accounts, countervalueCurrency)
+
   return (
     <Countervalues
       userSettings={{
-        autoFillGaps: true,
+        autofillGaps: true,
```
JunichiSugiura, line 42: 🙏

JunichiSugiura (approved):
We should def utilize sub modules and decouple each one.

gre (approved):

PR #1062, opened 2021-02-15
attempt to enable BCH, XSN, VIA

gre:


gre:
BCH and VIA is still no go
did a pr for #1063

PR #1064, opened 2021-02-15
Ll 4390 - fix standalone uppercase bc1 addresses

machard:


src/families/bitcoin/bridge/libcore.js
```diff
@@ -71,7 +71,13 @@ const createTransaction = () => ({
...
 });
 
-const updateTransaction = (t, patch) => ({ ...t, ...patch });
+const updateTransaction = (t, patch) => {
+  const updatedT = { ...t, ...patch };
+  if (updatedT.recipient.toLowerCase().indexOf("bc1") === 0) {
+    updatedT.recipient = updatedT.recipient.toLowerCase();
+  }
+  return updatedT;
+};
```
gre, line 80: to test this effectively work, can you modify one of the test-dataset.js file to change one of the test using an upper case address when it's used to validate a tx? thanks

gre (changes requested):

PR #1069, opened 2021-02-19
Introduce Account#used boolean to track account "exists"

gre:
- requires https://github.com/LedgerHQ/lib-ledger-core/pull/745
- replace most usage of isAccountEmpty by Account#used which track the information regarding the "creation" of the account in the past (where the account ever used by the user) which allows for Cosmos to support detection of empty account that used to have an history to not break the Scan Accounts discovery
- NB: before we merge this live-common PR itself, we'll make sure that we're good on test with dataset + bot.
- there should be no problem regarding the backward compatibility, account row gets migrated to introduce the field.


gre:
this is waiting a new libcore version with a fix from @gagbo

gagbo (approved):

PR #1074, opened 2021-02-23
fix(Polkadot): handle errors from nano app

rjarasson-ledger:
Some errors codes from Polkadot Nano App were not handled and considered as successful sign.

It should now throw a TransportError with message provided by the app.

Tested with Polkadot app 5.26.1 which is no longer compatible with Polkadot blockchain, and should return a `TransportError('Spec version not supported')` error when signing.


haammar-ledger (approved):
:ok_hand:

PR #1076, opened 2021-02-23
Polkadot: adapt registry chain properties to array format

rjarasson-ledger:


rjarasson-ledger:
Closed, it does not change anything.

I just discovered that Number(["10"]) returns 10...

PR #1081, opened 2021-02-24
fix(Polkadot): ensure Polkadot prefix is also valid in address - COIN-1379

rjarasson-ledger:
Address validation was only taking SS58 and checksum is valid, but not that the address belongs to Polkadot network.

This could lead to users sending DOTs to the wrong address.

NOTE: the nano app converts the address to a polkadot address corresponding to the same public key. Although, this could lead to misunderstanding to users, since the address displayed on device is different.

This fixes the issue by considering only POlkadot addresses as valid.


haammar-ledger (approved):
:ok_hand: good catch

juan-cortes (approved):

PR #1085, opened 2021-03-02
Upgrade dependencies

gre:


gre:
something weird with stellar but nothing in this libs upgrade impacts this so let's merge and run again from master

PR #1087, opened 2021-03-04
Update XRPL endpoint to alternative

WietseWind:
the .WS domain TLD has trouble and we're (XRP Ledger Foundation) are not comfortable using .ws in the future as our primary endpoint. We launched an alternative endpoint on a more reliable TLD:
wss://node.xrpledger.foundation

More info:
https://twitter.com/XRPLF/status/1367433363657809924
https://twitter.com/WietseWind/status/1367431235686006785

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -70,7 +70,7 @@ const envDefinitions = {
   },
   API_RIPPLE_WS: {
     parser: stringParser,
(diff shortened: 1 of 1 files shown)


gre (approved):

gre:
thanks

PR #1096, opened 2021-03-11
Coin 1273 stellar js migration

ksiva-ledger:


src/families/stellar/api/horizon.js
```diff
@@ -0,0 +1,266 @@
...
+    skipMemoRequiredCheck: true,
+  });
+  return res.hash;
+};
+
+export const buildPaymentOperation = (
+  destination: string,
+  amount: BigNumber
+) => {
+  const formattedAmount = formatCurrencyUnit(currency.units[0], amount, {
```
gre, line 226: I think it's a bit dangerous to use `formatCurrencyUnit` which is initially meant to user's display into actual logic.

Typically if we forget about disableRounding.. also the formatter may (or may not) have different behavior in future, depending on local etc..

To me we need a version of this function for logic purpose that is stable.

essentially it's `amount.div(BigNumber(10).pow(currency.units[0].magnitude)).toString(10)`

Can you do a local function for now?

src/families/stellar/api/horizon.js
```diff
@@ -0,0 +1,266 @@
...
+import type { RawAccount, RawTransaction } from "./horizon.types";
+import { getAccountSpendableBalance, rawOperationToOperation } from "../logic";
+
+const LIMIT = 200;
+const FALLBACK_BASE_FEE = 100;
+
+const currency = getCryptoCurrencyById("stellar");
+
+const baseAPIUrl = getEnv("API_STELLAR_HORIZON");
+const server = new StellarSdk.Server(baseAPIUrl);
```
gre, line 20: it's a low but I think we could make this:

- lazy
- handling the possible changes of API_STELLAR_HORIZON env over time (e.g. experimental settings)

src/families/stellar/api/horizon.js
```diff
@@ -0,0 +1,266 @@
...
+import {
+  formatCurrencyUnit,
+  getCryptoCurrencyById,
+  parseCurrencyUnit,
+} from "../../../currencies";
+import type { Account, NetworkInfo, Operation } from "../../../types";
+import type { RawAccount, RawTransaction } from "./horizon.types";
+import { getAccountSpendableBalance, rawOperationToOperation } from "../logic";
+
+const LIMIT = 200;
```
gre, line 14: env for this please.

src/families/stellar/js-getTransactionStatus.js
```diff
@@ -0,0 +1,122 @@
...
+  let amount = !useAllAmount
+    ? t.amount
+    : a.balance.minus(baseReserve).minus(estimatedFees);
+  let totalSpent = !useAllAmount
+    ? amount.plus(estimatedFees)
+    : a.balance.minus(baseReserve);
+
+  if (totalSpent.gt(a.balance.minus(baseReserve))) {
+    errors.amount = new NotEnoughSpendableBalance(null, {
+      minimumAmount: formatCurrencyUnit(a.currency.units[0], baseReserve, {
```
gre, line 67: for this, it's ok, it's the only legit case to use formatter as it's meant for user display purpose. you should not disable the grouping.

src/families/stellar/js-signOperation.js
```diff
@@ -0,0 +1,115 @@
...
+        }
+
+        // FIXME: Is this needed? (cf. log above)
+        /*
+        // Ensure amount is filled when useAllAmount
+        const transactionToSign = {
+          ...transaction,
+          amount: calculateAmount({ a: account, t: transaction }),
+        };
+        */
```
gre, line 74: please finalize a decision on this & remove code.

src/families/stellar/js-synchronization.js
```diff
@@ -0,0 +1,37 @@
+//@flow
+import { makeScanAccounts, makeSync, mergeOps } from "../../bridge/jsHelpers";
+import type { Account } from "../../types";
+import type { GetAccountShape } from "../../bridge/jsHelpers";
+import { fetchAccount, fetchOperations } from "./api";
+
+const getAccountShape: GetAccountShape = async (info) => {
+  const { id, address, initialAccount } = info;
```
gre, line 8: Did we confirm that migrating from libcore account will correctly works?

there is an implicit migration if users come with a libcore: account.
and i expect the output to be a `js:` id account. I fear a bit this do not happens here.
but maybe i fixed it generically in "makeSync" when i worked on the ethereum one

gre (changes requested):

gre (approved):

PR #1100, opened 2021-03-15
SOCKS proxy support

juan-cortes:
With the contribution from @G-Ray the apps seems to work with a valid provided SOCKS proxy. I've only tested it with the one launched when running the Tor Browser, it does not (as expected) work with proxies on other protocols but it would be nice to explore that since the original issue was users behind corporate http proxies if I recall correctly.

We need to test this further because there were some 422 errors when fetching the countervalue endpoints but I couldn't see any broken behaviour in the app, which is better than not working at all. We could probably do some error handling too when we detect an invalid SOCKS proxy value instead of crashing.

### DIFF
--- a/package.json
+++ b/package.json
@@ -124,6 +124,7 @@
     "semver": "^7.3.4",
     "sha.js": "^2.4.11",
     "stellar-sdk": "^7.0.0",
+    "socks-proxy-agent": "^5.0.0",
     "triple-beam": "^1.3.0",
     "winston": "^3.3.3",
     "xstate": "^4.16.2"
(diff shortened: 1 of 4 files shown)


gre:
Ok, for countervalues on LLD, I have feeling it's because the countervalues get fetched from the Electron renderer side.

One concern I have with the current approach is that it seems to only work for Node.js, the problem if we actually have 4 different JavaScript runtimes:
- Node.js (CLI and Ledger Live Desktop: for synchronisation, scan accounts, transactions,..)
- Electron Chromium V8 (Ledger Live Desktop: for UI things like Countervalues API, Status API, ..)
- React Native JSC (for Android)
- React Native JSC (for iOS – which can be a different version)
- and soon... Hermes to replace JSC in React Native


Essentially, we can't use a Node.js library i'm afraid. But maybe we can assume this is Node.js only and then we would need to see why this fails inside Chromium V8. We could bridge all networking to be done through a process, but it sucks because we would miss the ability to profile Networking in chrome debugger.

PR #1106, opened 2021-03-18
Stellar: check the correct memo size for hex formats

haammar-ledger:
Stellar `MEMO_HASH` and `MEMO_RETURN` memo types need to be exactly 32 bytes.
We currently checked that the hex string provided by the user was 32 characters, which is only 16 bytes.
Checking that the string is 64 characters does the trick.


gre:
can you rebase and include a `git revert 41f7e88f8430889ac3e0b1004d567a8d78292870` that would put back the Stellar js? thanks

gre:
@haammar-ledger please commit the all.libcore.js.snap from the snapshots-Linux.zip of the CI run https://github.com/LedgerHQ/ledger-live-common/actions/runs/664871775

it should be good then, for some reason, the PR was shipping a addition of a tx which actually is a no-diff with the previous implementation.

no hurry on this, we'll start again QA on this next week 👍

gre (approved):
CI pass again.
there is a small diff on the snapshot, but overall it looks ok,
one of the diff is on spendableBalance, but it looks correct, right? my understanding is that we were not correctly setting spendableBalance before and now we do? please confirm @haammar-ledger

haammar-ledger:
@gre Yes @henri-ly is telling me this is normal, as we were not taking Stellar trustlines into account.

Arnaud97234 (approved):

nabil-brn (approved):

PR #1108, opened 2021-03-18
COIN-1423 [Polkadot] Keep a safety buffer of (3 * fees) for bond max

haammar-ledger:
Customer Support reports that a lot of users do a "bond max" and are immediately locked, with not enough funds to pay the fees for any other transactions.

This quick fix reduces the amount when doing a bond max, by keeping an additional buffer of 3 times the fees.

More context in the issue: https://ledgerhq.atlassian.net/browse/COIN-1423


amougel:
@gre  I approve this change as discussed with @cthepot-ledger

rjarasson-ledger (approved):

gre (approved):

PR #1109, opened 2021-03-19
COIN-1440 [Polkadot] Set fees safety buffer to 0.1 DOT

haammar-ledger:
Follow-up of https://github.com/LedgerHQ/ledger-live-common/pull/1108

Context: https://ledgerhq.atlassian.net/browse/COIN-1440 linked to https://ledgerhq.atlassian.net/browse/COIN-1423

Set the safety buffer that we keep for paying fees of next transactions to 0.1 DOT, and use this amount for both `send max` and `bond max`.
- For send with manual amount: the user can go beyond (up to spendable) with a warning
- For bond: the user can't go beyond, not enough balance error.

### DIFF
--- a/src/families/polkadot/js-getTransactionStatus.js
+++ b/src/families/polkadot/js-getTransactionStatus.js
@@ -177,10 +177,6 @@ const getTransactionStatus = async (a: Account, t: Transaction) => {
         });
       }
 
-      if (t.useAllAmount) {
-        warnings.amount = new PolkadotAllFundsWarning();
-      }
-
       break;
 
     case "unbond":
--- a/src/families/polkadot/js-getTransactionStatus.js
+++ b/src/families/polkadot/js-getTransactionStatus.js
@@ -31,7 +31,7 @@ import { verifyValidatorAddresses } from "./api";
 import {
   EXISTENTIAL_DEPOSIT,
   MINIMUM_BOND_AMOUNT,
-  WARNING_FEW_DOT_LEFTOVER,
+  FEES_SAFETY_BUFFER,
   isValidAddress,
   isFirstBond,
   isController,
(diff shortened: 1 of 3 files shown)


src/families/polkadot/js-getTransactionStatus.js
```diff
@@ -290,9 +286,9 @@ const getTransactionStatus = async (a: Account, t: Transaction) => {
 
   if (
     t.mode === "bond" &&
-    a.spendableBalance.minus(totalSpent).lt(WARNING_FEW_DOT_LEFTOVER)
+    a.spendableBalance.minus(totalSpent).lt(FEES_SAFETY_BUFFER)
   ) {
-    warnings.amount = new PolkadotAllFundsWarning();
+    errors.amount = new NotEnoughBalance();
```
rjarasson-ledger, line 291: Why did you change this warning into an error ?
We don't want this behaviour i think.

rjarasson-ledger (changes requested):

gre:
tests are currently failing.

src/families/polkadot/test-dataset.js
```diff
@@ -252,25 +252,6 @@ const dataset: DatasetTest<Transaction> = {
                 warnings: {},
               },
             },
-            {
-              name: "bond extra - success but warn all funds",
```
rjarasson-ledger, line 256: Ok, but now we no longer have "success" tests anymore, because of the balance being too low on bot.

@gre can we do something about it ?
This PR requires having 0.1DOT available after transactions, so all existing tests cannot meet this predicate, we have either a warning, or an error.

haammar-ledger: Hmm... I could update the `bond extra - not enough spendable` test, with an amount that is right between "available" and "available - 0.1 DOT"?

gre: i don't understand because the test-dataset is supposely a static seed, not the one used by bot.
maybe we can send a bit on that sending account then. is it a static account? is it set up by @Arnaud97234  or not really? thx

rjarasson-ledger (approved):

gre:
- there is a situation with ETC that i reported again this morning. it is taking care of.
- since this test only have :x: on ETC it seems to be ok.
- i'll schedule a task to cover this feature in dataset + check potential implication for testing bot (we'll know when we merge and when it runs)

PR #1112, opened 2021-03-23
LL-4840 (Manager): disable install for given deprecated apps

LFBarreto:
(Manager): disable install for given deprecated apps:
- Crypto.com Chain

[LL-4839] [LL-4840]

[LL-4839]: https://ledgerhq.atlassian.net/browse/LL-4839
[LL-4840]: https://ledgerhq.atlassian.net/browse/LL-4840

### DIFF
--- a/src/manager/index.js
+++ b/src/manager/index.js
@@ -22,7 +22,12 @@ const ICONS_FALLBACK = {
   bitcoin_testnet: "bitcoin",
 };
 
-const oldAppsInstallDisabled = ["Crypto.com Chain"];
+const oldAppsInstallDisabled = [
+  "Crypto.com Chain",
+  "ZenCash",
+  "Ripple",
+  "Ontology",
+];
 const canHandleInstall = (app: ApplicationVersion) =>
   !oldAppsInstallDisabled.includes(app.name) &&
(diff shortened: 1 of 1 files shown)


src/manager/index.js
```diff
@@ -22,7 +22,7 @@ const ICONS_FALLBACK = {
   bitcoin_testnet: "bitcoin",
 };
 
-const oldAppsInstallDisabled = ["ZenCash", "Ripple", "Ontology"];
```
gre, line 25: we must keep them!

gre (changes requested):

src/manager/index.js
LFBarreto: i removed them since this function was not used on the live before and only on the test app. i figured it wasn't making sense to disable install of ripple app per instance :shrug:

gre: no, we must put them back.

- ZenCash is an old app that was renamed to Horizen
- Ripple is an old app that was renamed to XRP
- Ontology, same, it is ONT now.

we must keep this because we must not allow users to actually install these old apps.
the regression is probably on LLD side but we must keep this imo.

PR #1113, opened 2021-03-23
LL-4780 LL-4779 (Icons): update icons with react component script builder

LFBarreto:
(Icons): update icons with external library 
cryptocurrency-icons

updated svgr library and svg to react components script builder

This should handle arround 400 currencies based on ticker with tokens included;


valpinkman (approved):

valpinkman (approved):

src/react.js
```diff
@@ -1,15 +1,20 @@
...
 import * as icons from "./data/icons/react";
-import * as tokenIcons from "./data/icons/react/tokens";
 import type { CryptoCurrency, TokenCurrency } from "./types";
-import iconNameById from "./iconNameById";
 
 type Icon = React$ComponentType<{ size: number, color?: string }>;
 
+function getIconId({ ticker }: CryptoCurrency | TokenCurrency) {
+  let id = ticker.toLowerCase();
+  if (!isNaN(id.charAt(0))) id = `_${id}`; // fix variable name leading with a numerical value
```
gre, line 9: That approach is dangerous and I fear we will have a lot of mismatch.
Proof: how it's tedious to manage countervalues today, lot of collisions on ticker, etc..

for token only, we may need return `null` when `disableCountervalue` is true.

gre (changes requested):
You also did it for coins icons which was not scoped as part of the rework, so we'll need to be careful about this. please add this information in the description regarding impact of this.

- actually, there is currently a problem, if you test against **tools/** app, you can see it's not 1:1 with before. 

yalc locally and compare it with https://ledger-live-tools.vercel.app/currencies

Capture d’écran 2021-03-29 à 10 29 52

some icons are not iso, it's probably ok (confirm with Brian) but however we need to fix the scale.

- secondly, see my comment about tokens. it will not be great for colliding tokens.

- nice to have: it would be helpful to make a new page on /tools to visualize all the icons. (typically using "tickers by marketcap" sorting)

src/react.js
LFBarreto: this specific line is to handle tickers that start with a number since creating stuff like that cant work in JS 
```
import { 0xbtc } from "./0xbtc"
```
So i handled it that way since to escape this issue.

The new icons are named via ticker, tokens collision is not really an issue here since the icon color of tokens will be the one from the chain in the live.
Unless we have collision of ticker on currencies itself i dont think its an issue.

As for the scale it has been adapted on the frontend side with these two PR:
https://github.com/LedgerHQ/ledger-live-mobile/pull/1640
https://github.com/LedgerHQ/ledger-live-desktop/pull/3699

gre (approved):
let's go with it, we'll need to be clear on the bump that there is a light breaking changes on integration side to update the icon to new sizes.

PR #1115, opened 2021-03-24
[LL-3809] add tooltip field to device confirmation

henri-ly:
https://ledgerhq.atlassian.net/browse/LL-3809

We will need to add a field with translation key to implement a tooltip in desktop


gre (approved):
lgtm.

PR #1116, opened 2021-03-24
LL-4555 - Modification for wallet connect

machard:
transaction with 0 amount:
- Needed for wallet connect since the example dapp use transaction with amount = 0.
- also would be needed to cancel a previous transaction

sign message action
- allow to pass a command on desktop while remaining compatible as is with current funciton signature used by mobile

wallet connect provider
- clean current request id when disconnected

### DIFF
--- a/src/families/ethereum/modules/send.js
+++ b/src/families/ethereum/modules/send.js
@@ -10,6 +10,7 @@ import {
   NotEnoughBalance,
   FeeTooHigh,
   NotEnoughBalanceInParentAccount,
+  AmountRequired,
 } from "@ledgerhq/errors";
 import {
   inferTokenAccount,
@@ -59,6 +60,12 @@ const send: ModeModule = {
 
       if (!t.data) {
         if (
+          !t.allowZeroAmount &&
+          !result.errors.amount &&
+          result.amount.eq(0)
+        ) {
+          result.errors.amount = new AmountRequired();
+        } else if (
           !result.totalSpent.gt(0) ||
           result.totalSpent.gt(account.spendableBalance)
         ) {
--- a/src/families/ethereum/modules/send.js
+++ b/src/families/ethereum/modules/send.js
@@ -59,9 +59,7 @@ const send: ModeModule = {
       }
 
       if (!t.data) {
-        if (!result.errors.amount && result.amount.eq(0)) {
-          result.errors.amount = new AmountRequired();
-        } else if (
+        if (
           !result.totalSpent.gt(0) ||
           result.totalSpent.gt(account.spendableBalance)
         ) {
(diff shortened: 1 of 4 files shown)


src/families/ethereum/modules/send.js
```diff
@@ -59,9 +58,7 @@ const send: ModeModule = {
       }
 
       if (!t.data) {
-        if (!result.errors.amount && result.amount.eq(0)) {
-          result.errors.amount = new AmountRequired();
-        } else if (
```
gre, line 64: why would we want to send zero when data is not there? I guess some contract allows to do something like this? simple send without data i mean.

The current fix proposal is NOGO because it changes the behavior of Send flow and we don't want to do this, (introducing the possibility to send 0 on ETH). We have a long time history of doing that and rollbacking that (regarding UX of validation of send flow + many users would accidentally forget the amount, as we wouldn't block the NEXT).

I suggest a better fix is to have an explicit flag in the Transaction for you do say "allowZeroAmount" in context of wallet connect.

gre (dismissed):

src/families/ethereum/modules/send.js
machard: ok, will do. i was expeting that this change would not be ok in its current form ^^

src/families/ethereum/types.js
```diff
@@ -56,6 +56,7 @@ export type Transaction = {|
   estimatedGasLimit: ?BigNumber,
   feeCustomUnit: ?Unit,
   networkInfo: ?NetworkInfo,
+  allowZeroAmount?: Boolean,
```
gre, line 59: it's a subtility but i think it should be `boolean`

src/hw/signMessage/index.js
```diff
@@ -90,8 +95,9 @@ export const createAction = (
         return;
       }
       try {
-        result = await withDevice(device.deviceId)((t) =>
-          from(dispatch(t, request.message))
+        result = await (
+          signMessage(request) ||
+          withDevice(device.deviceId)((t) => from(dispatch(t, request.message)))
```
gre, line 100: according to the type, the signMessage can't be "falsy", i understand this is made for deprecation then.
can you add an explicit comment here to say "DEPRECATED: update client of live-common to explicit the signMessage implementation"

machard: actually i did not think that much. I did something that should be better now

src/families/ethereum/types.js
gre: Capture d’écran 2021-03-29 à 10 15 43

https://flow.org/try/#0PQKgBAAgZgNg9gdzCYAoVAXAngBwKZgCCYAvGAN6phgBGAXLXHDHgIYB2ANFWAEIO8mLDtwC+AbnQsMYVg2JlytBhgBOAVzyc+KjQQlA

gre (dismissed):

gre (approved):

PR #1122, opened 2021-03-29
Fixes swap to throw all errors

gre:
Before: swap was validating transaction and erroring if it's an error on recipient or amount
After: swap is validating **all** transaction cases and not just recipient or amount.

https://ledgerhq.atlassian.net/browse/LL-5092


juan-cortes (approved):
🤞🏼

PR #1123, opened 2021-03-29
LL-4222 - add data field to ethereum send module

machard:
https://ledgerhq.atlassian.net/browse/LL-4222

### DIFF
--- a/src/families/ethereum/modules/send.js
+++ b/src/families/ethereum/modules/send.js
@@ -108,8 +108,9 @@ const send: ModeModule = {
     }
     if (transaction.data?.length) {
       fields.push({
-        type: "data",
+        type: "text",
         label: "Data",
+        value: `Yes (${transaction.data.length} bytes)`,
       });
     }
   },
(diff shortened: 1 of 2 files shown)


src/families/ethereum/modules/send.js
```diff
@@ -99,13 +99,19 @@ const send: ModeModule = {
...
       fields.push({
         type: "amount",
         label: "Amount",
       });
     }
+    if (transaction.data?.length) {
+      fields.push({
+        type: "data",
+        label: "Data",
+      });
```
gre, line 113: do you think it's possible to use a `type: text` instead with the value that would display on device anyway? to me it's not to translate because device is in English.
and on long term future we want to kill all the extra type and try to only have type:text.

machard: yes { type: "text", "label": "Data", value: "YES/NO" }

machard: the only problem i see with this technique is regarding translation. maybe value should be an i18nkey

gre: TBD:
- is this the string that display on device? let's try to be ISO
- we don't need translation precisely because we need to be ISO and device is currently English only. if device in future gets translated, we'll reconsider this, but right now it's a choice done for similar parts.

PR #1127, opened 2021-03-30
LL-4555 LL-4222 - Wallet connect updates

machard:
https://ledgerhq.atlassian.net/browse/LL-4222
https://ledgerhq.atlassian.net/browse/LL-4555
~~https://ledgerhq.atlassian.net/browse/LL-4219~~ EDIT gre: we rollbacked that one. code is still here but commented.

- [x] allow send transaction with 0 amount with a special flag
- [x] sign message api update to allow desktop to pass its own sign command
- [x] wc: clean current request id when disconnected
- [x] add data field for send transaction with data
- [x] try to match device display for erc20 approve send transactions (50%, amount missing)


valpinkman (approved):

src/families/ethereum/modules/send.js
```diff
@@ -99,13 +106,60 @@ const send: ModeModule = {
...
         type: "amount",
         label: "Amount",
       });
     }
+    if (transaction.data?.length) {
+      const token = findTokenByAddress(transaction.recipient);
+      // $FlowFixMe (transaction data is not null, you flow)
+      const method = transaction.data.slice(0, 4).toString("hex");
+
+      if (method === "095ea7b3" && token) {
```
gre, line 124: 🤔 
to me this code belongs not to send but to `modules/erc20`

the code here feels quite difficult to maintain on long term and we don't want to add one entry for each future feature of ETH.

src/families/ethereum/modules/send.js
```diff
@@ -99,13 +106,60 @@ const send: ModeModule = {
...
+        fields.push({
+          type: "text",
+          label: "Amount (1/2)",
+          value: token.ticker,
+        });
+
+        // $FlowFixMe (transaction data is not null, you flow)
+        const amountHex = transaction.data.slice(36, 36 + 32).toString("hex");
+
+        if (uniq(amountHex.split("")) === ["f"]) {
```
gre, line 139: i'm not confident for us to maintain this ourself. we never did the "Amount (1/2)" display on Live side at the moment and i don't think we need to start doing this unless if it's addressed very generically.

gre (changes requested):
I think we need to rediscuss the task "display approve like on device" because the code here feels very complex to even maintain

gre (commented):
let's fix the lint and we should be good.

gre (changes requested):
let's fix the lint and we should be good.

gre:
@machard how do you estimate the impact of this work on mobile?
- is it a "no changes" for LLM perspective? if i bump just live-common alone on LLM, no impact / no changes to expect?

PR #1129, opened 2021-03-31
LL-5118 Deduct payout fees from swap operation history

juan-cortes:
### Description

We are currently storing the total amount (displayed on the device) instead of the total amount minus payout fees for the swap history entries. This can cause users to believe the should've received more and contact support. We should change LLC to deduct the payout fees if present for the floating rate trading method.

### Context
https://ledgerhq.atlassian.net/browse/LL-5118

### DIFF
--- a/src/exchange/swap/addToSwapHistory.js
+++ b/src/exchange/swap/addToSwapHistory.js
@@ -35,6 +35,11 @@ export default ({
       ? operation.subOperations[0].id
       : operation.id;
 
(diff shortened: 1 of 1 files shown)


valpinkman (approved):

valpinkman:
@juan-cortes LGTM but needs rebase

juan-cortes:
You got it chief @valpinkman

PR #1132, opened 2021-04-01
Enable floating rates in swap mocks

nabil-brn:
I just added some data to the table we return so the new button could work and is not disabled.


juan-cortes (approved):

juan-cortes:
You'll need to update the branch too

PR #1135, opened 2021-04-09
LL-4832 display actual content of error

valpinkman:
https://ledgerhq.atlassian.net/browse/LL-4832

Fix the error to display the actual message.

CleanShot 2021-04-09 at 18 26 45@2x


juan-cortes (approved):
lgtm

PR #1137, opened 2021-04-12
LL 4977 Fees strategy for BTC / ETH

henri-ly:


src/families/bitcoin/bridge/libcore.js
```diff
@@ -182,6 +185,10 @@ const getTransactionStatus = async (a, t) => {
     warnings.feeTooHigh = new FeeTooHigh();
   }
 
+  if (t.feesStrategy === null) {
```
gre, line 188: am I correct to see this is a breaking change in live-common? What is the default behavior of `ledger-live send` ?
ideally changes in live-common shouldn't break (meaning, i should be able to bump live-common on LLD/LLM without needed to change on LLM/LLD, so we need to make it a "no changes" by default, as much as possible)

henri-ly: my bad I totally forgot about LLC cli, it shouldn't work anymore...
The spec wasn't really clear about this, but I think I heard during a daily that we do need user to choose a fees speed strategy.
I would say we should probably enforce a `medium` strategy when we use `ledger-live send` and add option `--feesStrategy. But I don't know I felt something is wrong with this approach

gre: to me we should default to feesStrategy medium yes. because actually flows like Swap don't even allow to chose the fees so you can't even configure, we have to fallback on the "safe fee"

src/families/ethereum/modules/send.js
```diff
@@ -68,6 +69,10 @@ const send: ModeModule = {
           result.errors.amount = new NotEnoughBalance();
         }
       }
+
+      if (!result.errors.amount && !t.feesStrategy) {
+        result.errors.feesStrategy = new NoFeeStrategySelected();
```
gre, line 74: i'm not sure to understand the need of this this error?
to me this error is not possible if we settle feesStategy to a default, which we should as discussed earlier.

src/types/transaction.js
```diff
@@ -53,13 +54,25 @@ export type TransactionCommon = {|
...
   subAccountId?: ?string,
+  feesStrategy?: "low" | "medium" | "high" | "advanced",
 |};
 
 export type TransactionCommonRaw = {|
   amount: string,
   recipient: string,
   useAllAmount?: boolean,
   subAccountId?: ?string,
+  feesStrategy?: "low" | "medium" | "high" | "advanced",
```
gre, line 65: i think we should make this a non optional.

src/families/ethereum/modules/send.js
henri-ly: Well I add to cli-transaction a default value that would be set only on CLI, and stay undefined in LLM / LLD.
It was asked to force the user to select between one of these strategies in LLM and LLD.
Maybe should I just import that logic in both LLM / LLD ?

src/types/transaction.js
henri-ly: At the moment most of coins doesn't have the need to set this option

src/families/ethereum/modules/send.js
gre: the way to solve this is to put it as part of the "initTransaction" and not in the CLI inferTransactions

gre: createTransaction i mean https://github.com/LedgerHQ/ledger-live-common/blob/7ad2f9be55118ce258ded3f58bf85447be1da5d2/src/families/bitcoin/bridge/libcore.js#L62

henri-ly: Ok spec have change, so now we set "medium" by default and user will have "medium" by default, so we can remove the error 👍

valpinkman (approved):

gre (approved):

PR #1141, opened 2021-04-15
LL-5151 – portfolio & countervalues optimizations

gre:
### balance history cache

This introduced a new field on AccountLike types: `balanceHistoryCache` that store the balance history for various granularity so it can be accessed in O(1) complexity without need to recalculate with operations which was O(n) complexity.

This also effectively deprecate the old `balanceHistory` that is today in account and was partially used (in some libcore implementation) but it's essentially taking some ideas from it.

Today, Live need 3 granularity: HOUR, DAY, WEEK, and we will store it as array of balances for as long as the account existed. Here is an exemple (as seen in snapshot): 

Capture d’écran 2021-04-16 à 17 29 08

On a very extreme case (big accounts of mère denis) the current implementation has optimized by x10 the performance of portfolio V2 but it still about x2 slower than what V1 was with the "memoized" cache. (memoized cache that have known issue tho).

### coutervalues cache to use Map when possible

noticed a bottleneck on performance in JavaScript on usage of Object for a big map of data. Lookup appears to take some time and it's recommended to use Map for this. 

>  And last but not least, Map tends to perform better in storing large set of data, especially when keys are unknown until run time, and when all keys are the same type and all values are the same type. https://medium.com/front-end-weekly/es6-map-vs-object-what-and-when-b80621932373

### avoid objectSpread

There were a few more cases of usage of `{ ...foo, a, b }` which is very not recommended according to Performance i could measure when investigating bottleneck. one was pretty heavy on the "toOperationRaw/fromOperationRaw".


src/account/balanceHistoryCache.js
```diff
@@ -0,0 +1,135 @@
...
+): BalanceHistoryDataCache {
+  const { increment, startOf } = conf[g];
+  const latestDate = startOf(new Date());
+  const balances = [];
+  let { balance } = account;
+  const operationsLength = account.operations.length;
+  let date = latestDate.getTime();
+  for (let i = 0; i < operationsLength; ) {
+    // accumulate operations after time t
+    while (i < operationsLength && account.operations[i].date > date) {
```
JunichiSugiura, line 64: 👍

src/account/balanceHistoryCache.js
```diff
@@ -0,0 +1,135 @@
...
+      balance = balance.minus(
+        getOperationAmountNumberWithInternals(account.operations[i])
+      );
+      i++;
+    }
+    balances.unshift(BigNumber.max(balance, 0).toNumber());
```
gre, line 70: self note. I think it's more optimized to do `Math.max(balance.toNumber(), 0)`

gre, line 70: self note. I think it's more optimized to do `Math.max(balance.toNumber(), 0)`

src/account/balanceHistoryCache.js
```diff
@@ -0,0 +1,135 @@
...
+  const { increment, startOf } = conf[g];
+  const latestDate = startOf(new Date());
+  const balances = [];
+  let { balance } = account;
+  const operationsLength = account.operations.length;
+  let date = latestDate.getTime();
+  for (let i = 0; i < operationsLength; ) {
+    // accumulate operations after time t
+    while (i < operationsLength && account.operations[i].date > date) {
+      balance = balance.minus(
```
JunichiSugiura, line 65: Why we compute with BigNumber methods here instead of convert to number immediately?

JunichiSugiura: ah, sorry. it's a crypto balance not a countervalue 😅

gre:
I'm leaving the two tasks for later

- we need to see if it's good enough or if we can push it farther regarding cache and useMemo cache? but i feel it's not easy due to fact we have "now" datapoint meant to be everywhere.
- the cache only need to recalculate if balance changes OR operations list OR latestTime change. then we try to keep reference. and then we can optim up to a "useMemo".

and i think it's not critical for now.

machard:
could we have the balance history/countervalues/accounts working as a "backend service" in an other thread ?
The front end thread just query the info it needs to build the page, the worker manages all the complex time consuming logic to keep the db up to date and it does not impact rendering ? 
Its like having an embedded backend that runs on another thread than the render.

web worker // https://github.com/devfd/react-native-workers 

?

machard:
in simple words: remove everything react related from live-common and make of live common an embedded backend that run in a webworker / react native worker

gre:
@machard to me, it's the long term idea yes! actually an even more long term idea is to have this actually done on a backend, not even client side possibly,.. because ultimately this is on the path of allowing **pagination** => the balance history need to be calculated even if frontend don't have the full operations history available in future.

right now, i would not do what you suggested because it would introduce too much complexity and we want to deliver this first optimisation, also it would be very challenging to implement in React Native, which do not have natively any sort of thread on the JS side.

src/account/serialization.js
```diff
@@ -151,7 +173,15 @@ export const fromOperationRaw = (
...
-    ...op
+    id,
+    hash,
+    type,
+    senders,
+    recipients,
+    blockHeight,
+    blockHash,
+    transactionSequenceNumber,
+    hasFailed,
```
henri-ly, line 184: `hasFailed` and `transactionSequence` was optional, is that ok that we got `hasFailed: undefined` or `transactionSequenceNumber: undefined` on all operation object (see snapshot result) ?

gre: I think it's ok, unless we use `"transactionSequenceNumber" in t` anywhere

gre: i'll see if easy to optim

gre:
ok i need to fix flow

gre:
i need to remove the balanceHistoryCache from snapshot because it's always moving

src/account/balanceHistoryCache.js
```diff
@@ -0,0 +1,151 @@
...
+  HOUR: {
+    increment: hourIncrement,
+    startOf: startOfHour,
+  },
+};
+
+function generateHistoryFromOperationsG(
+  account: AccountLike,
+  g: GranularityId,
+  // partial=true allows a faster impleemntation that only recompose the last part of the history
```
JunichiSugiura, line 53: typo: `impleemntation` 😄

src/account/balanceHistoryCache.js
```diff
@@ -0,0 +1,151 @@
...
+    latestDate: null,
+    balances: [],
+  },
+  WEEK: {
+    latestDate: null,
+    balances: [],
+  },
+};
+
+const conf = {
```
JunichiSugiura, line 35: Can we merge this with conf from `src/portfolio/v2/range` ?

valpinkman (approved):

src/countervalues/logic.js
```diff
@@ -361,11 +381,9 @@ function generateCache(
   rateMap: RateMap,
   settings: CountervaluesSettings
 ): PairRateMapCache {
-  const map = { ...rateMap };
+  const map = new Map(rateMap);
 
-  const sorted = Object.keys(map)
-    .sort()
-    .filter((k) => k !== "latest");
+  const sorted = [...map.keys()].sort().filter((k) => k !== "latest");
```
JunichiSugiura, line 386: `Array.from(map.keys())`

src/countervalues/logic.js
```diff
@@ -380,33 +398,33 @@ function generateCache(
     // if autofillGaps is on, shifting daily gaps (hourly don't need to be shifted as it automatically fallback on a day rate)
     const now = Date.now();
     const oldestTime = oldestDate.getTime();
-    let shiftingValue = map[oldest];
+    let shiftingValue = map.get(oldest) || 0;
```
JunichiSugiura, line 401: `map.get(oldest) ?? 0` to be precise

src/portfolio/v2/index.js
```diff
@@ -97,7 +92,7 @@ export function getBalanceHistoryWithCountervalue(
   const counterValues = calculateMany(cvState, balanceHistory, {
     from: currency,
     to: cvCurrency,
-    disableRounding: true,
+    disableRounding: true, // FIXME why?
```
JunichiSugiura, line 95: Could you removed this line?

JunichiSugiura (approved):
No blocker except disableRounding.

src/countervalues/logic.js
gre: i'm not super confident by this idea. sometimes API returned null in past and we may want to be just "falsy case" fallback.

gre:
Untitled-2021-03-30-1406

PR #1142, opened 2021-04-19
LL 5253 Stellar JS: faster sync implementation

haammar-ledger:
Current implementation of account sync in Stellar JS does this:
1) Fetch all transactions of the account
2) For each transaction, fetch all operations it contains (the Stellar blockchain has both transactions, and operations contained in transactions)

This is sub-optimal and results in a significantly slower sync than with previous libcore impl, causing issues as described in https://ledgerhq.atlassian.net/browse/LL-5253.

This PR changes this mechanism to fetch directly all operations of an account in one pass.

**EDIT - more details on the fix**

Actually it's still done in two steps:
1) Fetch all operations of the account
2) Fetch the transaction of each operation

However this improves the situation because:
- it cuts down the nb of network requests in half
- the network requests are much faster this way around

**EDIT 2 - even better version**

Now it's really fetching operations + transactions at the same time, effectively reducing the number of requests to 1 per 200 operations. :ok_hand:


rjarasson-ledger:
LGTM, i still would like to test it to be sure on my side, but great improvements.

src/families/stellar/api/horizon.js
```diff
@@ -5,18 +5,17 @@ import { getEnv } from "../../../env";
 import { getCryptoCurrencyById, parseCurrencyUnit } from "../../../currencies";
 import type { Account, NetworkInfo, Operation } from "../../../types";
 import type { RawAccount, RawTransaction } from "./horizon.types";
```
gre, line 7: lint

gre (changes requested):

rjarasson-ledger:
We can still improve the requests by using the "join" request params to avoid N+1 requests.

gre:
BTW check the CI (windows) because it seems to fail.

```
● js bridge on account GRE's Stellar 2 › prepareTransaction › can be run in parallel and all yield same results

    Bad Request

      at new NetworkError (node_modules/stellar-sdk/src/errors.ts:25:4)
      at node_modules/stellar-sdk/lib/call_builder.js:332:41
      at step (node_modules/tslib/tslib.js:274:19)
      at Object.next (node_modules/tslib/tslib.js:205:16)
      at node_modules/tslib/tslib.js:177:69
      at Object.__awaiter (node_modules/tslib/tslib.js:156:12)
      at CallBuilder._handleNetworkError (node_modules/stellar-sdk/lib/call_builder.js:324:20)
          at runMicrotasks (<anonymous>)
```

gre:
still presence of   `Bad Request`  failures on CI

haammar-ledger:
@gre I know sorry about that, I'm on it, should be fixed soon(tm).

haammar-ledger:
@gre CI was failing because at some point the address passed here is empty: https://github.com/LedgerHQ/ledger-live-common/blob/master/src/bridge/jsHelpers.js#L129
I've sanitized the place where it was failing by checking against empty address, but I don't get why and how this can happen.

rjarasson-ledger (approved):

gre (approved):
great work guys. Go 🟦

PR #1143, opened 2021-04-19
LL-4612 Allow inline installation of applications in device flows

juan-cortes:
This functionality allows the user to install an application when, during a device flow, we detect that they don't have such application installed. This affects the Send, Receive, and Add account flows and enables the user to complete the flow without having to leave the modal/screen, go to the manager, install the application, and return.

This means that we can detach a bit more the user's awareness of what an app is, and what apps are installed on their device, minimise the friction on the flows too.

It's controlled by a new ENV variable `EXPERIMENTAL_INLINE_INSTALL` since we should thoroughly test this before releasing. First preliminary tests reveal that it works for both mobile and desktop although it's notably slow for mobile, but that's note caused by this.


src/hw/installApp.js
```diff
@@ -20,9 +20,14 @@ export default function installApp(
...
-    filter((e) => e.type === "bulk-progress"), // only bulk progress interests the UI
+    filter(
+      (e) =>
+        e.type === "bulk-progress" || // only bulk progress interests the UI
+        e.type === "device-permission-requested" || // exposing these for inline install
+        e.type === "device-permission-granted"
+    ),
     throttleTime(100), // throttle to only emit 10 event/s max, to not spam the UI
-    map((e) => ({ progress: e.progress })), // extract a stream of progress percentage
+    map((e) => (e.progress ? { progress: e.progress } : e)),
```
gre, line 30: i'm a bit scared by consequence of this & actually i feel we should maybe no longer remap the event to a `{progress}` specifically because that `e` of type "bulk-progress" already have a `.progress` (so tbd but shouldn't break that much). however, if you don't do it, then you have no clean way to "switch on type".

src/apps/runner.js
```diff
@@ -31,6 +32,7 @@ export const runAppOp = (
     // we need to allow a 1s delay for the action to be achieved without glitch (bug in old firmware when you do things too closely)
     defer(() => delay(getEnv("MANAGER_INSTALL_DELAY"))).pipe(ignoreElements()),
     defer(() => exec(appOp, deviceInfo.targetId, app)).pipe(
+      filter((e) => "progress" in Object(e)), // Ignore the exposed allow manager events from inline install
```
gre, line 35: that pattern is very weird (`in Object(e)`), can you explain what Object() is doing & why is it necessary?

src/apps/hw.js
```diff
@@ -31,6 +35,147 @@ export const execWithTransport = (transport: Transport<*>): Exec => (
 
 const appsThatKeepChangingHashes = ["Fido U2F"];
```
gre, line 37: Ok, I don't fully understand why we have to do so much code in this file. I feel it's getting repetitive with code in here.

specifically, why wasn't it possible to just use `listApps` function & then we pipe in this => https://github.com/LedgerHQ/ledger-live-common/blob/master/src/apps/runner.js#L61-L69
?

Not only we write again the same logic but also I feel this logic here don't cover all cases. typically, Bitcoin app might be outdated and we need to uninstall and reinstall it all. we have to use the same dependency logic because it's going to evolve in future and we don't want to create two legacy to maintain.



On paper, i think it could work like this:

```js
export const standaloneAppOp = (
  transport: Transport<*>,
  appOp: AppOp,
  exec: Exec
): Observable<State> =>
  concat(
    of({ type: "listing-apps" }),
    from(getDeviceInfo(transport)).concat(
      mergeMap((deviceInfo) =>
        listApps(transport, deviceInfo).pipe(
          filter((e) => e.type === "result"),
          map((e) => e.result),
          mergeMap((listAppsResult) =>
            runOneAppOp(initState(listAppsResult), appOp, exec)
          )
        )
      )
    )
  );

```

the only thing is it's not 1:1 with your types: (need Exec, also it returns observable of State, but it's relatively easy to map that to more precise information. specifically, you can have info of the apps to update, and progress events)

gre (changes requested):

src/apps/runner.js
juan-cortes: `Object(e)` guarantees that whatever the event is, will be an object so I can do the `key in obj` pattern. I think I had an issue where `e` would be either `null`, or undefined, then using this makes it work since it turns it into an empty object whereas doing `"progress" in e` with `e` either `undefined` or `null` throws an uncaught typeError. I see nothing inherently wrong in this fail safe but I'm open to alternatives.

src/api/socket.js
```diff
@@ -86,6 +86,7 @@ export const createDeviceSocket = (
                 o.next({
                   type: "device-permission-requested",
                   wording: "Ledger Manager",
+                  forManager: true,
```
gre, line 89: ok i don't understand why this is needed because I'm not seeing its usage in any place.

src/hw/connectApp.js
```diff
@@ -59,15 +65,20 @@ const openAppFromDashboard = (
...
         if (e && e instanceof TransportStatusError) {
           switch (e.statusCode) {
             case 0x6984:
             case 0x6807:
-              return of({ type: "app-not-installed", appName });
+              return getEnv("EXPERIMENTAL_INLINE_INSTALL")
+                ? concat(
+                    streamAppInstall(transport, appName),
+                    from(openAppFromDashboard(transport, appName))
+                  )
```
gre, line 80: sorry to add this requirement at last minute. but when I helped you on this, i noticed that an error thrown here would still "loop forever".
We must make sure throwing an error from there would be displayed to either because we can't afford having an infinite loop of hsm list apps when something is bad with our server for instance.

we must make sure we let the error flow and that it interrupts the upstream polling somehow.
IMO it's even a blocker for the "experimental" mode because if many users starts using that it's still bad for the load on our api.

src/api/socket.js
juan-cortes: dead code from the first version indeed, good catch

src/hw/connectApp.js
juan-cortes: As we talked on slack, I was unable to reach the endless loop when triggering an error inside the inlineInstall or the reducer. While there is still a chance it will happen, I am unable to reproduce, so if it's good enough to test (we could just not add it as an exposed experimental on the UI and make it only available if launched with the env for now) we could merge @gre

gre (approved):

PR #1147, opened 2021-04-21
[LL-2985] remove mandatory param freshaddressindex

henri-ly:
all the .receive from accountBridge warn that freshAddressIndex was missing, this should fix this


gre:
i did it on master directly, sorry didn't see the PR. unfortunately it still makes the flowtype mismatch, i guess we'll have to fix it anyway

gre:
for the record, `foo?: ?A` can be simplified to `foo?: A` and accepts `foo: undefined`

PR #1149, opened 2021-04-22
Fixes libcore to really support changing config

gre:
NB: i'm sending this PR for ci testing but it will be included as part of https://github.com/LedgerHQ/ledger-live-common/pull/1053

---

there is a libcore issue that makes this scenario not working:
getWallet --> updateWalletConfig --> getWallet

instead what this is doing is:
try { return createWallet } catch { updateWallet };  return getWallet

Scenario i was using is:

	VERBOSE=1 ledger-live sync --id dash:xpubhere
	EXPERIMENTAL_EXPLORERS=1 ledger-live sync --id dash:xpubhere
	VERBOSE=1 ledger-live sync --id dash:xpubhere

at any point in time it needs to use v2 or v3 but not stay on the first config used. this PR fixes it.

### DIFF
--- a/src/libcore/errors.js
+++ b/src/libcore/errors.js
@@ -20,3 +20,7 @@ export function isNonExistingAccountError(error: Error): boolean {
 export function isNonExistingWalletError(error: Error): boolean {
   return error.message.includes("doesn't exist");
 }
+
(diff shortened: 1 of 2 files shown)


gre:
it appears to be working (lint issue i fixed it on other PR).
closing to focus on other pr.

PR #1156, opened 2021-04-28
Allow connectApp command to have dependencies

juan-cortes:
### What is this
This is an evolution of the `connectApp` logic that allows us to provide an optional array of `AppRequest` (can be accounts, currencies, or app names) that will be required in order to complete the command. In case there are such dependencies we will check the apps on the device and either throw an error or attempt to install them if some are missing, or continue if all requirements are met.

It's completely plug and play into existing flows and requires close to no changes on LLD, basically changing the way we display the error case to account for multiple app dependencies.
image

### What can it be used for
For currently existing flows this enables us to remove the device wall on the swap feature and install all the missing dependencies but could potentially be used for all future features that can require plugins/apps on their flows.

### What is the device wall
The device wall is the current flow killer check we do when accessing the swap feature, if the user is missing a dependency app, or the exchange app itself we kill the user flow and prevent them from going further. In the case of the exchange app we do this even before the user has a chance to see what the feature is all about.


src/apps/hw.js
```diff
@@ -52,10 +51,15 @@ export type StreamAppInstallEvent =
...
+export const streamAppInstall = ({
+  transport,
+  appNames,
+  maybeOb,
+}: {
   transport: Transport<*>,
-  appName: string
-): Observable<StreamAppInstallEvent | ConnectAppEvent> =>
+  appNames: string[],
+  maybeOb?: () => Observable<*>,
```
gre, line 61: need more explicit naming.

src/apps/hw.js
```diff
@@ -70,20 +74,29 @@ export const streamAppInstall = (
...
-          });
-          if (isOutOfMemoryState(predictOptimisticState(state))) {
+          let state = initState(e.result);
+          appNames.forEach(
+            (name) => (state = reducer(state, { type: "install", name }))
+          );
+
+          if (
+            isOutOfMemoryState(predictOptimisticState(state)) ||
+            !getEnv("EXPERIMENTAL_INLINE_INSTALL")
```
gre, line 84: i don't understand why the `if` is here.

src/hw/actions/app.js
```diff
@@ -46,7 +47,7 @@ type State = {|
   isLoading: boolean,
   requestQuitApp: boolean,
   requestOpenApp: ?string,
-  requiresAppInstallation: ?{ appName: string },
+  requiresAppInstallation: ?{ appName: string } | ?{ appNames: string[] },
```
gre, line 50: in one side you try not to be too breaking (splitting into two actions) but on the state itself, i feel it's actually breaking (if UI side expects .appName)
so i have feeling we should just move to `Array<*>` paradigm everywhere to avoid all the ifs.

src/hw/requireApps.js
```diff
@@ -0,0 +1,82 @@
...
+  forceFormat?: string,
+|};
+
+export type Input = {
+  modelId: DeviceModelId,
+  devicePath: string,
+  appNames: string[],
+};
+
+const cmd = ({ devicePath, appNames }: Input): Observable<ConnectAppEvent> =>
```
gre, line 33: I don't understand why this is not done in actions/app , yet you add to change code here.

i miss context of this PR to have any conclusion on this.

let's rediscuss next week.

juan-cortes: That was the whole point of wanting you to take a look. I can rework it so it's in the same connectApp but it seemed odd because the end result is not "opening" an app. 🤔

src/apps/hw.js
juan-cortes: Doing the listapps and install plan allows me to know what _would be_ installed even if I don't finally do it because the user didn't opt in for inline installation. I need a way of knowing what is the error case data if you get what I mean

src/hw/actions/app.js
juan-cortes: Makes sense

src/apps/filtering.js
```diff
@@ -30,11 +30,16 @@ type UpdateAwareInstalledApps = {
...
+  const queries = query
+    .split(",")
+    .map((t) => t.toLowerCase().trim())
+    .filter(Boolean);
   const currency = currencyId ? getCryptoCurrencyById(currencyId) : null;
   const terms = `${name} ${
     currency ? `${currency.name} ${currency.ticker}` : ""
   }`;
-  return terms.toLowerCase().includes(query.toLowerCase().trim());
+  return queries.some((query) => terms.toLowerCase().includes(query));
```
juan-cortes, line 42: Would drop this if we go the delete all route

src/hw/connectApp.js
```diff
@@ -154,19 +160,28 @@ const cmd = ({
...
           concatMap((appAndVersion): Observable<ConnectAppEvent> => {
             timeoutSub.unsubscribe();
 
             if (isDashboardName(appAndVersion.name)) {
+              // check if we meet dependencies
+              if (dependencies?.length) {
+                return streamAppInstall({
+                  transport,
+                  appNames: [appName, ...dependencies],
+                  onSuccessObs: () => innerSub({ appName }), // NB without deps
```
juan-cortes, line 182: Return the same thing we just did, but without the dependency array since we completed the inline install

src/hw/actions/app.js
```diff
@@ -263,7 +264,7 @@ const reducer = (state: State, e: Event): State => {
         allowOpeningRequestedWording: null,
         allowManagerGranted: false,
         allowManagerRequestedWording: null,
-        requiresAppInstallation: { appName: e.appName },
+        requiresAppInstallation: { appNames: e.appNames },
```
gre, line 267: this requires a change on UI, correct? should we temporarily still have a `appName: e.appNames[lastIndex]` ?

gre (commented):
minor concern regarding one possible breaking change

src/hw/actions/app.js
juan-cortes: This would definitely need a minor adaptation of the UI, but it's aligned with the redesign that we discussed today with @cthepot-ledger, I think passing appName temporarily to not break at all can be fine if you deem it necessary

src/apps/hw.js
```diff
@@ -70,20 +74,31 @@ export const streamAppInstall = (
...
-            type: "install",
-            name: appName,
-          });
-          if (isOutOfMemoryState(predictOptimisticState(state))) {
-            // it will not be possible to install so we will fallback to app-not-installed.
-            return of({ type: "app-not-installed", appName });
+          let state = initState(e.result);
+          appNames.forEach(
+            (name) => (state = reducer(state, { type: "install", name }))
+          );
```
gre, line 80: you can refactor this to:

```
const state = appNames.reduce((state, name) => reducer(state, { type: "install", name }), initState(e.result));
```

as it's a reducer :)

gre (commented):
i have one slight request to polish but to me we are good.

src/apps/hw.js
juan-cortes: nice, nice

juan-cortes: You got it

gre (approved):

PR #1157, opened 2021-04-29
StellarJS: workaround Stellar SDK bad error handling

ghost:
**How to reproduce**
On Android, at the end of an add accounts for Stellar, a 404 error message is systematically displayed.

**Root cause**
The Stellar SDK badly handles network answers from the underlying explorer (see https://github.com/stellar/js-stellar-sdk/issues/638), which makes it impossible for us to catch end of add accounts in a clean way.

**Fix**
Only dirty workarounds are possible at our level... we have no other way than to parse the errors the Stellar SDK returns, which are supposed to be cleanly typed but are actually completely different things depending on the cases...

**Things to test for QA**
- Add accounts on LLD with network connectivity
- Add accounts on LLD without network connectivity
- Add accounts on LLM with network connectivity
- Add accounts on LLM without network connectivity

### DIFF
--- a/src/families/stellar/api/horizon.js
+++ b/src/families/stellar/api/horizon.js
@@ -1,6 +1,10 @@
 //@flow
 import { BigNumber } from "bignumber.js";
-import StellarSdk, { AccountRecord } from "stellar-sdk";
+import StellarSdk, {
+  AccountRecord,
+  NotFoundError,
+  NetworkError,
+} from "stellar-sdk";
 import { log } from "@ledgerhq/logs";
 import { getEnv } from "../../../env";
 import { getCryptoCurrencyById, parseCurrencyUnit } from "../../../currencies";
@@ -9,6 +13,7 @@ import {
   getAccountSpendableBalance,
   rawOperationsToOperations,
 } from "../logic";
+import { NetworkDown } from "@ledgerhq/errors";
 
 const LIMIT = getEnv("API_STELLAR_HORIZON_FETCH_LIMIT");
 const FALLBACK_BASE_FEE = 100;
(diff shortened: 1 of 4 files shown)


gre:
Is there a test that could assess the previous behavior is fixed? i'm not sure to understand exactly what to tell our QA regarding what was the problem.

src/families/stellar/api/horizon.js
```diff
@@ -123,9 +128,22 @@ export const fetchOperations = async (
...
       return [];
     }
+    if (
+      e instanceof NetworkError ||
+      (e && e.toString().match(/Error: getaddrinfo ENOTFOUND/g)) ||
+      (e && e.toString().match(/undefined is not an object/g))
+    ) {
+      throw new NetworkDown();
+    }
+
```
rjarasson-ledger, line 146: I don't understand this part.
Can you add a helper to "remap errors" for the whole logic ?
It should be reused everywhere there is a catch all errors that could lead to network issues.

`.match(/status code 404/)` -> no throw (return logic after the remap)

`.match(/status code 4[0-9]{2}/)` => LedgerAPI4XX

`.match(/status code 5[0-9]{2}/)` => LedgerAPI5XX

`.match(/ECONNRESET|ECONNREFUSED|ENOTFOUND|EPIPE|ETIMEDOUT/)` => NetworkDown

ghost (changes requested):

src/families/stellar/api/horizon.js
haammar-ledger: There's not always a status code, sometimes it's a raw JS `undefined` error coming from the Stellar SDK...

haammar-ledger: But I can replace this one `.match(/Error: getaddrinfo ENOTFOUND/g))` with the more detailed cases you mention.

rjarasson-ledger: yeah ok, so we can just add some cases. Strange behaviour though. Idea is just to handle as much (known) cases as possible.

ghost (approved):

gre:
thanks for the comment; important tech debt on our side to track.

PR #1158, opened 2021-04-29
added logic so announcement refresh themselves on parameter change

IAmMorrow:
This PR should ensure that announcement are correctly refreshed whenever parameters change. Please be careful that client implementation don't change the context object on every render, but rather when a param changed.

### DIFF
--- a/src/notifications/AnnouncementProvider/index.js
+++ b/src/notifications/AnnouncementProvider/index.js
@@ -1,6 +1,6 @@
 // @flow
 
-import React, { createContext, useMemo, useCallback, useContext } from "react";
+import React, { createContext, useMemo, useCallback, useContext, useEffect } from "react";
 import differenceBy from "lodash/differenceBy";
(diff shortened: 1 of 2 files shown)


gre (changes requested):
check linter.

valpinkman (changes requested):
flow and eslint issues

otherwise seems good to me

gre (approved):

src/notifications/AnnouncementProvider/index.js
```diff
@@ -123,6 +129,10 @@ export const AnnouncementProvider = ({
     },
   });
 
+  useEffect(() => {
+    send({ type: "UPDATE_DATA" });
+  }, [context]);
```
gre, line 134: i don't really understand how this works. it is called on mount no?

gre:
@IAmMorrow no update on this so as discussed i'm closing. please create equivalent on Jira so we can plan the work.

PR #1160, opened 2021-04-30
bot-silicium <> libcore 4.2

gre:


gre:
Dataset tests are promising. However bot seems to crash badly. Let's dig next week

gre:
https://github.com/LedgerHQ/ledger-live-common/actions/runs/806492667
@hzheng-ledger can you check in bot-logs.txt if you see anything suspicious? It seems the bot crash after 6mn where it is supposed to run for a longer time and then send a comment on GitHub (comment don't appear so I suspect a segfault)

valpinkman (approved):

cli/package.json
```diff
@@ -36,7 +36,7 @@
     "@ledgerhq/hw-transport-mocker": "5.51.1",
     "@ledgerhq/hw-transport-node-hid": "5.51.1",
     "@ledgerhq/hw-transport-node-speculos": "^5.51.1",
-    "@ledgerhq/ledger-core": "beta",
+    "@ledgerhq/ledger-core": "7.0.0-beta.4",
```
gre, line 39: we shouldn't bump somethng that is not yet prebuild released.

- https://github.com/LedgerHQ/lib-ledger-core-node-bindings/runs/2620426034 is failing
- it's not yet published https://github.com/LedgerHQ/lib-ledger-core-node-bindings/releases

these two actions are required for prebuild to be fetchable. (otherwise we enter unexpected bugs if there are no prebuilds)

.github/workflows/bot3.yml
```diff
@@ -48,11 +48,9 @@ jobs:
           GITHUB_RUN_ID: ${{ github.run_id }}
           GITHUB_WORKFLOW: ${{ github.workflow }}
           SLACK_API_TOKEN: ${{ secrets.SLACK_API_TOKEN }}
+          VERBOSE_LIBCORE: "1"
         run: COINAPPS=$PWD/coin-apps yarn ci-test-bot
         timeout-minutes: 120
-      - name: Run coverage
-        if: failure() || success()
-        run: CODECOV_TOKEN=${{ secrets.CODECOV_TOKEN }} npx codecov
```
gre, line 55: can we put this back on?

src/libcore/getOrCreateAccount.js
```diff
@@ -24,7 +24,9 @@ export const getOrCreateAccount: F = atomicQueue(
     log("libcore", "getOrCreateAccount", { xpub, index });
     let coreAccount;
     try {
+      log("libcore", "getting libcore account");
       coreAccount = await coreWallet.getAccount(index);
+      log("libcore", "got libcore account");
```
gre, line 29: can you remove these two?

src/libcore/platforms/nodejs.js
```diff
@@ -369,9 +369,7 @@ export default (arg: {
           } = statics[method];
           if (njsInstanciateClass) {
             m[method] = function met(...vargs) {
-              if (process.env.VERBOSE) {
-                log("libcore-call", id + "." + method, vargs);
-              }
+              log("libcore-call", id + "." + method, vargs);
```
gre, line 372: please either undo, or use an env. if you don't use the env, we must completely remove it 'VERBOSE_LIBCORE' didn't seem to work.

gre (commented):
@hzheng-ledger we're almost there! diff reveal some things that i think we should revert. at the end the only diff is the libcore bump + the fix on pickInputs i think + the snapshot fix.

PR #1164, opened 2021-05-03
Fix the reported Compound supply values – LL-5470

stephenjust:
Previously, the reported supply balance was inaccurate. First, we were
using the transaction time to look up the historical exchange rate. If
the tx was in the mempool for a while, this could create a significant
delta to when the block was written.

* Updated fetchHistoricalRates to use the block depth, this is more
  accurate than querying based on transaction time.
* Use the underlying token transaction instead of using the historical
  rate for calculation. Even the block number-based implementation has
  noticeable error.

Technically, fetchHistoricalRates isn't needed anymore, but I have left it in
place to not affect anything that may depend on `extra.rate`.


gre:
@stephenjust thanks for your contribution! we'll review the PR in our sprint.

gre:
@stephenjust can you run `yarn prettier` ? thanks

src/families/ethereum/modules/compound.js
```diff
@@ -396,11 +396,16 @@ export function prepareTokenAccounts(
   return subAccounts.concat(implicitCTokenAccounts);
 }
 
-const cdaiToDaiOpMapping: { [_: OperationType]: ?OperationType } = {
+const ctokenToGeneratedTokenOpMapping: { [_: OperationType]: ?OperationType } = {
```
gre, line 399: @stephenjust can you run `yarn prettier` ? thanks

gre (changes requested):

JunichiSugiura (approved):

src/families/ethereum/modules/compound.js
stephenjust: Looks like Junichi got it. Thanks!

src/families/ethereum/modules/compound.js
```diff
@@ -471,23 +478,35 @@ export async function digestTokenAccounts(
         // cOUT => REDEEM
         const rates = await fetchHistoricalRates(
           ctoken,
-          ctokenAccount.operations.map((op) => op.date)
+          ctokenAccount.operations.reduce(
+            (prev, op) => (op.blockHeight ? [...prev, op.blockHeight] : prev),
```
gre, line 482: - maybe reduce+spread is not helping perf, could do more specific logic maybe?
- would we want to dedup even tho there are low chance to have two ops on same block?

gre (commented):
overall it's ok. i have a small comment on one part.

I think we'll wait to validate internally the fix and we will be good to merge.

gre:
I've tested and compared on my account the two implementations, on the small amounts I have on my test account there are no significant change only 1 cent here and here, which means it's a GO for this PR as it seems to slightly adjust the number to their more correct block.

Thanks @stephenjust for the contribution.

PR #1166, opened 2021-05-03
added LL platform types and logic

IAmMorrow:


src/platform/serializers.js
```diff
@@ -0,0 +1,151 @@
...
+}
+
+export function deserializePlatformTransaction(
+  rawTransaction: RawPlatformTransaction
+): PlatformTransaction {
+  switch (rawTransaction.family) {
+    case "ethereum":
+      return deserializePlatformEthereumTransaction(rawTransaction);
+    case "bitcoin":
+      return deserializePlatformBitcoinTransaction(rawTransaction);
```
gre, line 134: note for future: in future we want to split these into modules ethereum.js , bitcoin.js etc..
actually even better is to have a file `src/families/*/platform.js` that export the coin specifics.

gre (approved):
please ack on the feedback. merging it now.

PR #1168, opened 2021-05-06
LL-5519 Fix scan account error handling

JunichiSugiura:
The cause of scan account error (we believe) was due to the nesting try statement doesn't delegate errors to top level try statement.
This PR removes the nesting try catch statement all together.

We need to make sure this change doesn't affect any coins we support.

we must also try with older firmware.


gre (changes requested):
Lint fails

gre:
I've tested BTC, BCH, ETH, ETC. all is good. we should be good.

gre (approved):

PR #1186, opened 2021-05-14
Thomas/add crypto org

thomas-nguy:
Implement skeleton and basic functionnalities for crypto.org

Feature added : 
- sync enable : ( ledger-live sync -c cro) Fetch all the transaction history
- send transaction : (ledger-live send -c cro) Send token to a single address

This branch needs to use a version of ledgerjs which has this ([PR](https://github.com/LedgerHQ/ledgerjs/pull/590)) merged


src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,218 @@
...
+import { BigNumber } from "bignumber.js";
+import network from "../../../network";
+
+import type { Operation, OperationType } from "../../../types";
+import { getEnv } from "../../../env";
+import { encodeOperationId } from "../../../operation";
+
+const CRYPTO_ORG_INDEXER = getEnv("CRYPTO_ORG_INDEXER");
+const CRYPTO_ORG_RPC_URL = getEnv("CRYPTO_ORG_RPC_URL");
+const PAGINATION_LIMIT = 20;
```
thomas-nguy, line 19: I have hard coded this value by default to limit the number of result returned by the transaction history endpoints (it does have pagination). Let me know if this value is fine.

haammar-ledger: You can probably use a higher value (in other integrations we use from 50 to 200).
Basically the higher the better since you will need less requests to fetch all data, but we don't want too high response times.
Try with 200, then if a response takes more than a couple seconds you can try lower values.

haammar-ledger:
Early feedback: an empty wallet will cause an HTTP 404 on the explorer, which currently causes sync to fail with a `LedgerAPI4xx` error:
`404 GET https://crypto.org/explorer/croeseid/api/v1/accounts/tcro1urt70xd2naehl3jxtjhpsrmjhaspd2az3r734f (2193ms): Record not found {}`

You need a way to detect and support empty accounts.

haammar-ledger:
Is it normal that `crypto_org` gives a `tcro` address?
```
$ ledger-live getAddress -c crypto_org --path "44'/394'/0'/0/0" --derivationMode ""
{
  address: 'tcro1urt70xd2naehl3jxtjhpsrmjhaspd2az3r734f',
  publicKey: '03d13148dccd0e094d64d2d642b8d4fcea8451cc479c0082e826c240769acfaf8a',
  path: "44'/394'/0'/0/0"
}

```

thomas-nguy:
Thanks for your comment

```
Early feedback: an empty wallet will cause an HTTP 404 on the explorer, which currently causes sync to fail with a LedgerAPI4xx error:
404 GET https://crypto.org/explorer/croeseid/api/v1/accounts/tcro1urt70xd2naehl3jxtjhpsrmjhaspd2az3r734f (2193ms): Record not found {}

You need a way to detect and support empty accounts.
```

Regarding this, do you have a suggestion how to handle it?

Currently if I try to sync with a non existing account, I receive 

"Record not found"

Is it not an acceptable behavior?

haammar-ledger:
For empty accounts, since they translate to the indexer and node answering 404, what you can do is something like that:
```
  let data;
  try {
    const response = await network({method: "GET", url: `${CRYPTO_ORG_INDEXER}/accounts/${addr}`});
    data = response.data;
  } catch (error) {
    if (error?.status !== 404) {
      throw error;
    }
  }
```
Then you have to make sure your empty `data` object is correctly handled afterwards.
The same applies to calls to the node eg. `const response = await client.getAccount(addr);`

src/families/crypto_org/types.js
```diff
@@ -0,0 +1,64 @@
...
+
+
+export const reflect = (declare: (string, Spec) => void) => {
+  return {
+    OperationMethods: {
+    },
+    AccountMethods: {
+    },
+  };
+}
```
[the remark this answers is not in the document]
thomas-nguy: thanks, fixed

src/families/crypto_org/types.js
```diff
@@ -0,0 +1,64 @@
+// @flow
+
+import type { Spec } from "../../libcore/types";
```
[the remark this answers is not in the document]
thomas-nguy: fixed

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,218 @@
...
+        const msgSend: CryptoOrgMsgSendContent = msgs[j].content;
+        rawTransactions.push(
+          transactionToOperation(
+            accountId,
+            addr,
+            msgSend,
+            accountTransactions[i]
+          )
+        );
+      }
```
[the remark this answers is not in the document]
thomas-nguy: added a switch to make it more clean. does it help?

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,218 @@
...
+    }
+  }
+
+  return rawTransactions;
+};
+
+/**
+ * Broadcast blob to blockchain
+ */
+export const submit = async (blob: string) => {
```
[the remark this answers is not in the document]
thomas-nguy: fixed

src/families/crypto_org/js-getTransactionStatus.js
```diff
@@ -0,0 +1,60 @@
...
+
+  const amount = useAllAmount
+    ? a.balance.minus(estimatedFees)
+    : BigNumber(t.amount);
+
+  if (totalSpent.gt(a.balance)) {
+    errors.amount = new NotEnoughBalance();
+  }
+
+  // If MyCoin needs any specific requirement on amount for instance
```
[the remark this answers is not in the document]
thomas-nguy: removed

src/families/crypto_org/logic.js
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Returns true if address is a valid md5
+ *
+ * @param {string} address
+ */
+export const isValidAddress = (address: string): boolean => {
+  if (!address) return false;
+
+  const network = CRYPTO_ORG_USE_TESTNET
+    ? CroNetwork.Testnet
```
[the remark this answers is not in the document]
thomas-nguy: thanks, fixed

src/families/crypto_org/js-buildTransaction.js
```diff
@@ -0,0 +1,59 @@
...
+import type { Transaction } from "./types";
+import type { Account } from "../../types";
+import { CroNetwork, CroSDK, Units, utils } from "@crypto-com/chain-jslib";
+import { getAccountParams } from "./api/sdk";
+import { getEnv } from "../../env";
+import { TestnetCroeseid3 } from "./logic";
+
+let sdk = getEnv("CRYPTO_ORG_USE_TESTNET")
+  ? CroSDK({ network: TestnetCroeseid3 })
+  : CroSDK({ network: CroNetwork.Mainnet });
```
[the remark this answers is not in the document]
thomas-nguy: correct, I will refactor this part 👍

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,223 @@
...
+import { encodeOperationId } from "../../../operation";
+
+const CRYPTO_ORG_INDEXER = getEnv("CRYPTO_ORG_USE_TESTNET")
+  ? getEnv("CRYPTO_ORG_TESTNET_INDEXER")
+  : getEnv("CRYPTO_ORG_INDEXER");
+const CRYPTO_ORG_RPC_URL = getEnv("CRYPTO_ORG_USE_TESTNET")
+  ? getEnv("CRYPTO_ORG_TESTNET_RPC_URL")
+  : getEnv("CRYPTO_ORG_RPC_URL");
+const CRYPTO_ORG_USE_TESTNET = getEnv("CRYPTO_ORG_USE_TESTNET");
+const PAGINATION_LIMIT = 200;
```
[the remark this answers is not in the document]
thomas-nguy: thanks, fixed

src/families/crypto_org/hw-getAddress.js
```diff
@@ -0,0 +1,21 @@
...
+
+import Cosmos from "@ledgerhq/hw-app-cosmos";
+import type { Resolver } from "../../hw/getAddress/types";
+import { getEnv } from "../../env";
+
+const CRYPTO_ORG_USE_TESTNET = getEnv("CRYPTO_ORG_USE_TESTNET");
+
+const resolver: Resolver = async (transport, { path, verify }) => {
+  const cosmos = new Cosmos(transport);
+  const cointype = CRYPTO_ORG_USE_TESTNET ? "tcro" : "cro";
```
gre, line 11: we should use the currency.id to determine if we use testnet or not

as we have a currency `crypto_org_croeseid` we probably can use it to know if we want a testnet or not :)

 https://github.com/LedgerHQ/ledgerjs/blob/master/packages/cryptoassets/src/currencies.js#L2735

basically overall we should try to not use `CRYPTO_ORG_USE_TESTNET ` but use the account.currency / currency contextual info to know if you are on testnet or not. that way it allows to have both testnet accounts and normal accounts at same time & prevent to need an env.

src/families/crypto_org/js-synchronisation.js
```diff
@@ -0,0 +1,52 @@
...
+    blockHeight,
+    balance,
+    bondedBalance,
+    redelegatingBalance,
+    unbondingBalance,
+    commissions,
+  } = await getAccount(address);
+
+  // Merge new operations with the previously synced ones
+  const newOperations = await getOperations(id, address, startAt);
```
gre, line 28: to support pagination we would need to loop here. a bit like done here. 

https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/ethereum/synchronisation.js#L393-L408

gre (approved):

src/families/crypto_org/js-synchronisation.js
thomas-nguy: thanks, I added a loop.
(I also notice that the startAt takes the blockHeight, which does not work in my case because its just an incremental number. Not sure if its possible to store the "last index" fetch in the operation object to reduce the number of call)

src/families/crypto_org/hw-getAddress.js
thomas-nguy: do you know how to get the currency in a static way?  
Also I am confused how can I switch currency from mainnet to testnet?  (by default, if I plug my ledger and open the crypto_org app, it uses the currency "crypto_org" , not "crypto_org_testnet")

gre: > how can I switch currency from mainnet to testnet?

```sh
ledger-live sync -c crypto_org
ledger-live sync -c crypto_org_croeseid
```

would sync different accounts.


> do you know how to get the currency in a static way?

```js
import { getCryptoCurrencyById } from "../../currencies";

getCryptoCurrencyById("crypto_org_croeseid")
```

src/families/crypto_org/js-synchronisation.js
gre: i'm not exactly sure what you mean. We don't enforce anything here, startAt is something you have internally in your implementation here, so you can do whatever you want, it's not generalised yet at this stage, the idea is just to "fetch all operations", in an incremental way (e.g. if user come with an account already filled, it will just ask the last page).

src/families/crypto_org/hw-getAddress.js
thomas-nguy: thanks!

```
import { getCryptoCurrencyById } from "../../currencies";

getCryptoCurrencyById("crypto_org_croeseid")
```

For this part, I mean how to get the currency from user input (-c) in a static way?

src/families/crypto_org/js-synchronisation.js
thomas-nguy: Yeah I mean, I think it would be more optimal to store the index 'startA' in the Account so that you don't need to fetch all the pages every time. 
In the implementation in the example, they use the field "blockHeight" from the account before the 'startA' takes the blockHeight as a value instead of index. 

In my case I dont know if there is any field in Account where I can store that. It is not a big deal though, we can ignore this issue.

src/families/crypto_org/hw-getAddress.js
gre: Currency is part of the options you have in second parameter

`{ path, verify, currency }`

https://github.com/LedgerHQ/ledger-live-common/blob/dcf2db107d8551271e43517139845e4a00cd1c99/src/hw/getAddress/types.js#L26

src/families/crypto_org/js-synchronisation.js
gre: makes sense. let's consider that in future iterations indeed.

src/families/crypto_org/hw-getAddress.js
thomas-nguy: thanks, I removed the env variable and am using the currency in account to do the switch now

src/env.js
```diff
@@ -145,6 +145,26 @@ const envDefinitions = {
...
+    parser: stringParser,
+    desc: "location of the crypto.org indexer API",
+  },
+  CRYPTO_ORG_TESTNET_INDEXER: {
+    def: "https://crypto.org/explorer/croeseid3",
+    parser: stringParser,
+    desc: "location of the crypto.org indexer testnet API",
+  },
+  CRYPTO_ORG_RPC_URL: {
+    def: "",
```
haammar-ledger, line 159: Missing value here.
```
$ ledger-live sync -c crypto_org
Expected string `endpoint` to be an url, got ``
```

gre (approved):
lgtm

ghost (changes requested):

src/env.js
thomas-nguy: yes please only use crypto_org_croeseid for now. Mainnet node is still being build but because they are overloaded with tasks, it might takes month for them to setup

haammar-ledger:
Also please rebase on origin.

src/env.js
haammar-ledger: Noted :+1:

ghost (approved):

src/families/crypto_org/js-signOperation.js
```diff
@@ -0,0 +1,165 @@
...
+import {
+  CryptoOrgWrongSignatureHeader,
+  CryptoOrgSignatureSize,
+} from "./errors";
+import type { Transaction } from "./types";
+import type { Account, Operation, SignOperationEvent } from "../../types";
+
+import { open, close } from "../../hw";
+import { encodeOperationId } from "../../operation";
+import CryptoOrgApp from "@ledgerhq/hw-app-cosmos";
```
rjarasson-ledger, line 14: Is it ok to use the hw js bindings for the cosmos app ?
Wouldn't be more future-proof to use a cryptoorg hw app, since we don't know if cosmos app would always follow the same lifecycle ?

src/families/crypto_org/js-signOperation.js
```diff
@@ -0,0 +1,165 @@
...
+          if (sLen === 33) {
+            sOffset++;
+          } // as above
+          const sigR = signature.slice(rOffset, rOffset + rLen); // skip e.g. 3045022100 and pad
+          const sigS = signature.slice(sOffset);
+
+          const signatureFormatted = Buffer.concat([sigR, sigS]);
+          if (signatureFormatted.length !== 64) {
+            throw new CryptoOrgSignatureSize();
+          }
```
rjarasson-ledger, line 128: I have a feeling that this part should be in a separate helper, or handle by the hw app js lib.

src/families/crypto_org/js-synchronisation.js
```diff
@@ -0,0 +1,55 @@
...
+
+  // Merge new operations with the previously synced ones
+  let startAt = 0;
+  let maxIteration = 20;
+  let operations = oldOperations;
+  let newOperations = await getOperations(id, address, startAt++, useTestNet);
+  do {
+    operations = mergeOps(operations, newOperations);
+    newOperations = await getOperations(id, address, startAt++, useTestNet);
+  } while (--maxIteration && newOperations.length != 0);
```
rjarasson-ledger, line 32: This pagination handling logic should be in the api i think.
I think mergeOps is quite a resource consuming operation (merges arrays by comparing ids).

Plus, it always starts at zero here, and does not take into consideration the previous synced operation.
LedgerLive sync is iterative, every 2 minutes, each account is synced, but inherits from its previously fetched state.

Here, i don't see that this is taken into consideration, and might burden the api heavily. Can't you set a minimum blockHeight as a starting point when fetching operations ?

src/families/crypto_org/logic.js
```diff
@@ -0,0 +1,63 @@
...
+  rpcUrl: "https://testnet-croeseid-3.crypto.org:26657",
+};
+
+let croSdk = null;
+
+/**
+ * Get CroSdk
+ * @param {boolean} useTestNet
+ */
+export function getCroSdk(useTestNet: boolean) {
```
rjarasson-ledger, line 32: Kind of the same singleton issue as for API here.
If multiple accounts are added to Ledger Live with different networks, it might not work.

ghost (approved):

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,244 @@
...
+  useTestNet: boolean
+) => {
+  let result = BigNumber(0);
+  const currency = useTestNet ? CryptoOrgTestnetCurrency : CryptoOrgCurrency;
+  amounts.map(function (currentObject) {
+    if (currentObject.denom == currency)
+      result = result.plus(BigNumber(currentObject.amount));
+  });
+
+  return result;
```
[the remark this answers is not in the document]
thomas-nguy: fixed

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,244 @@
...
+  let result = BigNumber(0);
+  const amounts = messageSendContent.amount;
+  for (let k = 0; k < amounts.length; k++) {
+    const amount: CryptoOrgAmount = amounts[k];
+    const currency = useTestNet ? CryptoOrgTestnetCurrency : CryptoOrgCurrency;
+    if (amount.denom == currency) {
+      result = result.plus(amount.amount);
+    }
+  }
+  return result;
```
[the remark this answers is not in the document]
thomas-nguy: yes, fixed

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,244 @@
...
+
+  const { data } = await network({
+    method: "GET",
+    url:
+      crypto_org_indexer +
+      `/api/v1/accounts/` +
+      addr +
+      `/transactions?pagination=offset&page=${
+        startAt + 1
+      }&limit=${PAGINATION_LIMIT}`,
```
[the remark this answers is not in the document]
thomas-nguy: fixed

src/families/crypto_org/js-broadcast.js
```diff
@@ -0,0 +1,33 @@
...
+  signedOperation: { signature, operation },
+}: {
+  account: Account,
+  signedOperation: SignedOperation,
+}): Promise<Operation> => {
+  const useTestNet = account.currency.id == TESTNET_CURRENCY_ID ? true : false;
+  const broadcastResponse = await broadcastTransaction(signature, useTestNet);
+
+  if (isBroadcastTxFailure(broadcastResponse)) {
+    throw new CryptoOrgErrorBroadcasting();
```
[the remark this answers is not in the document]
thomas-nguy: error code added in the message

src/families/crypto_org/js-signOperation.js
thomas-nguy: I think for now it is acceptable. CryptoOrg is based on cosmosSDK so it is unlikely that the implementation will diverge, if so we will create a new binding in the future

src/families/crypto_org/logic.js
thomas-nguy: fixed

src/families/crypto_org/js-synchronisation.js
thomas-nguy: unfortunately api does not take block height but index. If I can store this index somewhere in Operation, then I can optimize this loop

src/families/crypto_org/js-signOperation.js
thomas-nguy: I move it to a separate function for now since I don't feel the necessity to have a separate hw app js lib for now. 
In the future if needed we can move this function to the js lib.

src/families/crypto_org/js-buildTransaction.js
```diff
@@ -0,0 +1,61 @@
+// @flow
+import type { Transaction } from "./types";
+import type { Account } from "../../types";
+import { Units, utils } from "@crypto-com/chain-jslib";
+import { getAccountParams } from "./api/sdk";
+import { getCroSdk, TESTNET_CURRENCY_ID } from "./logic";
+
+const getTransactionAmount = (a: Account, t: Transaction) => {
+  const useTestNet = a.currency.id == TESTNET_CURRENCY_ID ? true : false;
```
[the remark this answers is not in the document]
thomas-nguy: I refactored the code in some extend, please check again

src/families/crypto_org/api/sdk.js
```diff
@@ -0,0 +1,244 @@
...
+async function getClient(useTestNet: boolean) {
+  const crypto_org_rpc_url = useTestNet
+    ? getEnv("CRYPTO_ORG_TESTNET_RPC_URL")
+    : getEnv("CRYPTO_ORG_RPC_URL");
+  if (!api) {
+    api = await getCroSdk(useTestNet).CroClient.connect(crypto_org_rpc_url);
+  }
+
+  return api;
+}
```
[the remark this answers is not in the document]
thomas-nguy: I refactored the code in some extend, please check again

src/families/crypto_org/js-getFeesForTransaction.js
```diff
@@ -0,0 +1,14 @@
+// @flow
+import { BigNumber } from "bignumber.js";
+import { FIXED_GAS_PRICE, FIXED_DEFAULT_GAS_LIMIT } from "./logic";
+
+/**
+ * Fetch the transaction fees for a transaction
+ */
+const getEstimatedFees = async (): Promise<BigNumber> => {
+  // Todo call gas station to get a more accurate tx fee in the future
```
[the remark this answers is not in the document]
thomas-nguy: fixed. It is not something that need to be fixed because it practice, it should work. The TODO is more like an optimisation if we ever have dynamic gas fee in the future

src/families/crypto_org/js-buildTransaction.js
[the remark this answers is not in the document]
thomas-nguy: (I am tracking a instances array as suggested)

ghost (approved):

PR #1187, opened 2021-05-14
LL-5324 (Network): use axios interceptors to log network

LFBarreto:
(Network): use axios interceptors to log network


src/network.js
```diff
@@ -1,11 +1,112 @@
...
+  const { url, method = "", metadata = {} } = response.config;
+  const { startTime = 0 } = metadata;
+
+  log(
+    "network-success",
+    `${response.status} ${method} ${url} (${(Date.now() - startTime).toFixed(
+      0
+    )}ms)`,
+    getEnv("DEBUG_HTTP_RESPONSE")
+      ? { data: limitLogDataSize(response.data) }
```
gre, line 52: i'm not sure why we need to do this here.
i think if DEBUG_HTTP_RESPONSE is on, it never should limit. however if it's off, we always we include the response and possibly limit. but imo we shouldn't "shrunk" it at all, we should just not include it. we shouldn't slice() the data, we will not want partial data, especially we can't automatically know it's "limited" or not.

gre (changes requested):

src/env.js
```diff
@@ -155,6 +155,11 @@ const envDefinitions = {
     parser: boolParser,
     desc: "includes HTTP response body in logs",
   },
+  DEBUG_HTTP_RESPONSE_MAX_SIZE: {
+    def: 256,
```
gre, line 159: That's too small, i think we want to get rid of the extreme payload here.
Threshold can be something like 100k

gre (changes requested):

src/network.js
gre: also are you sure this works? I think response.data can be a json object no? how would we detect the size then. (JSON.stringify to be avoided because cost)

LFBarreto: Yeah i'm not sure how to limit the size of the response logs. :thinking: to do it i did a stringify indeed to return a sliced string in case its too long but its not great.
What we can do though is maybe use something like https://www.npmjs.com/package/object-sizeof to check the response size in Bytes and just filter out the responses that are too long :shrug: 
Do we have a limitation for not wanting to keep all responses in logs?

gre: @LFBarreto i would prefer we remove out the idea to limit for now and we'll split out in a future task.

gre (approved):

gre (changes requested):
CI appears to say that stellar is broken with this rework.

gre:
@LFBarreto i'm pretty sure that your changes broke stellar as we touched the stellar interceptor which was needed for specifically this 404 issue in the past 👍

LFBarreto:
> @LFBarreto i'm pretty sure that your changes broke stellar as we touched the stellar interceptor which was needed for specifically this 404 issue in the past

yes testing locally to check what exactly causes this i still would like to keep all the logs out of the horizon file but if that causes an issue i'll revert the changes there

PR #1189, opened 2021-05-18
[WIP]Fix libcore 4.2

hzheng-ledger:
[WIP] Fix the live bot with new libcore version


src/promise.js
```diff
@@ -91,7 +91,7 @@ export async function promiseAllBatched<A, B>(
   }
 
   // initially, we schedule <batch> items in parallel
-  await Promise.all(Array(Math.min(batch, items.length)).fill().map(step));
+  await Promise.all(Array(1).fill().map(step));
```
gre, line 94: can we put back the previous code? the optimisation is important to me. typically when we have lot of inputs/outputs it's very important on mobile (that is async) to batch things.

on top of this, changing this implementation it impacting not just libcore but all work we want to 'parallelize'

gre:
oh maybe we should close this PR? it was for test?

hzheng-ledger:
yes, we can close it. Thanks

PR #1216, opened 2021-05-26
LL-4430 (Hw ConnectApp): handle dongle lock status as a retry with condition

LFBarreto:
(Hw ConnectApp): handle dongle lock status as a retry with condition

needs some extra review on the rx implem. maybe theres a better solution to the issue

### DIFF
--- a/src/hw/connectApp.js
+++ b/src/hw/connectApp.js
@@ -1,7 +1,15 @@
 // @flow
 
 import { Observable, concat, from, of, throwError, defer } from "rxjs";
-import { concatMap, map, catchError, delay } from "rxjs/operators";
+import {
+  concatMap,
+  map,
+  catchError,
+  delay,
+  filter,
+  retryWhen,
+  tap,
+} from "rxjs/operators";
 import {
   TransportStatusError,
   FirmwareOrAppUpdateRequired,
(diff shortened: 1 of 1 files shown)


src/hw/connectApp.js
```diff
@@ -222,6 +231,18 @@ const cmd = ({
...
+            errors.pipe(
+              filter(
+                (e) =>
+                  !!e &&
+                  e instanceof TransportStatusError &&
+                  e.statusCode === 0x6982 // device is locked in a special case
+              ),
+              tap(() => o.next({ type: "unresponsiveDevice" })),
+              delay(5000)
+            )
```
gre, line 244: I feel using retryWhen is dangerous here, specifically as connectApp is supposed to be one step of the bigger loop which is actions/app.js

but i miss a bit to understand the big picture here.

to me the fix should not happen here but more on the connectApp to returns something like 

```
          return of({ type: "unresponsiveDevice" });
```

at the place where it used to return "ask-open-app"

gre (changes requested):
feedback to try out. i'm not confident to do a retryWhen in this place as i also identified that we already spend too much time on connectApp which may cause the timeout in the polling mode (see discussion on slack)

src/hw/connectApp.js
LFBarreto: i tried that but it just moves the blocking issue elsewhere.
Since we unlock the device on an app opened we dont get the loop restart from a device change we just get stuck on the "unlock device" (which just replaces the "open app" from before)
The only way to restart the job is then to quit the app and then the loop restarts
maybe i haven't checked too much how other apps behave in that case :thinking:

gre:
please reboot in a future sprint / maybe schedule with a task.

PR #1221, opened 2021-05-27
LL-5329 (Bitcoin bridge): clear all lru cache used during tx at broadcast

LFBarreto:
(Bitcoin bridge): clear all lru cache used during tx at broadcast

### DIFF
--- a/src/families/bitcoin/bridge/libcore.js
+++ b/src/families/bitcoin/bridge/libcore.js
@@ -262,7 +262,7 @@ const accountBridge: AccountBridge<Transaction> = {
   sync,
   signOperation,
   broadcast: async ({ account, signedOperation }) => {
-    await calculateFees.reset();
+    calculateFees.reset();
     return broadcast({
       account,
       signedOperation,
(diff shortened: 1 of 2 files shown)


src/families/bitcoin/bridge/libcore.js
```diff
@@ -262,7 +261,13 @@ const accountBridge: AccountBridge<Transaction> = {
   receive,
   sync,
   signOperation,
-  broadcast,
+  broadcast: async ({ account, signedOperation }) => {
+    await calculateFees.reset();
```
gre, line 265: i don't think it's async. is it?

gre (changes requested):
we need to have this on all bridges that have cache. i wonder if we need to consider a generic solution where we record all LRU caches of the app and clear them all. might be overkill but it could be made "implicitly" that way.

I feel we could brainstorm with a few of us on topic. i would like more dev review on this 🙏

gre (changes requested):
to me we don't need to await reset(). apart from this it appears ok. I still miss more code review on this PR 🙏

PR #1225, opened 2021-06-01
Fixes % to be displayed on user land

gre:
also provide a better function for better perf of useNotEnoughMemoryToInstall


juan-cortes (approved):

juan-cortes:
Works for inline install on swap too 🚀

PR #1226, opened 2021-06-01
LL-5325 - add logs

machard:
https://ledgerhq.atlassian.net/browse/LL-5325

### DIFF
--- a/src/hw/actions/transaction.js
+++ b/src/hw/actions/transaction.js
@@ -14,7 +14,8 @@ import type {
   SignOperationEvent,
 } from "../../types";
 import { TransactionRefusedOnDevice } from "../../errors";
-import { getMainAccount } from "../../account";
+import { getMainAccount, formatAccount } from "../../account";
+import { formatTransaction } from "../../transaction";
 import { getAccountBridge } from "../../bridge";
 import type { ConnectAppEvent, Input as ConnectAppInput } from "../connectApp";
 import type { Action, Device } from "./types";
@@ -88,6 +89,10 @@ const reducer = (state: State, e: Event): State => {
       return { ...initialState, transactionSignError };
     }
     case "signed":
+      log(
+        "transaction-summary",
+        `✔️ has been signed! ${JSON.stringify(e.signedOperation)}`
+      );
       return { ...state, signedOperation: e.signedOperation };
     case "device-signature-requested":
       return { ...state, deviceSignatureRequested: true };
@@ -128,6 +133,18 @@ export const createAction = (
       }
 
       const bridge = getAccountBridge(mainAccount);
+
+      log(
+        "transaction-summary",
+        `→ FROM ${formatAccount(mainAccount, "basic")}`
+      );
+      log(
+        "transaction-summary",
+        `✔️ transaction ${formatTransaction(transaction, mainAccount)}`
+      );
+
+      // log("transaction-summary", `STATUS ${formatTransactionStatus(transaction, status, mainAccount)}`);
+      // status, how to get it ?
 
       const sub = bridge
         .signOperation({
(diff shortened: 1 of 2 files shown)


valpinkman (approved):

src/hw/actions/transaction.js
```diff
@@ -88,6 +89,10 @@ const reducer = (state: State, e: Event): State => {
       return { ...initialState, transactionSignError };
     }
     case "signed":
+      log(
+        "transaction-summary",
+        `✔️ has been signed! ${JSON.stringify(e.signedOperation)}`
```
gre, line 94: signed operation is not serializable. We need to pass throuh `toSignedOperationRaw` (indeed it was also probably a mistake in cli.

I suggest we should also put it in 3rd param. maybe like this:

```
log("transation-summary", "✔️ has been signed!", { signedOperation: toSignedOperationRaw(e.signedOperation) })
```

src/hw/actions/transaction.js
```diff
@@ -129,6 +134,18 @@ export const createAction = (
...
+        "transaction-summary",
+        `→ FROM ${formatAccount(mainAccount, "basic")}`
+      );
+      log(
+        "transaction-summary",
+        `✔️ transaction ${formatTransaction(transaction, mainAccount)}`
+      );
+
+      // log("transaction-summary", `STATUS ${formatTransactionStatus(transaction, status, mainAccount)}`);
+      // status, how to get it ?
```
gre, line 147: it's a good point. maybe it means we are not doing this at the correct place. but now where would that be🤔

on LLD/LLM side we indeed will use `bridge.broadcast` which is implementation dependant so we don't want to do it N time for each family.

However, i wonder if we could always force us to use a higher level "broadcast". which we already do here: 
https://github.com/LedgerHQ/ledger-live-desktop/blob/766fdf662b78d1d9824367687a5ffe725feef730/src/renderer/hooks/useBroadcast.js#L21

indeed we also need to handle DISABLE_TRANSACTION_BROADCAST which is a generic thing too.

it sounds like we could generalize it back into live common and have the logs in this place. (also the signed op log 🤔  )

so to me it sounds like this broadcast is to be in live-common.

I just know that platform started doing a tiny differ version of it, so maybe we also need to have this in scope (i think we need the non hook version in live-common at least).

src/transaction/signOperation.js
```diff
@@ -59,11 +65,17 @@ export const fromSignOperationEventRaw = (
...
   e: SignOperationEvent
 ): SignOperationEventRaw => {
+  let signedOperation;
   switch (e.type) {
     case "signed":
+      signedOperation = e.signedOperation;
+      log(
+        "transaction-summary",
+        `✔️ has been signed! ${JSON.stringify(e.signedOperation)}`
+      );
```
gre, line 75: not a good idea to have log in a simple serialise function, it shouldn't do a side effect because we may use it for N usecases, plus on mobile we actually don't even use this i think. (as we don't need serialization)

src/transaction/signOperation.js
```diff
@@ -45,11 +45,17 @@ export const fromSignOperationEventRaw = (
...
   accountId: string
 ): SignOperationEvent => {
+  let signedOperation;
   switch (e.type) {
     case "signed":
+      signedOperation = fromSignedOperationRaw(e.signedOperation, accountId);
+      log(
+        "transaction-summary",
+        `✔️ has been signed! ${JSON.stringify(e.signedOperation)}`
+      );
```
gre, line 55: see comment below too.

gre (changes requested):

gre:
we'll have to handle this differently. closing the work for now so we can focus on other prios.

PR #1229, opened 2021-06-03
LL-4979 (Errors): add specific platform flow errors for use in LLD and LLM

LFBarreto:
(Errors): add specific platform flow errors for use in LLD and LLM

### DIFF
--- a/src/errors.js
+++ b/src/errors.js
@@ -196,3 +196,34 @@ export const FreshAddressIndexInvalid = createCustomErrorClass(
 );
 
 export * from "./families/polkadot/errors";
+
+/** Platform request account flow canceled */
+export const PlatformRequestAccountCanceled = createCustomErrorClass(
+  "PlatformRequestAccountCanceled"
+);
+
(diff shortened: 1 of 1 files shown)


rjarasson-ledger:
IMHO, these errors does not fit our needs.
First because there is a pending spec for the JSONRPC errors (naming included) that is still being written.& validated
Second, because we will need to add a "code" integer for those, and make the JSONRPC server being able to catch and serialize them correctly (this is not the case), so that the SDK we are working on, will also be able to cast those errors, back to TS types.

For now, we chose not to type the platform errors, because on the SDK end, we do not have anything to interpret them anyway (non of the provider even give visual feedbacks to retrieved errors).
So, `new Error("Something is not ok")` is more for logs than humans currently.

gre:
thanks for the work but per Rémi's comment let's wait for a spec on this then. I close the PR until we can have a clear spec on this and because TypeScript is going to conflict with this and we don't need to maintain this until we know where we go.

PR #1238, opened 2021-06-09
Ledgerjs ts bump

gre:
https://ledgerhq.atlassian.net/browse/LL-3928


gre:
this PR was here to prove that all seems good with ledgerjs ts. it did prove it. indeed test fails but it's because ongoing issue on algo

PR #1244, opened 2021-06-10
LL-5866 Remove payoutNetworkFees from toAmount value to match binaryPayload data

juan-cortes:
The estimated amount to receive contained inside the binary payload no longer includes the `payoutNetworkFees` so we don't have to break down the amount on the confirmation/summary screens anymore. 

This is only available currently on the staging swap API but will be rolled to production.
If there's a gap of time where this code is in prod but the backend is not aligned, the users will see a different amount between the device and the summary (device being higher because it includes the fees) don't know how to proceed in merging this + the LLD/LLM prs + putting it all in prod. Needs to be planned properly.

### DIFF
--- a/src/exchange/swap/getExchangeRates.js
+++ b/src/exchange/swap/getExchangeRates.js
@@ -42,7 +42,8 @@ const getExchangeRates: GetExchangeRates = async (
 
   return res.data.map(
     ({
-      rate,
+      rate: maybeRate,
+      payoutNetworkFees: maybePayoutNetworkFees,
       rateId,
       provider,
       amountFrom,
@@ -50,7 +51,6 @@ const getExchangeRates: GetExchangeRates = async (
       minAmountFrom,
       maxAmountFrom,
       tradeMethod,
-      payoutNetworkFees,
     }) => {
       let error;
       let magnitudeAwareRate;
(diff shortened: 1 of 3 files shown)


gre:
> This is only available currently on the staging swap API but will be rolled to production.

what does it mean for:
- the mergeability of this PR?
- the backward compatiblity of this new code with whatever we have in production? (aka no impact if i bump live-common right now on develop of LLD/LLM?)

juan-cortes:
That was my point, if we bump this on LLD/LLM it will not **break** anything as in it will not cause a crash, but it depends on the LLD/LLM prs being merged too and the changes to the swap API being rolled to production to show the right amount on the summary/confirmation screens on swap floating rate. 

- If this live-common goes gets bumped and the LLD pr is not merged, the confirmation amount will be incorrect
- If the LLD pr gets merged but this is not yet bumped, the confirmation amount will be incorrect
- If both are merged but backend is not rolled into production yet, the confirmation amount will be incorrect

**Only** affecting floating rates, but it's the default so basically all swaps will be showing the wrong amount until everything is aligned. I added the notice specifically so that we have that in mind. Also added a warning on the LLD pr

src/exchange/swap/getExchangeRates.js
```diff
@@ -84,26 +84,29 @@ const getExchangeRates: GetExchangeRates = async (
       } else {
         // NB Allows us to simply multiply satoshi values from/to
         magnitudeAwareRate = (tradeMethod === "fixed"
-          ? BigNumber(rate)
+          ? BigNumber(maybeRate)
```
rjarasson-ledger, line 87: Since it's a maybe, shouldn't we fallback to 0 ?

rjarasson-ledger (changes requested):

src/exchange/swap/getExchangeRates.js
juan-cortes: I couldn't find a better name, if it's "fixed", as used in this ternary it won't be a maybe since all fixed rates have a `rate`, line 100 fallback to the calculation so it's never undefined when accessed if that makes sense

rjarasson-ledger: Yeah but isn't it problematic to have magnitudeAwareRate (which is returned, i don't know how it's used after) being a NaN bignumber ( BigNumber(undefined) ) ?

juan-cortes: `magnitudeAwareRate` will never be `BigNumber(undefined)`. If the `tradeMethod` is fixed, the rate exists.

rjarasson-ledger: mmh ok, did not know the specifics of this part of the code.
LGTM then

rjarasson-ledger (approved):
Should wait for the API to be updated right ?

juan-cortes:
> Should wait for the API to be updated right ?

I guess this is a call for core team cc @gre the order in which these get merged/bumped
It's available on staging v2 and I think it should not impact users on older versions, even if it goes into prod.
The only thing we need is to align this PR with the LLD and LLM ones 🤔

gre:
I'm still confused by your first comment @juan-cortes 

>  if we bump this on LLD/LLM it will not break anything

and

> If this live-common goes gets bumped and the LLD pr is not merged, the confirmation amount will be incorrect

are opposite :)

if it's not a noop, then it's a "breaking change", if it's a breaking change then we can't merge this work easily but to organize a careful release, which in context of this feature is hard because we have no clue when swap will be merged on LLD/LLM.

So if we can make it so it's a noop it would be a big plus, otherwise it's a pain to maintain.

as for the backend, we expect it to be in production ready at the time we merge LLD/LLM but not blocker to merge in live-common if it's not "active" yet.

juan-cortes:
I think we need to talk about this

gre:
we'll take this and make sure to include the minimal work on LLD/M side. thanks @juan-cortes

PR #1252, opened 2021-06-11
Improved model detection for proxy transport

juan-cortes:
With this new logic, we are both able to automatically detect the device model based on the target id and override that detection with a new env in case we want to test specific models on the manager via the env `DEVICE_PROXY_MODEL` this means that when using a proxy connection on LLM we will be able to see the correct information on the manager allowing us to install applications and see the remaining storage on the device instead of falling back to a nano s with 0 storage.

### Before this PR / After this PR (added lines for emphasis)
This is using the exact same device, with the same installed applications.
image

### How to test
- Yalc this or on a LLM setup
- Connect a nano s via proxy
- Launch LLM and choose that device while accessing the manager
- Ensure the data and storage is correct
- Repeat with a nano X.

### DIFF
--- a/src/apps/hw.js
+++ b/src/apps/hw.js
@@ -1,6 +1,7 @@
 // @flow
 
 import Transport from "@ledgerhq/hw-transport";
+// $FlowExpectedError Caused by TS migration, flow def support stopped at v5.
 import { getDeviceModel, identifyTargetId } from "@ledgerhq/devices";
 import { UnexpectedBootloader } from "@ledgerhq/errors";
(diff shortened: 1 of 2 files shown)


JunichiSugiura (commented):
LGTM. Flow error on CI is expected during TS transition period.

gre:
maybe we can $FlowFixMe on the import tho.

juan-cortes:
> maybe we can $FlowFixMe on the import tho.

Addressed with f2aa5a6789213f4971ca3572f91e2db50d20b5c4

gre (approved):
let's include this work.

PR #1256, opened 2021-06-14
Fixes Ripple to forward error in some cases

gre:
crash logs shows that sometimes we had

`undefined is not an object (evaluating 'e.data.error')`


juan-cortes (approved):
lgtm

PR #1262, opened 2021-06-15
disable algorand tests

gre:
https://ledgerhq.atlassian.net/browse/LL-6006


gre:
snapshot will have to be regenerated in a second commit

PR #1265, opened 2021-06-16
LL-6016 (ServiceStatusProvider): add currencies filter

LFBarreto:
(ServiceStatusProvider): add filtered context hooks
this should not break the current work since it adds a new hook and doesnt modify the existing.

We use components name in order to filter out status incidents by ticker in order to not colide between values such as `Bitcoin` or `Bitcoin cash`.


valpinkman (dismissed):

IAmMorrow (changes requested):
For me this won't be enough, this assume that incident names are well written and formatted to match the regex. It also assume that every incident is matched with a currency while this is not always the case.

To solve this we need either 
- a mapping between infra service ids and topics (a currency can be a topic). 
- specific meta data to be added on servicestatus API side (if it is even doable).

gre:
please come back when we have a more viable solution with infra

PR #1266, opened 2021-06-16
LL-5877 - feat(Platform): add CatalogProvider with static manifest

rjarasson-ledger:
This is an implementation of a CatalogProvider for the Platform, which holds multiple App manifests for defining apps available in the catalog.

You can override the manifest by using the `PLATFORM_MANIFEST_PATH` to point to a local JSON file.

JIRA ticket: https://ledgerhq.atlassian.net/browse/LL-5977

# TODO:
* [x] Catalog Provider
* [ ] Upload icons to CDN (https://github.com/LedgerHQ/ledger-live-assets/pull/66)
* [x] Add helpers for generating app derivatives (like translatable contents and URL)
* [ ] MAYBE - Fetch manifest remotely
* [ ] MAYBE - versions check
* [ ] Testing (unit testing mainly)

FUTURE:
*  implement domain restriction and permission checking
*  Catalog Provider reducer with additional data (filtering/sorting/grouping for persisting UX)


rjarasson-ledger:
This can't be done like that for the manifest...
`fs` module is not available in the mobile, and it makes no sense.

rjarasson-ledger:
I added the same logic as for the announcement api, to enable fetching manifest from remote.
But by default, the manifest is the mock, so that we still have a static catalog, but with the possibility of overriding it.

src/platform/CatalogProvider.js
```diff
@@ -0,0 +1,80 @@
...
+  apps: [],
+};
+
+export const PlatformCatalogContext = React.createContext<State>(initialState);
+
+const PlatformCatalogProvider = ({ children }: Props) => {
+  const [state, setState] = useState(initialState);
+
+  useEffect(() => {
+    api.fetchManifest().then((manifest) => setState({ apps: manifest }));
```
gre, line 28: so the code to be dynamic is already ready? nice

src/platform/api/api.js
```diff
@@ -0,0 +1,30 @@
...
+import type { AppManifest, PlatformApi } from "../types";
+
+// expose a function to fetch data from the cdn (data from ledger-live-assets)
+// https://cdn.live.ledger.com/
+
+const basePlatformUrl = () => getEnv("PLATFORM_API_URL");
+const platformVersion = () => getEnv("PLATFORM_API_VERSION");
+
+async function fetchManifest(): Promise<AppManifest[]> {
+  const url = `${basePlatformUrl()}/v${platformVersion()}/data.json?t=${Date.now()}`;
```
gre, line 14: is there any reason we pass the time here? It doesn't make a lot of sense to me and it bypass CDN capability to cache which may not be a good idea.

gre (commented):
nice job. it's good you prepared the code for future, would help QA while we can enable it.

gre:
I need more review and check from @IAmMorrow  on this but let's try to merge this tomorrow morning to gather all the work for live-common requirements.

src/platform/api/api.js
rjarasson-ledger: This was inspired by the Announcement api, so i thought there were reasons to do so.
I agree that it may bypass CDN's cache, i think it's intended to only bypass a local xhr cache.

src/platform/api/api.mock.js
```diff
@@ -0,0 +1,217 @@
+// @flow
+import type { AppManifest, PlatformApi } from "../types";
+
+const manifest: AppManifest[] = [
+  {
+    id: "paraswap",
+    name: "ParaSwap",
+    url:
+      "https://ledger-live-platform-apps.vercel.app/app/dapp-browser?url=https%3A%2F%2Fparaswap.io%2F%3Fembed%3Dtrue%26referrer%3Dledger1&nanoApp=Paraswap&dappName=paraswap",
```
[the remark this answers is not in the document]
rjarasson-ledger: This must be ledger2 now, to be fixed

src/platform/CatalogProvider.js
```diff
@@ -0,0 +1,80 @@
...
+  }, []);
+
+  return (
+    <PlatformCatalogContext.Provider value={state}>
+      {children}
+    </PlatformCatalogContext.Provider>
+  );
+};
+
+export const useCatalog = (
```
IAmMorrow, line 38: We need a way to "get manifests from all branches" at once or else we will lose ordering. Product insisted they want to fine tune ordering at the beginning (before using more advanced ordering later).

IAmMorrow (changes requested):

IAmMorrow (approved):

gre:
I see you merged master in, beware it means you will now need to be in sync with LLD/LLM side, meaning we have a dependency on the live-common 20 merge.
We had some tests issue (bad merge from my end on the 'jest' fixes) which i should have restored on master, which i why i merged again.

PR #1269, opened 2021-06-18
ledgerjs@rc version with TypeScript migration work and latest erc20

gre:


gre:
for monday: ledgerjs seems to have wrong deps and this creates lib dup issue...

gre:
Due to the heavy changes of this PR, it is better to HODL it until we prioritize on the bigger priority like #1166

gre:
CI failed for tmp issue but all is good on this.

PR #1270, opened 2021-06-21
remove env EXPERIMENTAL_INLINE_INSTALL to enable for all

gre:
https://ledgerhq.atlassian.net/browse/LL-5462 Activate in-app device flow as non-experimental

dev impact: will be part of next live-common release and will need also to be removed from the experimental on LLD/LLM.
qa impact: app will be like if you had the experimental before but the experimental will just disappear as it will be for all users now.

### DIFF
--- a/src/hw/connectApp.js
+++ b/src/hw/connectApp.js
@@ -24,7 +24,7 @@ import getAddress from "./getAddress";
 import openApp from "./openApp";
 import quitApp from "./quitApp";
 import { mustUpgrade } from "../apps";
-import { getEnv } from "../env";
+
 export type RequiresDerivation = {|
   currencyId: string,
   path: string,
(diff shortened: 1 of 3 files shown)


src/hw/connectApp.js
```diff
@@ -76,18 +76,12 @@ export const openAppFromDashboard = (
           switch (e.statusCode) {
             case 0x6984:
             case 0x6807:
-              return getEnv("EXPERIMENTAL_INLINE_INSTALL")
```
juan-cortes, line 79: drop the import

juan-cortes (changes requested):
We'd still need to remove it from the exp settings screen on LLD/LLM right?

gre:
@juan-cortes yes as mentioned in description :)

juan-cortes (approved):

PR #1279, opened 2021-06-24
update polkadot libraries

gre:


gre:
we're going to test this in next release in any case.

rjarasson-ledger:
We also should change the app version required for polkadot: `7.9050.0`

PR #1280, opened 2021-06-28
LL-6072 LL-6161 [Polkadot] Dynamically fetch minimum bond value

ghost:
Context: on Polkadot the minimum amount allowed to bond can now change at any time with a governance decision. We need to fetch it dynamically to adapt the warnings and errors we display to the users.

### DIFF
--- a/src/families/polkadot/preload.js
+++ b/src/families/polkadot/preload.js
@@ -12,7 +12,7 @@ import { getStakingProgress, getValidators } from "./validators";
 let currentPolkadotPreloadedData: PolkadotPreloadData = {
   validators: [],
   staking: null,
-  minimumBondBalance: BigNumber(0),
+  minimumBondBalance: "0",
 };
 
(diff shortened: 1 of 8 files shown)


gre (approved):

src/families/polkadot/preload.js
```diff
@@ -56,11 +58,15 @@ function fromHydratePreloadData(data: mixed): PolkadotPreloadData {
         bondingDuration: Number(bondingDuration) || 28,
       };
     }
+
+    // $FlowFixMe
+    minimumBondBalance = BigNumber(data.minimumBondBalance);
```
gre, line 63: i think you should do 

```
BigNumber(data.minimumBondBalance || 0)
```

because there is a slight possibility that you will first get hydrate() WITHOUT the data when coming on old Live.

gre (changes requested):
one slight change to do

PR #1286, opened 2021-07-01
LL-6099 - Enable remotely loaded catalog

IAmMorrow:


juan-cortes (approved):
lgtm

gre:
closing until we figure out if we want this.
also we would need at minimum a test that prove that fetching the api returns "something". like a test that verify there are at least one result is needed.

PR #1294, opened 2021-07-07
LL-6294 Allow connectApp to require latest fw version

juan-cortes:
In order to use this, just modify the input of the `connectApp` request to include the flag. For instance:

```
<DeviceAction
  action={action}
  request={{ account: mainAccount, tokenCurrency, requireLatestFirmware: true }}
  onResult={() => transitionTo("receive")}
/>
```
And it will be the first thing it checks when we connect a device, throwing an error if needed, or letting pass if OK.

### DIFF
--- a/src/hw/connectApp.js
+++ b/src/hw/connectApp.js
@@ -230,7 +230,11 @@ const cmd = ({
             }
 
             // in order to check the fw version, install deps, we need dashboard
-            if (dependencies?.length || appAndVersion.name !== appName) {
+            if (
+              dependencies?.length ||
+              requireLatestFirmware ||
+              appAndVersion.name !== appName
+            ) {
               return attemptToQuitApp(transport, appAndVersion);
             }
 
--- a/src/hw/connectApp.js
+++ b/src/hw/connectApp.js
@@ -1,7 +1,8 @@
 // @flow
 
+import semver from "semver";
 import { Observable, concat, from, of, throwError, defer } from "rxjs";
-import { concatMap, map, catchError, delay } from "rxjs/operators";
+import { mergeMap, concatMap, map, catchError, delay } from "rxjs/operators";
 import {
   TransportStatusError,
   FirmwareOrAppUpdateRequired,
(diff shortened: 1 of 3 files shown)


src/hw/connectApp.js
```diff
@@ -160,19 +167,53 @@ const cmd = ({
...
+        dependencies,
+        requireLatestFirmware,
+      }: any) =>
         defer(() => from(getAppAndVersion(transport))).pipe(
           concatMap((appAndVersion): Observable<ConnectAppEvent> => {
             timeoutSub.unsubscribe();
 
             if (isDashboardName(appAndVersion.name)) {
+              // check if we meet minimum fw
+              if (requireLatestFirmware) {
```
gre, line 189: if user is not in dashboard in the first place, then we will never check this. is that expected? are we sure to be doing this in the correct flow?

I imagine this only work in context of also using dependencies, right? but we can't use this reliably if we just ask an app , right?

It's good for me, but please doc this limitation because that could bitten us later.

src/hw/connectApp.js
```diff
@@ -160,19 +167,53 @@ const cmd = ({
...
+                return from(getDeviceInfo(transport)).pipe(
+                  mergeMap((deviceInfo: DeviceInfo) =>
+                    from(manager.getLatestFirmwareForDevice(deviceInfo)).pipe(
+                      mergeMap((latest: ?FirmwareUpdateContext) => {
+                        if (
+                          !latest || // Fixme is it safe to let it pass here?
```
gre, line 195: you can remove this comment => yes it is, it means that it's up to date.

src/hw/connectApp.js
```diff
@@ -160,19 +167,53 @@ const cmd = ({
...
+                          semver.eq(deviceInfo.version, latest.final.version)
+                        ) {
+                          o.next({ type: "latest-firmware-resolved" });
+                          return innerSub({ appName }); // NB without the fw version check
```
gre, line 199: not understanding the comment.

src/hw/connectApp.js
juan-cortes: Done

src/hw/connectApp.js
juan-cortes: We return the innerSub removing the firmware requirement, just like we did for the check dependencies, no sure if it's the cleanest approach but it's what we came up with on the peer programming session last time 🦐

src/hw/connectApp.js
juan-cortes: You are correct, we need to add the same quitapp as for the dependencies, added the check on https://github.com/LedgerHQ/ledger-live-common/pull/1294/commits/649506c7bb410b35fe84cdbbae427fa300c97cc1

PR #1295, opened 2021-07-07
externalize the versioning of platform

gre:
usecase:
as LLD/LLM are implementers of the platform, they are the one to provide the version, it's not a constant
it allows a true loosely coupling of these parts

**impact:**

LLD/LLM must adapt this:

when there are code that do `setSupportedCurrencies` you also must add `setPlatformVersion`


valpinkman (approved):

src/walletconnect/walletconnect.test.js
```diff
@@ -9,6 +9,9 @@ import { getCryptoCurrencyById, setSupportedCurrencies } from "../currencies";
 import type { Account } from "../types/account";
 import { emptyHistoryCache } from "../account";
 import { setEnv } from "../env";
+import { setPlatformVersion } from "../platform/version";
+
+setPlatformVersion("0.0.1");
```
IAmMorrow, line 14: walletconnect and platform are not related 🤔 

They could be factorized in the future 👍

IAmMorrow (approved):

PR #1296, opened 2021-07-07
Makes v3 the default, merge aligned with LLD/LLM prs

juan-cortes:
Both https://github.com/LedgerHQ/ledger-live-desktop/pull/3854 and https://github.com/LedgerHQ/ledger-live-mobile/pull/1745 need to depend on a live-common version including this switch before they can be merged.

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -416,7 +416,7 @@ const envDefinitions = {
     desc: "dev flag to skip onboarding flow",
   },
   SWAP_API_BASE: {
-    def: "https://swap.ledger.com/v2",
+    def: "https://swap.ledger.com/v3",
     parser: stringParser,
     desc: "Swap API base",
   },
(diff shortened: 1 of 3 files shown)


gre:
am i right to understand we want to merge this only when it's a GO? then i think we can tmp close the PR @juan-cortes

juan-cortes:
> am i right to understand we want to merge this only when it's a GO? then i think we can tmp close the PR @juan-cortes

Sure, close if it's clearer for you 🤷🏼

valpinkman (approved):

gre:
so when we have a go to wyre feature, we merge it on desktop/mobile with this? @juan-cortes

juan-cortes:
> so when we have a go to wyre feature, we merge it on desktop/mobile with this? @juan-cortes

That is correct, we need this in a live-common version so that it's v3 prod by default. I also had to add a fix for a small bug that triggered the open app twice when we required the latest firmware, it didn't break the flow but it required an extra ok from the user, which is not great. Sorry about that

PR #1299, opened 2021-07-08
LL-6138 Fix accounts ordering algorithm

imqdee:
Until now, the sort algorithm didn't take into account numeric values, so 11 was a lower value than 2.
By passing the numeric option to the localeCompare function, numeric values are now taken into account. 11 is now a higher value than 2 as expected.

### Type

Bug fix

### Context

[LL-6138]

### Parts of the app affected / Test plan

Once a new version of this package will be published and bumped from LLD/LLM, UIs from these projects will be fixed.

[LL-6138]: https://ledgerhq.atlassian.net/browse/LL-6138


LFBarreto (approved):
:rocket:

jgiamporcaro-ledger:
I don't understand how my tiny change drops the coverage by ~4%, the result shared by the bot seems to be unrelated to my change.
Is it something I have to take into account? (@LFBarreto)

PR #1301, opened 2021-07-12
Platform app provider v2

IAmMorrow:
Added PlatformAppProvider that looks more like our usual data providers
Added an option to push "extra manifests" (useful for test builds)
Added update lifecycle (every 1800 secs)


src/platform/CatalogProvider.js
```diff
@@ -1,92 +0,0 @@
-// @flow
```
gre, line 1: Removing a file, if you confirm it's used in current LLD/LLM side is to be avoided, if you work on a v2 you must keep the file, have a console.warn that deprecates it if you want, but keep it until we port the LLD/LLM side.

otherwise it's a breaking changes and it must be avoided as much as possible.

src/platform/PlatformAppProvider/api/api.js
```diff
@@ -1,8 +1,8 @@
 // @flow
```
gre, line 1: moving a file is the same problem. please don't.

gre (changes requested):
we must not do breaking changes in live-common unless justified. if you work on a "v2" you may do a "v2" folder instead.

src/platform/api/api.mock.js
```diff
@@ -99,10 +99,10 @@ const manifest: AppManifest[] = [
     currencies: ["ethereum"],
     content: {
       shortDescription: {
-        en: "The smart way to manage your DeFi portfolio.",
+        en: "Manage your entire Ethereum DeFi portfolio from one place.",
       },
       description: {
-        en: "The smart way to manage your DeFi portfolio.",
+        en: "Manage your entire Ethereum DeFi portfolio from one place.",
```
gre, line 105: accidental revert.

PR #1305, opened 2021-07-13
LL-6400 Updated domains to fix mobile issue

IAmMorrow:


gre (approved):
why is the api.mock.js on two parts? shouldn't one import the other?

IAmMorrow:
Because i'm not thinking :pls:

PR #1308, opened 2021-07-16
LL-5794 -Manager - "Apps installed" tab: Apps order should be corresponding to the install queue

jelbaz-ledger:
## Context (issues, jira)

[LL-5794]

## Description / Usage

In the LLD/LLM `Manager` / `Installed Apps` tab, apps with status installing will now be listed by installation order with the next installation target at the top.

<details>
  <summary>Screens 📸 </summary>

Simulator Screen Shot - iPhone 11 - 2021-07-16 at 14 05 08
<br/><hr />
Capture d’écran 2021-07-16 à 11 32 47

</details>

## Expectations

[LL-5794]: https://ledgerhq.atlassian.net/browse/LL-5794

### DIFF
--- a/src/apps/react.js
+++ b/src/apps/react.js
@@ -110,19 +110,15 @@ export function useAppsSections(
     { type: "default", order: "asc" }
   );
 
-  const device = installedAppList
-    .sort(({ name: _name }, { name }) =>
-      installed.indexOf(_name) > installed.indexOf(name) ? -1 : 0
-    )
-    .sort(({ name: _name }, { name }) => {
-      // place install queue on top of list
-      // with the app being installed at the top
-      let pos1 = installQueue.indexOf(_name);
-      let pos2 = installQueue.indexOf(name);
-      pos1 = pos1 < 0 ? Number.MAX_VALUE : pos1;
-      pos2 = pos2 < 0 ? Number.MAX_VALUE : pos2;
-      return pos1 - pos2;
-    });
+  const device = installedAppList.sort(({ name: _name }, { name }) => {
+    // place install queue on top of list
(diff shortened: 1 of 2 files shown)


valpinkman (approved):

src/apps/react.js
```diff
@@ -114,10 +114,15 @@ export function useAppsSections(
...
-    );
+    .sort(({ name: _name }, { name }) => {
+      // place install queue on top of list
+      // with the app being installed at the top
+      let pos1 = installQueue.indexOf(_name);
+      let pos2 = installQueue.indexOf(name);
+      pos1 = pos1 < 0 ? Number.MAX_VALUE : pos1;
+      pos2 = pos2 < 0 ? Number.MAX_VALUE : pos2;
+      return pos1 - pos2;
+    });
```
gre, line 125: is there a way to rework this sort to avoid the "double sort()" chain? I feel this part is not really well written today / possibly have perf issue on the complexity.

asking this now because maybe you have a hint on this part as you digged it.

gre (changes requested):
please try to add a test in https://github.com/LedgerHQ/ledger-live-common/blob/master/src/apps/react.test.js

src/apps/react.js
jelbaz-ledger: > is there a way to rework this sort to avoid the "double sort()" chain? I feel this part is not really well written today / possibly have perf issue on the complexity.

I removed the first .sort which I think was not needed 👍. (`installed` seems to be an object[] but was assumed to be an array of strings in the comparison function).

gre (approved):
great!

PR #1309, opened 2021-07-19
add displayName to App

valpinkman:
## Context (issues, jira)

[LL-6299]

## Description / Usage

Add a new key `displayName` in the App type used by Manager/AppList in LLD-LLM

## Expectations

[LL-6299]: https://ledgerhq.atlassian.net/browse/LL-6299

### DIFF
--- a/src/notifications/AnnouncementProvider/index.js
+++ b/src/notifications/AnnouncementProvider/index.js
@@ -85,7 +85,7 @@ export const AnnouncementProvider = ({
         updateTime: Date.now(),
       };
     },
-    [context, onNewAnnouncement]
+    [context, onNewAnnouncement, fetchApi]
   );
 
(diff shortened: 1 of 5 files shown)


src/notifications/AnnouncementProvider/index.js
```diff
@@ -85,7 +85,7 @@ export const AnnouncementProvider = ({
         updateTime: Date.now(),
       };
     },
-    [context, onNewAnnouncement]
+    [context, onNewAnnouncement, fetchApi]
```
gre, line 88: why is this in this PR?

valpinkman: becuase you asked me to fix the lint iissues

gre: it's ok

PR #1317, opened 2021-07-20
LL-6300 - Experimental - Add swap v2 env variable

imqdee:
That will allow us to work on the next version of the swap.

## Context (issues, jira)

[LL-6300]

## Expectations

Changes made here don't need any tests because I only added an item to the list (tbh, lots of functions and logic are located in the modified time and that could be a good idea to one day write some tests here).
This PR has no impact. No errors/warnings will be triggered if these changes are deployed without any change from LLD/LLM.

[LL-6300]: https://ledgerhq.atlassian.net/browse/LL-6300

### DIFF
--- a/src/env.js
+++ b/src/env.js
@@ -264,6 +264,11 @@ const envDefinitions = {
     parser: boolParser,
     desc: "enable an experimental implementation of USB support",
   },
+  EXPERIMENTAL_SWAP: {
+    def: false,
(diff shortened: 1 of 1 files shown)


juan-cortes:
Looks good to me, does LLD still work without adding this here? 🤔

gre:
yeah no impact when adding envs.

PR #1324, opened 2021-07-23
added final lido manifest

IAmMorrow:
## Context (issues, jira)

LL-6376

## Description / Usage

Added Lido manifest data

## Expectations


gre (changes requested):
lint issues

PR #1325, opened 2021-07-27
LL-6521  - Swap providers fetching

imqdee:
Create custom hooks to fetch, filter and sort swap providers.

## Context (issues, jira)

Main task [LL-6300]
Subtask [LL-6521]

## Description / Usage

Users only have to use the custom hook from one of his component.
Example

```javascript
const Swap = () => {
  const { providers, isLoading, error } = useSwapProviders();
  
  if (error) return <p>{error.message}</p>
  if (isLoading) return <p>loading...</p>

  return <ul>{providers.map(({provider}) => <li key={provider}>{provider}</li>)}</ul>;
};
```  

## Expectations

[LL-6300]: https://ledgerhq.atlassian.net/browse/LL-6300
[LL-6521]: https://ledgerhq.atlassian.net/browse/LL-6521

### DIFF
--- a/src/__tests__/exchange/swap/hooks.test.js
+++ b/src/__tests__/exchange/swap/hooks.test.js
@@ -1,0 +1,114 @@
+// @flow
+import { renderHook } from "@testing-library/react-hooks";
+import { useSwapProviders, initialState } from "../../../exchange/swap/hooks";
+import * as mocks from "../../../exchange/swap/mock";
+import { setEnv } from "../../../env";
+
+describe("exchange/swap/hooks", () => {
+  const OLD_ENV = process.env;
+
+  beforeEach(() => {
+    jest.resetModules();
+    jest.clearAllMocks();
+
+    // LLC internal variable
+    setEnv("MOCK", "true");
+
+    // All env variables including ones setted by LLM/LLD
+    process.env = { ...OLD_ENV };
+  });
+
+  afterAll(() => {
+    // Restore old environment
+    process.env = OLD_ENV;
+  });
+
+  test("fetches providers on mount", async () => {
+    const spy = jest.spyOn(mocks, "mockGetProviders");
+    renderHook(() => useSwapProviders());
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+
+  test("returns initial state on mount", async () => {
+    const { result } = renderHook(() => useSwapProviders());
+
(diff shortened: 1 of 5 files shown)


src/exchange/swap/sortProvidersByWeight.js
```diff
@@ -0,0 +1,27 @@
+// @flow
+import type { AvailableProvider } from "./types";
+
+const providerPriorityMap = { changelly: 1 };
```
jgiamporcaro-ledger, line 4: Wanted to move this map to a JSON file but apparently *.json files aren't included in the post-build bundle. Tell me if I missed something, will be happy to move this data to allow isolate updating. 

Plus we could imagine moving these priorities from here to the CMS, It's a simple idea.

src/__tests__/exchange/swap/hooks.test.js
```diff
@@ -0,0 +1,114 @@
...
+    await waitForNextUpdate();
+
+    expect(result.current.isLoading).toBe(false);
+    expect(result.current.error).toBe(null);
+    expect(result.current.providers.length).toBe(2);
+    expect(spy).toHaveBeenCalledTimes(1);
+  });
+
+  test("returns sorted providers", async () => {
+    const providers = ["bitfinex", "kraken", "bitstamp", "changelly", "wyre"];
```
jgiamporcaro-ledger, line 51: That would be awesome in the future to create some factories by using libraries like Fishery and faker. That will allow us to write tests faster and increase our confidence in them. I'm voluntary if needed.

src/exchange/swap/providersPriority.js
```diff
@@ -0,0 +1,3 @@
+const providersPriorityMap = { changelly: 1 };
```
jgiamporcaro-ledger, line 1: Wanted to move this map to a JSON file but apparently *.json files aren't included in the post-build bundle. Tell me if I missed something, will be happy to move this data to a better format file.

Plus we could imagine moving these priorities from here to the CMS, It's a simple idea.

juan-cortes: I don't think we need priority but rather just rely on the rates themselves, in any case, this mapping is fine by me in case we want to prioritize one over another

jgiamporcaro-ledger:
Will be moved to LLD in waiting that swap work is finished from this side.

PR #1327, opened 2021-07-28
[FIX] remove duplicated src/platform/api folder

ghost:
Remove duplicated `src/platform/api` folder which is a duplicate of `src/platform/PlatformAppProvider/api`

The duplicate might be the other way around (i.e `platform/PlatformAppProvider/api` is a duplicate of `src/platform/api`) but I doubt it since `PlatformAppProvider` is more recent.

## Context (issues, jira)

N/A

## Description / Usage

Should not affect user land nor any part of the code since it remove what seems to be duplicated dead code

## Expectations

It has no impact since it's just a curation of duplicated file with an internal import update. No "api contract" are updated. It is therefore invisible on the user (i.e any repo using `ledger-live-common`) side.

Integration tested on LLD ✅


gre:
@achabrolin-ledger Can you resolve conflict?

I believe the duplicate was here to not be a breaking changes but now is acceptible time to merge this if not used as both LLD / LLM are now in sync. is that correct @IAmMorrow ?

achabrolin-ledger:
> @achabrolin-ledger Can you resolve conflict?

Done, in a nutshell this PR removes deprecated / unused `src/platform/api` folder and `src/platform/CatalogProvider.tsx` component.
LLD and LLM use `src/platform/PlatformAppProvider` component and API.

> I believe the duplicate was here to not be a breaking changes but now is acceptible time to merge this if not used as both LLD / LLM are now in sync. is that correct @IAmMorrow ?

@gre 
Food for thoughts regarding how we handle duplication (to avoid breaking change) and api migration: it could wise, when we do an update to the API (like here, a `src/platform/PlatformAppProvider` folder and api has been created to replace `src/platform/api`) to do a draft PR removing the deprecated part of the API. This draft would then be merged a few releases later to officially remove deprecated parts of the API once relevant / acceptable. 

To illustrate what I mean, lets use this platform api example:

The original API is `src/platform/api`. One day we do un upgrade (breaking change) of this API, and create a new (duplicated) version in `src/platform/PlatformAppProvider`.
To do so without breaking change, we keep the original one create a PR with new version in `src/platform/PlatformAppProvider`.
At the same time we create this PR, we could create a draft PR removing the original (now deprecated) version located in `src/platform/api` (so in this example, this draft PR would have the exact content of this PR we are discussing right here).

This allow us to keep track of the fact that this is now deprecated, protect us against forgetting to remove this deprecated code and avoid any headache later one when trying to figure out why we have duplicated code and which version is the used one.  

Hope my explanation is clear 😅  It's a pretty simple thing to do and could save us time and improve code maintenance.

gre:
I'm not sure we should create a PR straightaway but definitely it should be:
- a task
- a console.warn that happens when you use the "deprecated function"

we'll do better in future.

PR #1328, opened 2021-07-29
[LL-6935] Update LLC to use live-app-sdk

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-6935

## Description / Usage

Remove SDK duplicated code from LLC and use SDK package.

⚠️  Note:
As of today, and in order to avoid breaking change, LLC can't use SDK implementation of `serializePlatformSignedTransaction` and `deserializePlatformSignedTransaction` because SDK is not aware (and should not be aware) of Live opinionated `Operation` specific logic (i.e: serialize / deserialize)

## Expectations

Tests have been added for the `serializePlatformSignedTransaction` and `deserializePlatformSignedTransaction` functions. All functions imported from the SDK have associated tests [there](https://github.com/LedgerHQ/live-app-sdk/tree/main/tests).
The changes of this PR should have zero impact on userland since it uses similar (copy / pasted) code from the SDK. The only 2 functions with a different implementation (`serializePlatformSignedTransaction` and `deserializePlatformSignedTransaction`) have been kept as is.

-----
# Previous discussion on architecture / dependencies issues (kept for reference)

## Context (issues, jira)

Relates to https://ledgerhq.atlassian.net/browse/LL-6370

Today a large chunk of the [ledger-live-platform-sdk (LLP-SDK)](https://github.com/LedgerHQ/ledger-live-platform-sdk) code is duplicated in `ledger-live-common/src/platform/`, this is not sustainable.

### Proposal 1

This is a draft proposal to use the SDK in live-common to remove this duplicated code.

Today it looks like `ledger-live-common` does not use any of these functions internally and just passes them to LLD and LLM by exporting theme. It merely acts as a proxy between LLD/LLM and LLP-SDK (ledger-live-platform-sdk).

This chained dependency architecture might hinder development and deployment since each time a function is added or updated in the SDK, we will need to update LLC (ledger-live-common), LLM and LLD.

```
                             |--> Bump LLD
Bump LLP-SDK ---> Bump LLC --|            
                             |--> Bump LLM

```

### Proposal 2

Another proposition is to break this dependency chain and only use what's actually needed from the SDK in LLC (i.e only 2 types today, `PlatformAccount` and `PlatformCurrency`, used in `src/platform/converters.js`) and make LLD and LLM import the SDK for their own needs (serializers, types, etc...).

We could then update LLD and LLM independently from LLC in regard to SDK update

```
              |--> Bump LLD
Bump LLP-SDK -|--> Bump LLC (only if need be, i.e if we update the 2 types used by LLC today)           
              |--> Bump LLM

```

Testwise, this second proposal would also be easier since all the SDK related tests would be done in the SDK repo, we would not need to duplicate such tests in LLC.

## Avoiding code duplication and centralisation

Some code from LLC is needed by LLP-SDK (mainly types such as [operation](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/types/operation.js) for instance). We should find a way to properly handle this. Some solutions are possible, each one with its pro and cons (list is not exhaustive):

- duplicate needed code from LLC to LLP-SDK
Easy and quick to implement but a bullet in the foot mid/long term because of code duplication -> harder to maintain and a good chance to blow up in our face as soon as we start updating one version of this duplicated code and not the other 🤯 

- move duplicated logic to LLP-SDK and make LLC depend on LLP-SDK
Avoid code duplication and related maintenance issues, but is it the concern of LLP-SDK (i.e: in the scope of this project) to implement this shared logic

- make LLP-SDK depends on LLC
Reinforce the centralisation and dependence of the platform around LLC, making LLC even more of a bottleneck as it is today. Furthermore, this would make LLP-SDK import the whole LLC package with all of it's dependencies whereas it would just use a fraction of a tiny part of the lib (today only some types)

- split up LLC and extract duplicated code in separate smaller and independent packages (as proposed during last hack-the-live event)
We could think of a `ledger-live-types` kind of package regrouping common types used throughout the platform, such as operation (today used by LLC and LLP-SDK at least).
Avoid code duplication and centralisation. Avoid bottleneck around LLC.

----

**EDIT:**

For now we are going to KISS (keep it simple and stupid) and stay with this PR as it is.

Once the SDK will be more stable (once a v1 will be out for example) we could implement **Proposal 2** and start brainstorming about possible refacto opportunities.


src/platform/types.js
```diff
@@ -48,53 +53,4 @@ export type PlatformApi = {
...
-export interface PlatformBitcoinTransaction extends PlatformTransactionCommon {
-  family: "bitcoin";
-  feePerByte: ?BigNumber;
-}
-
-export type PlatformTransaction =
-  | PlatformEthereumTransaction
-  | PlatformBitcoinTransaction;
-
 export type PlatformSignedTransaction = SignedOperation;
```
achabrolin-ledger, line 56: This type should also be implemented in the SDK, but since it depends on a more complex operation type this has not been done today. cf. https://github.com/LedgerHQ/ledger-live-platform-sdk/blob/main/src/types.ts#L15-L22

Regarding this operation type, we will need to find a clever way to avoid duplication between LLC and LLP-SDK (maybe extracting it on a separate package / repo from which LLC and LLP-SDK would depend 🤔 ).
This could also make sense if some other project, present or future (maybe Vault) also use a quite similar operation type

package.json
```diff
@@ -117,6 +117,7 @@
     "invariant": "^2.2.2",
     "isomorphic-ws": "^4.0.1",
     "json-rpc-2.0": "^0.2.18",
+    "ledger-live-platform-sdk": "git+ssh://git@github.com:LedgerHQ/ledger-live-platform-sdk.git#add-serializeAccount",
```
achabrolin-ledger, line 120: For now the dependence is a git repo because no package has been published yet (the SDK is still actively WIP) but obviously once a first version is published, it will be handled as any other packaged dep

gre:
beware we're likely going to merge TS before this.

achabrolin-ledger:
The [`export type PlatformSignedTransaction = SignedOperation;` ](https://github.com/LedgerHQ/ledger-live-common/pull/1328/files#r678975989)should also be implemented in the SDK, but since it depends on [a more complex operation type](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/types/operation.js) this has not been done today.
cf. https://github.com/LedgerHQ/ledger-live-platform-sdk/blob/main/src/types.ts#L15-L22

Regarding this operation type, we will need to find a clever way to avoid duplication between LLC and LLP-SDK (maybe extracting it in a separate package / repo from which LLC and LLP-SDK would depend 🤔 ).

haammar-ledger:
My :v: cents. Purely from a design perspective, **proposal 2** and a `ledger-live-types` package seem the best solutions.
We just need to reality-check what they imply in dev cost and side effects.

achabrolin-ledger:
By the way, it looks like we already have a breakdown of LLC in the backlog https://ledgerhq.atlassian.net/browse/LL-6548

gre:
I think CI should be involved / important, once the work is ready. if it's not yet pushed on NPM, it's not ready to get in (i'm putting it back to Draft)

we can push to NPM if you think it's good.

achabrolin-ledger:
> I think CI should be involved / important, once the work is ready. if it's not yet pushed on NPM, it's not ready to get in (i'm putting it back to Draft)
> 
> we can push to NPM if you think it's good.

Indeed makes sense. Not sure it's ready yet to be published to NPM. We will si this in the coming days / weeks when we have more insights regarding SDK requirements.
I should be able to use yalc for testing purposes for now 👌

src/platform/types.js
achabrolin-ledger: Here is a proposed solution to decouple operation from LLP-SDK

tldr; SDK is operation agnostic and all operation related logic (highly correlated to Live specific logic) stays in LLC.

https://github.com/LedgerHQ/live-app-sdk/pull/20

LFBarreto (approved):

PR #1330, opened 2021-07-30
LL 6247 - Make tron resources mandatory

ghost:
## Context (issues, jira)
[LL-6247](https://ledgerhq.atlassian.net/browse/LL-6247)

## Description / Usage

Until now, when trying to get the `accountShape` of a Tron account, if the Tron API did not return an account (which happens when an account has transaction yet), it would early return an object not containing any `tronResources` which led to unexpected behaviours in the live apps.

## Expectations

&nbsp;
#### Impact
As of right now, the absence of `tronResources` is handled in both LLD and LLM. Adding it again should not have any impact, and once this PR merged, the fail safe in LLD & LLM should/will be removed.

### DIFF
--- a/src/__tests__/__snapshots__/all.libcore.ts.snap
+++ b/src/__tests__/__snapshots__/all.libcore.ts.snap
@@ -39579,6 +39579,20 @@ Array [
     "starred": false,
     "swapHistory": Array [],
     "syncHash": undefined,
+    "tronResources": Object {
+      "delegatedFrozen": Object {
+        "bandwidth": undefined,
+        "energy": undefined,
+      },
+      "frozen": Object {
+        "bandwidth": undefined,
+        "energy": undefined,
+      },
+      "lastVotedDate": undefined,
+      "lastWithdrawnRewardDate": undefined,
+      "tronPower": 0,
+      "votes": Array [],
+    },
     "unitMagnitude": 6,
     "used": false,
   },
--- a/src/__tests__/test-helpers/setup.ts
+++ b/src/__tests__/test-helpers/setup.ts
@@ -1,7 +1,19 @@
+import BigNumber from "bignumber.js";
 import { setSupportedCurrencies } from "../../currencies";
 import { setPlatformVersion } from "../../platform/version";
 
 jest.setTimeout(180000);
+
+expect.extend({
+  toBeBigNumber(value) {
+    const pass = BigNumber.isBigNumber(value);
+    const message = pass
+      ? () => `${value} is a BigNumber`
+      : () => `${value} is not a BigNumber`;
+
+    return { message, pass };
+  },
+});
 
 setPlatformVersion("0.0.1");
 
--- a/src/api/Tron.ts
+++ b/src/api/Tron.ts
@@ -22,6 +22,7 @@ import {
   formatTrongridTxResponse,
   formatTrongridTrc20TxResponse,
   hexToAscii,
+  defaultTronResources,
 } from "../families/tron/utils";
 import { log } from "@ledgerhq/logs";
 import { TronTransactionExpired } from "../errors";
@@ -475,11 +476,21 @@ export const extractBandwidthInfo = (
     gainedLimit: new BigNumber(0),
   };
 };
-export const getTronResources = async (
+
+export function getTronResources(): Promise<TronResources>;
+export function getTronResources(
   acc: Record<string, any>,
   txs: TrongridTxInfo[],
   cacheTransactionInfoById: Record<string, TronTransactionInfo>
-): Promise<TronResources> => {
+): Promise<TronResources>;
+export async function getTronResources(
+  acc?: Record<string, any>,
+  txs?: TrongridTxInfo[],
+  cacheTransactionInfoById: Record<string, TronTransactionInfo> = {}
+): Promise<TronResources> {
+  if (!acc) {
+    return defaultTronResources;
+  }
   const frozenBandwidth = get(acc, "frozen[0]", undefined);
   const frozenEnergy = get(
     acc,
@@ -547,8 +558,8 @@ export const getTronResources = async (
     const lastOp = txs.find(({ type }) => type === "VoteWitnessContract");
     return lastOp ? lastOp.date : null;
   };
+  const lastVotedDate = txs ? getLastVotedDate(txs) : undefined;
 
-  const lastVotedDate = getLastVotedDate(txs);
   return {
     energy,
     bandwidth,
@@ -561,7 +572,8 @@ export const getTronResources = async (
     lastVotedDate,
     cacheTransactionInfoById,
   };
-};
+}
+
 export const getUnwithdrawnReward = async (
   addr: string
 ): Promise<BigNumber> => {
--- a/src/api/__mocks__/Tron.ts
+++ b/src/api/__mocks__/Tron.ts
@@ -1,3 +1,5 @@
+import BigNumber from "bignumber.js";
+
 import type { SuperRepresentative } from "../../families/tron/types";
 import superRepresentatives from "./superRepresentativesData";
 
@@ -12,3 +14,10 @@ export const __NEXT_VOTING_DATE__ = new Date(0);
 export const getNextVotingDate = (): Promise<Date> => {
   return Promise.resolve(__NEXT_VOTING_DATE__);
 };
+
+export const extractBandwidthInfo = (): Record<string, unknown> => ({
+  freeUsed: new BigNumber(0),
+  freeLimit: new BigNumber(0),
+  gainedUsed: new BigNumber(0),
+  gainedLimit: new BigNumber(0),
+});
--- a/src/families/tron/bridge/js.ts
+++ b/src/families/tron/bridge/js.ts
@@ -308,9 +308,12 @@ const getAccountShape = async (info, syncConfig) => {
   const tronAcc = await fetchTronAccount(info.address);
 
   if (tronAcc.length === 0) {
+    const defaultTronResources = await getTronResources();
+
     return {
       blockHeight,
       balance: new BigNumber(0),
+      tronResources: defaultTronResources,
     };
   }
 
--- a/src/families/tron/synchronisation.test.ts
+++ b/src/families/tron/synchronisation.test.ts
@@ -1,0 +1,64 @@
+import BigNumber from "bignumber.js";
+
+import { getCryptoCurrencyById } from "../../currencies";
+import { syncAccount } from "../../__tests__/test-helpers/bridge";
+import { defaultTronResources } from "./utils";
+import bridge from "./bridge/js";
+import { Account } from "../../types";
+
+const tron = getCryptoCurrencyById("tron");
+const defaultSyncConfig = {
+  paginationConfig: {},
+  blacklistedTokenIds: [],
+};
+
+const dummyAccount: Account = {
+  type: "Account",
+  id: "",
+  derivationMode: "",
+  seedIdentifier: "",
+  name: "Tron",
+  starred: false,
+  used: false,
+  currency: tron,
+  index: 0,
+  freshAddress: "",
+  freshAddressPath: "",
+  swapHistory: [],
+  freshAddresses: [],
+  blockHeight: 0,
+  balance: new BigNumber(0),
+  spendableBalance: new BigNumber(0),
+  operationsCount: 0,
+  operations: [],
+  pendingOperations: [],
+  unit: tron.units[0],
+  lastSyncDate: new Date(0),
+  creationDate: new Date(),
+  balanceHistoryCache: {
+    HOUR: {
+      balances: [],
+      latestDate: undefined,
+    },
+    DAY: {
+      balances: [],
+      latestDate: undefined,
+    },
+    WEEK: {
+      balances: [],
+      latestDate: undefined,
+    },
+  },
+};
+
+describe("Tron Accounts", () => {
+  test("should always have tronResources", async () => {
+    const account = await syncAccount(
+      bridge.accountBridge,
+      dummyAccount,
+      defaultSyncConfig
+    );
+
+    expect(account.tronResources).toEqual(defaultTronResources);
+  });
+});
--- a/src/families/tron/test-specifics.ts
+++ b/src/families/tron/test-specifics.ts
@@ -1,10 +1,19 @@
+import sumBy from "lodash/sumBy";
+import { syncAccount } from "../../__tests__/test-helpers/bridge";
+import dataset from "../../families/tron/test-dataset";
+import { getAccountBridge } from "../../bridge";
+import type { Account } from "../../types";
+import {
+  fromAccountRaw,
+  decodeAccountId,
+  encodeAccountId,
+} from "../../account";
 import {
   fetchTronAccountTxs,
   getAccountName,
   getBrokerage,
   getTronSuperRepresentativeData,
 } from "../../api/Tron";
-import sumBy from "lodash/sumBy";
 
 export default (): void => {
   describe("tron super representative data", () => {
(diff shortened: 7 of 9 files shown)


klambert-ledger:
## About testing

I would love to add a specific test for this use case (empty account getting the right `accountShape`) but `getAccountShape` isn't exported and I don't know yet if there is an already used solution to this kind of issue (e.g. `rewire` ?), so I might need a little bit of help to cover this.

gre:
- generally speaking, the dataset are tested https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/tron/test-dataset.js . in theory there is a way to add a test on each account https://github.com/LedgerHQ/ledger-live-common/blob/7b5f27a3eb82ff7b7d754f02143b869331d9557b/src/types/bridge.js#L177 but apparently no way to do for "all accounts".
- example of testing a sync "independently" https://github.com/LedgerHQ/ledger-live-common/blob/ab7c58f65e261d37be39cc8ebe7ea59ddbf44912/src/families/tezos/synchronisation.test.js

src/__tests__/test-helpers/bridge.js
```diff
@@ -425,6 +425,38 @@ export function testBridge<T>(family: string, data: DatasetTest<T>) {
...
+                "delegatedFrozen",
+                "energy",
+                "frozen",
+                "lastVotedDate",
+                "lastWithdrawnRewardDate",
+                "tronPower",
+                "unwithdrawnReward",
+                "votes",
+              ]).toEqual(expect.arrayContaining(Object.keys(tronResources)));
+            });
```
klambert-ledger, line 458: Jest has really been a pain in the $$$ to make what looked like a pretty simple test "Does tronResources have all this properties/type combo and some can be undefined". 

That's the best thing I could find without risking 10 sick days for mental breakdown.

src/api/Tron.js
```diff
@@ -491,10 +491,31 @@ export const extractBandwidthInfo = (
...
+      },
+      votes: [],
+      tronPower: 0,
+      energy: BigNumber(0),
+      bandwidth: extractBandwidthInfo(),
+      unwithdrawnReward: BigNumber(0),
+      lastWithdrawnRewardDate: undefined,
+      lastVotedDate: undefined,
+      cacheTransactionInfoById: {},
+    };
```
haammar-ledger, line 516: I think you don't have to modify the `getTronResources` method, you can just define this data as a constant somewhere, maybe in `utils.js`.

src/families/tron/synchronisation.test.js
```diff
@@ -0,0 +1,70 @@
...
+        freeUsed: BigNumber(0),
+        freeLimit: BigNumber(0),
+        gainedUsed: BigNumber(0),
+        gainedLimit: BigNumber(0),
+      },
+      unwithdrawnReward: BigNumber(0),
+      lastWithdrawnRewardDate: undefined,
+      lastVotedDate: undefined,
+      cacheTransactionInfoById: {},
+    };
```
haammar-ledger, line 66: If you define a constant as I suggest, you can use it here as well.

src/families/tron/synchronisation.test.js
```diff
@@ -0,0 +1,70 @@
...
+  balance: BigNumber(0),
+  spendableBalance: BigNumber(0),
+  operationsCount: 0,
+  operations: [],
+  pendingOperations: [],
+  unit: tron.units[0],
+  lastSyncDate: new Date(0),
+  creationDate: new Date(),
+  balanceHistoryCache: {},
+});
```
haammar-ledger, line 40: It seems you don't need a function for that, you can just define the data as a constant as well.

src/families/tron/test-dataset.js
```diff
@@ -705,6 +705,7 @@ const dataset: DatasetTest<Transaction> = {
               To re-enable when the support will be done.
             */
             "balance is sum of ops",
+            "account always has tronResources",
```
haammar-ledger, line 708: Does this test still belong in FIXME?

src/api/Tron.js
klambert-ledger: As you prefer 👍 

It felt to me that it was a bit more logical to always call the `getTronResources` no matter what use case you have (with account or not) to actually getting `tronResources` (like an unfailable API pretty much). Kinda like the opposite of "separation of concerns", a "reunion of concerns" if that exists. 😆 

I tried to avoid modifying the api too much by letting `txs` & `cacheTransactionInfoById` mandatory to avoid being too confusing. 

Do you think I can mix your approach with this one by declaring those default `tronResources` in `utils.js` and returning it in `getTronResources` ? If not, I'll make the change asap 👍

src/families/tron/synchronisation.test.js
klambert-ledger: Yeah I know, I tried to copy what Gaetan did [here](https://github.com/LedgerHQ/ledger-live-common/blob/ab7c58f65e261d37be39cc8ebe7ea59ddbf44912/src/families/tezos/synchronisation.test.js#L69) in the case at some point we would need to add parameters to create customisable accounts. 

I'll remove it for now and we'll add it later if needed 👍

src/families/tron/test-dataset.js
klambert-ledger: I actually have no idea, I tried adding a `tests` property in the dataset, but it was falling if I remember correctly, so I just moved with the flow here without really knowing what the FIXME was for 😛 

What would be the actual good way to declare the test ?

src/api/Tron.js
haammar-ledger: Actually it's the call that bugs me ^^ `await getTronResources(null, [], {})` with these dummy params.
I'd rather keep the current expectations of the API method, but up to you.

src/families/tron/test-dataset.js
haammar-ledger: Yeah I agree this can be pretty confusing :sweat_smile: 
The `FIXME_` prefixed sections are actually used to bypass things. Here, the `FIXME_tests` section belongs to an account, and currently says that the generic `"balance is sum of ops" `test shouldn't be executed for this account.
Adding your new test here will do the same, I'm guessing this is not what you wanted?

haammar-ledger: However, it just hit me that you kinda added a currency-specific test with the generic ones.
I didn't run your new tests, but doesn't it fail for currencies other than `tron`?
e.g. if you run: `yarn test --testNamePattern "account always has tronResources"`.
I think the way to do it is to "just" move this test to the `test-specifics.js` file in the tron family.

src/api/Tron.js
klambert-ledger: Yeah I agree... It was that or making `txs` & `cacheTransactionInfoById` optional which felt to me like it was super confusing when using the method since it's more a question of "all or nothing" for the parameters 😕 

I think I found a better solution from this (stackoverflow)[https://stackoverflow.com/a/36274962], just don't know if it's doable with flow yet, I have to check.

src/families/tron/test-dataset.js
klambert-ledger: > Adding your new test here will do the same, I'm guessing this is not what you wanted?

Yes it is, I need to test again doing this, cause from what I can remember it didn't work for whatever reason... 

> However, it just hit me that you kinda added a currency-specific test with the generic ones.
I didn't run your new tests, but doesn't it fail for currencies other than tron?

Actually I followed the insights of @gre just above, but I might have misunderstood how the `makeTest` is working. To me it felt like I was creating a test that would then be addable to a `dataset` specifically, not all the coins... If it's not the case then yes, it's going to fail everywhere else... 
As of right now I tested my tests with `FAMILY=tron yarn test path/to/all.libcore.js` and ignored the tests that were falling which were not mine cause they were related to cosmos mostly (I thought it was just a bug of the filtering method)... I launched a complete `yarn test` 2 hours ago pretty much and was waiting to see if it worked as expected 😖

haammar-ledger: Yeah running the full `yarn test` is painful...
Some filters can be very handy, for example:

yarn test --runTestsByPath <one single test file, e.g. `src/__tests__/all.libcore.js`> --testNamePattern <name or description of a test, e.g `"balance is sum of ops"`>

src/api/Tron.js
klambert-ledger: Ok so I found a solution for Flow but my IDE is getting crazy for whatever reason... I tried in the Flow online validator which declare no issue... I don't know if it's just VSCode getting a false positive or not 😞 

Do you see any problem in [this](https://flow.org/try/#0PTAEFpKrQEQeQHIHIAqoAy94GlQFVFYBRAJVFQAkBJAZVGtWXoCl9b0BZeAYR2sQBxUAEEiFAJoAFYrQBQICNGihB8CuowDioABy7QAQjkmALgE8ADgFNQAIQCWAc0QBXALYAjawCdQAXlAAOw9vHwBuMytbVB8AeyCnHwcAE1QAD2oggDM4gNAAQyDzSLkLGzhrABtrJwLTaxSAMXiAL2sgrNz8gG8AHzlQQvc41yDTAC57ZzcvXwAaOT6AX1Ly2xa49s6cvMD+weHR8anHF1CFw+t0ywcfRpFJuHrrRZW16NAANTiG3oGhgUUil7gBnUFTUGmZKJRZDABuv2sPGOTxCcx8b1WUQqdiKKQA7qlTAALLp7UAHIbZe7WfCgxqnGYXTGHGnWawYBzuBxPM6zMJw0B1BxBRr0xnTc4YoUisUpLk8vnMmVLbFlT6xBKxIqggoAY1MDgS5P+bI5U3RgsOniqcX1AGsBb5LSyhUTSSCCgSAPoFEZjNFuw5jdnWdp+gMnYLB944mLxIKkayg0Y+fUps3UtodKY9Q5DTz4okpUlTAD8m225KFQw6vic5grVY6NcOyyFKWqtRezRzQTzBdARaChOJJIrsG7dQafa2rd2tdA9Z8jcn097LZ2uSFHcOiIaEO+SIA2gBdIXQhJSOIEl0x1V1sWrptS52swvF8enL+lsmLkMgg9EkvQJJNrAJAofBSJlpWtIYqgKKEAHVeRAnxvXAyDoNgF5JxeIVEKhH5Z1whp8IaIV9QNElrB1II9UNY1tziOxzGoGDKVAE8fUhaFRScM8pi1IJ6MYo0TV2UA9zjOR9QSKFlyCeSuwAVl0HhaMdKYAAooRhJwAEoAgAPlAfSBPyUFTPM0p5IY0xhWsJ4dKKcx5kKYoPLc4z-DMtz8lc+ZPHmfVfLM5B5NcKLkDshTHKcZyRJEfUovGRBnIJOIfAdXSfJsqR4h5BkAB43LMwIdPhcLQEKuJiusAA6ME4iqeFrCqwzwlAEx7MUxLTHwIC0NArCoM41zgR8PiDJquqGpK-kWQq0BJpBOaiocBlmpTVr2p0paMSaIIdIABkMwzevit8WWOwKgk8V0MRqh64oc5d0mhA1TDxUcS1JU1KoKGr8yGIZ7lMVwfCCSkhzDCVOMOsJjrOwylzDRVeVg98UfOpc5XFBlEZVZGTrxocCYVbksZuo6ybR9tImWEwFDAZQVAQFB0CwXBRDseAvh0Kg6AYJhWHYLheH4IRRHEVBpFkVmlHZ8A4CQNBMGwPARH5wWKBoehGGYUA2A4UBuD4ARhDEWBJBkeRFBVyA1a5zXeZ1gWhYN0XjdNyXLZlm27cVlnQ5MLt9UQ+5QGyMYmISJzTBE5NUyhjMdMMqZ5q26wSuT3a05TEzIkUFMqlFUxwBSLaCltaxwDFT7wHLsVgjicB7gjqPrDkLuoNsWPlIkmGBvz1P0w6w4DX1KZ4E8AAraxDUvdIjxEpJUgyclzyomi6IwhjvuY8k2I4vNuN48z+MSISKETMSj8k7pmcz2rNtKse0wzUFi6VsuK6rjXOuDdriVxbrYII7dO5L27nIa4lhsqOSQuYZSMc47D0Tp-QuOkp6pXLLPBeS9TAr1BPgu+CQN5pEyLsHehxqL6log-eOLFT4pDIT0C+M0BK3xEkw4epoX5Z3frnLBE8f6wyGA4bIq1DDT2MqDMGEMoYwwUWDGO-ZBxqLUSOMcf4phjC7NkUUjQlxqJXGuUABjrBGPlKY6Sdiuw1BnI0LcmitGfj+t+Sxo5rHGJSHYp8DZXxWJsSYocQw9zuIPCmKYtD3FXiCDeO801QCnTseY18SNfC4wZu4nR-0JwfS+oaX6uiAa7AznYsYwFRrJmwsTOC2T6Z2KIqYVCnoMJgTqeNMi1h9E+NCf48JoBWkkUaL0-phi-F2PoYwg+4lj67FYXmSJYNsQRJMEMPqjkaTziCKUgp+QBquVSh5AARLs7YJ5TpnjOR5EJfiup0OupcjoxBnyNiOc5HBajp5LjOdPVEPoWqF0aq8oIPoiyIWUtYH0uQfA+gyXcocDz5SHCec896jieyzi3Ac8cXzTA-LBn8ocZzsXOJSHC-skKCjQozHC7KtLPF-mRWo1FjR0WRC2ddClm5+zvKCYS4lgJTlksBYGYFBcJ6NT5bOalezaX0thfCxFHzzBsrBhylIXLNmgG2UpFSDwpopmsoEDoRr1KaSXg6E5+pGpAhBKajFPL3oJIyqYLKOVAaFEgryTBiYUppVMB6r1tqLVxC7CkEQJrwRPNdYpapI1OljWgvkb0BR-UDSGjUlN3ToI6QjVGmNTq42lATY5DJ+R3WZWyg6ckjUMmY1MI1HkIRQQ6RrZ6utDaMkI3jfq66+SCXms+hhEpv5ym5E7YmUNPbdgusHe9cFvQhzDr0eovZ+K-zDPLBI9xJKozKkaT4FG4Lt2kjBTSqFRQMy5IPXWG4dwHhoggs8BoOlz2TpJI2p99wfRGncNYe97jmbuMmb42xQ4MlTHBYKl8u790Pv9KibGt0TpwfVVexVN6YUgYPfA590bX0EnfR1TDQTf23H-YB4DATpLDIg4M3c5al2KTlS4-sq7tHfqmBxuc2wL0kkQxwlDgY0N0x0vxvF37jJga0Ux6Z0H1V8Y3LigV6qRNHHE7TUmUm1Ocb2fBxscnGPeKmVBiJrGDUJKSb4fIWTT0nWOeC8567L1ifGHc1JF1rPXWiWaxOdrzkBe8+eQyraCiWB0p1GyOlVGFFjUeeEjVol+iS0uaJKIdMpbS8G3cvm9UGtae09CmF80pF6em1KjVEKHlMD6XN3oAPch7movdYpSO9LtbVl4UJGvJua7Ry6ajFPylY4oVA8AEBTHuFUcwoAE6klsBK8Yi3CGGlAASWiMMEmUK2w4KoVRQCWHiPCVIthSRbVAKKXIjVMX9WchgJCpgxmVZeIFUwq9hKJkoVvGhZ5X7liq35JDbHHKtPgJYatq8wWihSDFjh6xpI1WR-4dHoAzljPaWKcEKJxjjtMGcgd4NnLKJGS9qHoA92Q8sLKj7rojvcuksz4rL23sg9AF96ye6BrPeIkid7H7ufGTG40VjSjoZg4yUudzJJ0b9iXNJxXQ4AtLls7eS47LhodPKxBcaS4SuDa6frnCBEhyjMF70pcsz966kfiw9inFbd8MWbkVh1O909FWdiZmchR6JhTl-DqTyA8JCD9gkIR3Q9JUD9K9O3uY9Jzj+PBPHZuKRVGDFQHkQw-gVTx1b3HkTyZ+iqMZAF5KTLC6kAA) ?

gre:
@klambert-ledger @haammar-ledger  I hear the pain.

it's quite machineàgaz right now but the dataset testing is pretty complete too.

there are a annoying tech constraint that made us having to do this all.libcore.js test unfortunately: lib-ledger-core.

if you try to split into multiple test file, libcore will just 🤯  due to fact jest instanciating diff thread, it will LOCK, we never figured out a way to solve this.

when libcore is out of the picture, we'll split these all into test file on each family and this will be a BIG RELIEF.

gre:
a trick right now is to do

```
FAMILY=tron yarn test all.libcore
```

but still painful.

klambert-ledger:
> ```
> FAMILY=tron yarn test all.libcore
> ```

Thx for the even shorter version of this @gre 👍  

But discussing with @haammar-ledger, it's still unclear to me if my test will be executed on all currencies instead of just in the `dataset` that contains the test that I added. 😖 

My understanding of this automation was that using a `makeTest` in `src/__tests__/test-helpers/bridge.js` would make a test available (but not mandatory) to all `test-dataset.js` files and adding the test by name in `FIXME_tests` would allow me to make this test mandatory for this specific currency. 

Am I understanding this correctly ? (Because for example running `FAMILY=tron yarn test all.libcore` gives me failing tests related to @ledgerhq/ledger-core, but it looks that it's also going through tests related to cosmos and tezos ?)

@haammar-ledger is also suggesting to move this test to `test-specifics.js`, so I really don't know where to put this test anymore ! 😆 

Right now I'm thinking about robbing a pharmacy which would be the only way to get enough Aspirin to stop my pain. 💊

src/api/Tron.js
```diff
@@ -490,11 +491,27 @@ export const extractBandwidthInfo = (
...
+// eslint-disable-next-line no-redeclare
+export async function getTronResources(
+  acc?: Object,
+  txs?: TrongridTxInfo[],
+  cacheTransactionInfoById?: { [_: string]: TronTransactionInfo }
+): Promise<TronResources> {
+  if (!acc) {
+    return defaultTronResources;
+  }
+
```
klambert-ledger, line 514: Ok so best solution I got to allow for getTronResources to be called without params or with all params has been to overload the function and give it 3 declarations (1 without params, 1 with all the params, and 1 that is the combination of all which is mandatory). 

It's exactly what we want I guess, but unfortunately, between the ESLint plugin for flowtype not handeling the `declare` keyword correctly to ignore redeclaration + Flow just below that magically forget that the function is `async` and thinking that our return isn't a Promise, it's not most beautiful code in the world... 

I would need the opinion of other people on this. @haammar-ledger suggested to just let `getTronResources` as it was and just create another util to do a `getDefaultTronResources` which would work obviously but kinda lose the point of reuniting all the concerns about _getting tron resources_. 

I some of you have the time to give me an opinion on this, it would be greatly appreciated 😄

haammar-ledger:
@klambert-ledger can you please merge or rebase to solve the conflict, so the CI can run? Thanks.

gre:
dataset tests are clearly not meant to be ran locally, at least not all.libcore.js. indeed it's an issue, not really related by jest but more to libcore constraints of inability to run in parallel when using the same database. but well, tron is not backed by libcore, so we could already migrate all non libcore coins to a better approach = basically moving back that into a test file on each family.

src/api/Tron.js
klambert-ledger: Clarity has been resolved with the TS migration 👍

gre (approved):
go

PR #1337, opened 2021-08-06
add: GMT icon

alexgev:
## Expectations


src/data/icons/svg/GMT.svg
```diff
@@ -0,0 +1,18 @@
+<?xml version="1.0" encoding="utf-8"?>
+<!-- Generator: Adobe Illustrator 25.3.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Слой_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 24 24" style="enable-background:new 0 0 24 24;" xml:space="preserve">
+<style type="text/css">
+	.st0{fill:#FFFFFF;}
+</style>
+<g id="Слой_2_1_">
+	<g id="Слой_1-2">
```
gre, line 9: here is an acceptable example of svg https://raw.githubusercontent.com/LedgerHQ/ledger-live-common/master/src/data/icons/svg/BAT.svg

we must have:
- no id in svg (otherwise if adding twice same id in dom, it would conflict)
- no `<style> / class` because it wouldn't work in React Native

you can see if there is a way to clean the svg but right now we can't merge it. thanks

gre (changes requested):

gre:
Please reopen when it fits the requirements. thanks.

PR #1345, opened 2021-08-12
Add operationMode type for Polkadot

ghost:
## Context (issues, jira)

Define a strict `PolkadotOperationMode` instead of using general `string` type, since there is a set list of available Polkadot modes.
As already done with [Algorand](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/algorand/types.ts#L106), [Tezos](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/tezos/types.ts#L27), [Tron](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/tron/types.ts#L10) or [Cosmos](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/cosmos/types.ts#L113).

Also fix an unrelated type clash (cf. 9d5c19ccc62b3d762e1d843d6e96e0afb2ba2d56) 

## Description / Usage

Having a stricter type checking for Polkadot mode in order to prevent issues.

## Expectations

### DIFF
--- a/src/families/polkadot/js-signOperation.ts
+++ b/src/families/polkadot/js-signOperation.ts
@@ -5,7 +5,12 @@ import { u8aConcat } from "@polkadot/util";
 import { FeeNotLoaded } from "@ledgerhq/errors";
 import Polkadot from "@ledgerhq/hw-app-polkadot";
 import type { Transaction } from "./types";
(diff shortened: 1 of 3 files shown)


gre (approved):

gre:
any context on jira side?

PR #1351, opened 2021-08-18
[LL-6803] generalized sign transaction flow

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-6803

⚠️ linked PR:
~🔗  https://github.com/LedgerHQ/ledger-live-common/pull/1328~
🔗  https://github.com/LedgerHQ/ledger-live-desktop/pull/4112
~🔗  https://github.com/LedgerHQ/ledger-live-platform-sdk/pull/18~

## Description / Usage

Add transactions adapters from LLP-SDK tx to LLC tx, used by generalisation work done in the [desktop PR](https://github.com/LedgerHQ/ledger-live-desktop/pull/4112)

Also update the `useFeesStrategy` for BTC and ETH to drop invariant and return empty array `[]` when no network info available yet (needed by [BTC](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/families/bitcoin/SendAmountFields.js#L60) and [ETH](https://github.com/LedgerHQ/ledger-live-desktop/blob/develop/src/renderer/families/ethereum/SendAmountFields.js#L33) `SendAmountFields` components)

## Expectations

Should have no impact on userland since it mainly adds functions. Must be merged before the related LLD PR since LLD relies on the newly implemented functions in this PR.

----

## TODO

- [x] Add tests regarding platform converter / adapters
- [x] publish new version of SDK and update dependency (to fix Polkadot transaction mode type error)

### DIFF
--- a/src/families/bitcoin/react.ts
+++ b/src/families/bitcoin/react.ts
@@ -1,9 +1,10 @@
-import invariant from "invariant";
 import type { NetworkInfo, Transaction } from "./types";
 import type { FeeStrategy, Account } from "../../types";
 export const useFeesStrategy = (a: Account, t: Transaction): FeeStrategy[] => {
   const networkInfo = t.networkInfo;
-  invariant(networkInfo, "no network info");
+
+  if (!networkInfo) return [];
+
   const strategies = (networkInfo as NetworkInfo).feeItems.items
     .map((feeItem) => {
       return {
--- a/src/families/polkadot/platformAdapter.ts
+++ b/src/families/polkadot/platformAdapter.ts
@@ -1,0 +1,11 @@
+import { PolkadotTransaction as PlatformTransaction } from "ledger-live-platform-sdk";
+import { Transaction } from "./types";
+
+export const CAN_EDIT_FEES = false;
+
+export const convertToLiveTransaction = (
+  tx: PlatformTransaction
+): Partial<Transaction> => ({
+  ...tx,
+  era: `${tx.era}`,
+});
--- a/src/families/polkadot/platformAdapter.ts
+++ b/src/families/polkadot/platformAdapter.ts
@@ -1,5 +1,5 @@
 import { PolkadotTransaction as PlatformTransaction } from "@ledgerhq/live-app-sdk";
-import { PolkadotOperationMode, Transaction } from "./types";
+import { Transaction } from "./types";
 
 export const CAN_EDIT_FEES = false;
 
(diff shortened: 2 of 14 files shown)


src/families/polkadot/platformAdapter.ts
```diff
@@ -0,0 +1,12 @@
...
+import { PolkadotOperationMode, Transaction } from "./types";
+
+export const CAN_EDIT_FEES = false;
+
+export const convertToLiveTransaction = (
+  tx: PlatformTransaction
+): Partial<Transaction> => ({
+  ...tx,
+  era: `${tx.era}`,
+  mode: tx.mode as PolkadotOperationMode, // FIXME: update SDK polkadot transaction type to include a PolkadotOperationMode type
```
achabrolin-ledger, line 11: Should be fixed by https://github.com/LedgerHQ/live-app-sdk/pull/22

lkrust (changes requested):
-

lkrust (approved):

src/platform/converters.ts
```diff
@@ -29,3 +41,66 @@ export function currencyToPlatformCurrency(
     })),
   };
 }
+
+export const getPlatformTransactionSignFlowInfos = (
```
juan-cortes, line 45: Isn't this the ideal place for something that gets exported `byFamily` and we just provide the family to the object to get the method? Like we do with any of the generated code by family, ie https://github.com/LedgerHQ/ledger-live-common/blob/master/src/generated/hw-getAddress.ts you would then import this byFamily and pas the platform 🤷🏼‍♂️

juan-cortes (approved):
lgtm

src/platform/converters.ts
achabrolin-ledger: Maybe, I'm not yet familiar with this `byFamily` kind of abstraction. I'll take a look at this and propose an update accordingly

achabrolin-ledger: Updated here e0c3b5ffeaea9e200421fab8bc1afb075b971dcf

IAmMorrow (approved):

src/families/bitcoin/react.ts
```diff
@@ -1,9 +1,10 @@
-import invariant from "invariant";
 import type { NetworkInfo, Transaction } from "./types";
 import type { FeeStrategy, Account } from "../../types";
 export const useFeesStrategy = (a: Account, t: Transaction): FeeStrategy[] => {
   const networkInfo = t.networkInfo;
-  invariant(networkInfo, "no network info");
+
+  if (!networkInfo) return [];
```
valpinkman, line 6: this actually changes the behavior, wouldn't it be better with a 
```suggestion
  if (!networkInfo) throw new Error("no network info")
```

?

valpinkman (approved):

PR #1356, opened 2021-08-23
LL-5511 Bitcoin JS

ghost:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-5511

## Description / Usage

The full Bitcoin implementation in JS (TypeScript).
Relies on a separate lib created in the scope of this migration: https://github.com/LedgerHQ/wallet-btc

## Expectations

--> Existing codebase of unit & bot tests.
--> Minor PRs will be required on LLD & LLM just to update some import paths and data fields. No functional change expected.


src/bridge/jsHelpers.ts
```diff
@@ -116,11 +122,19 @@ export const makeSync =
...
+          const freshAddressPath = getSeedIdentifierDerivation(
+            initial.currency,
+            initial.derivationMode
+          );
+
           const shape = await getAccountShape(
             {
               currency: initial.currency,
               id: accountId,
+              index: initial.index,
```
juan-cortes, line 134: Is the index merely the last part of the derivation path?

juan-cortes (commented):
From my limited knowledge in the matter, it's looking great, some work pending as per all the comments you left, but what's here looks good.

src/families/bitcoin/js-synchronisation.ts
```diff
@@ -0,0 +1,378 @@
...
+
+  if (currency.id === "bitcoin") {
+    await requiresSatStackReady();
+  }
+
+  const paramXpub = initialAccount?.xpub;
+  // FIXME Hack because makeScanAccounts doesn't support non-account based coins
+  // Replaces the full derivation path with only the seed identification part
+  // 44'/0'/0'/0/0 --> 44'/0'
+  const path = derivationPath.split("/", 2).join("/");
```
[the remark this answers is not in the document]
haammar-ledger: Hmm I think I should remove the first comment line, which is nothing but confusing, but I don't have a workaround for the rest unfortunately.
The `derivationPath` param contains the full path and wallet-btc requires only the two first parts when calling `generateAccount`.

src/bridge/jsHelpers.ts
haammar-ledger: Soo actually I got bit by that, I just found out that for some obscure reason, the CLI provides a full derivation path (`44'/0'/0'/1/5`) but LLD provides only the root path! (`44'/0'`)
Pretty weird tbh. But I need to keep this `index` param to be safe...

cli/tests/encryption/test.sh
```diff
@@ -2,35 +2,35 @@
 
 set -e
 
-ledger-live sync -c bitcoin --xpub xpub6BosfCnifzxcFwrSzQiqu2DBVTshkCXacvNsWGYJVVhhawA7d4R5WSWGFNbi8Aw6ZRc1brxMyWMzG3DSSSSoekkudhUd9yLb6qx39T9nMdj -f summary
+ledger-live sync -c cosmos --id cosmospub1addwnpepqwyytxex2dgejj93yjf0rg95v3eqzyxpg75p2hfr6s36tnpuy8vf5p6kez4 -f summary
```
juan-cortes, line 5: Why is this cosmos now?

juan-cortes (commented):
Github actions has more complains than lines, incredible

cli/tests/encryption/test.sh
haammar-ledger: Because all tests in this file are testing libcore specifically, so I just replaced with the next remaining coin, until there are none left hopefully.

src/migrations/accounts.ts
```diff
@@ -16,6 +16,24 @@ function sameSwap(a: SwapOperation, b: SwapOperation) {
 }
 
 export const implicitMigration = (accounts: Account[]): Account[] => {
+  // TODO Need this for BTC-JS?
```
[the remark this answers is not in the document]
haammar-ledger: Some findings of mine on this topic:
- Already added accounts are actually migrated pretty well already. Former `libcore` ids are changed into `js` ids and everything keeps working seamlessly for the user.
- However add accounts doesn't seem to get this, and will therefore add a new duplicate for each...

Probable related code: https://github.com/LedgerHQ/ledger-live-common/blob/a9d706412456eb51815803a216e1ac8d11df8400/src/bridge/impl.ts#L72-L88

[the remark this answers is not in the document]
gre: check what we did for XTZ , we should do the same
it first deliver a versoin where libcore is still the default but the js is "experimental".

https://github.com/LedgerHQ/ledger-live-common/blob/8c1c8ba9edb5ec5a0dfe5e3eccff0efc677c5920/src/account/support.ts#L23

juan-cortes (approved):

src/families/bitcoin/js-synchronisation.ts
```diff
@@ -0,0 +1,378 @@
...
+      return "Native SegWit";
+
+    default:
+      return "Legacy";
+  }
+};
+
+// Map LL's currency ID to wallet-btc's Account.params.network
+const toWalletNetwork = (currencyId: string): "testnet" | "mainnet" => {
+  return ["bitcoin_testnet"].includes(currencyId) ? "testnet" : "mainnet";
```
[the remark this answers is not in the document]
gre: mmh bad idea to hardcode this here.
we should rely on the data of the currency. let me fix this.

src/families/bitcoin/types.ts
```diff
@@ -35,17 +42,116 @@ export type BitcoinOutputRaw = [
   number,
   number | null | undefined,
   string | null | undefined,
+  number,
```
gre, line 45: ok to me this is a problem. we are changing the type `BitcoinOutputRaw ` which is stored in app.json.  it also obviously impact libcore impl too.

gre (changes requested):
@haammar-ledger if you could help on one comment that I think is a breaking change.

src/families/bitcoin/js-synchronisation.ts
```diff
@@ -0,0 +1,384 @@
...
+  for (const output of tx.outputs) {
+    if (output.address) {
+      if (accountAddresses.includes(output.address)) {
+        accountOutputs.push(output);
+
+        if (changeAddresses.includes(output.address)) {
+          if (
+            (recipients.length === 0 &&
+              output.output_hash ===
+                tx.outputs[tx.outputs.length - 1].output_hash) || // Is that condition safe?
```
gre, line 140: comment to drop if doubt is cleared.

src/migrations/accounts.ts
```diff
@@ -16,6 +16,24 @@ function sameSwap(a: SwapOperation, b: SwapOperation) {
...
+        return a.id;
+      }
+      if (decoded.currencyId !== "bitcoin") {
+        return a.id;
+      }
+      return `js:2:${decoded.currencyId}:${a.freshAddress}:${decoded.derivationMode}`;
+    }
+    return a.id;
+  }
+  */
```
gre, line 35: please remove the code :)

src/families/bitcoin/js-signOperation.ts
```diff
@@ -0,0 +1,164 @@
...
+        const txInfo = await buildTransaction(account, transaction);
+        let senders: string[] = [];
+        let recipients: string[] = [];
+        let fee = new BigNumber(0);
+        // Maybe better not re-calculate these fields here, instead include them
+        // in Transaction type and set them in prepareTransaction?
+        await calculateFees({
+          account,
+          transaction,
+        }).then((res) => {
```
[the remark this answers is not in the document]
haammar-ledger: Actually dumb c/p from current impl :open_mouth: but you're right.
https://github.com/LedgerHQ/ledger-live-common/blob/d40619cfc2ffbddae8a10a8e2dc9d59c4d75a1cf/src/families/bitcoin/bridge/libcore.ts#L134

src/families/bitcoin/types.ts
haammar-ledger: Moved the new field to the end.

src/families/bitcoin/js-synchronisation.ts
```diff
@@ -0,0 +1,384 @@
...
+    blockHeight: utxo.block_height,
+    address: utxo.address,
+    value: new BigNumber(utxo.value),
+    rbf: utxo.rbf,
+    isChange: false, // wallet-btc limitation: doesn't provide it
+    path: "",
+  };
+};
+
+// wallet-btc limitation: returns all transactions twice (for each side of the tx)
```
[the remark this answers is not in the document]
haammar-ledger: Task is here https://ledgerhq.atlassian.net/browse/LL-7582 with related doc in confluence here https://ledgerhq.atlassian.net/wiki/spaces/WALLETCO/pages/3365732403/BTC-JS+-+Maintenance.
Good catch I updated the document.

gre (approved):

PR #1363, opened 2021-08-27
LL-6995 Add Platform exchange device actions and commands

juan-cortes:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-6832

Allows us to run the nonce generation and exchange validation commands independently, to be used with platform apps paradigm. There's a bunch of TODO comments that I need to address before I undraft it.

## Expectations


alacombe-ledger:
@pollastri-pierre @lcastillo-ledger what do you need to review this please? Thank you

PR #1365, opened 2021-08-30
Add Filecoin blockchain support

emmanuelm41:
## Context (issues, jira)
- Depends on https://github.com/LedgerHQ/ledgerjs/pull/651

## Description / Usage
Add support to filecoin cryptocurrency. The list of commands supported in CLI are 

```
balanceHistory
broadcast
derivation
estimateMaxSpendable
exportAccounts
generateTestTransaction
getAccountNetworkInfo
getAddress
getTransactionStatus
portfolio
receive
repl
send
signMessage
swap
sync
testDetectOpCollision
validRecipient
walletconnect
```

## Expectations


src/account/serialization.ts
```diff
@@ -135,7 +135,7 @@ export const toOperationRaw = (
     blockHeight,
     extra: e,
     date: date.toISOString(),
-    value: value.toString(),
+    value: value.toFixed(),
```
gre, line 138: could you explain the reason to switch? Thanks

src/bridge/jsHelpers.ts
```diff
@@ -286,8 +289,19 @@ export const makeScanAccounts =
             const stopAt = isIterableDerivationMode(derivationMode) ? 255 : 1;
             const startsAt = getDerivationModeStartsAt(derivationMode);
 
+            log(
+              "degub",
```
gre, line 293: small typo :)

src/derivation.ts
```diff
@@ -208,6 +208,10 @@ const modes = Object.freeze({
   polkadotbip44: {
     overridesDerivation: "44'/354'/<account>'/0'/<address>'",
   },
+  // FIXME Filecoin - Check if this is correct
```
gre, line 211: who need to check? thanks

src/families/filecoin/bridge/utils/addresses.ts
```diff
@@ -0,0 +1,150 @@
+import base32Decode from "base32-decode";
```
gre, line 1: some CI fail here.

src/families/filecoin/bridge/utils/serializer.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  // Adding byte sign
+  let signByte = "00";
+  if (v.lt(0)) {
+    signByte = "01";
+  }
+
+  if (v.toString() === "") {
+    // to test with null bigint
+    return Buffer.from(signByte, "hex");
```
gre, line 16: please import from "buffer", because "Buffer" may not be a global in all environments.

gre (commented):
robust integration so far.

- I notice some "FIXME" in the code but i guess it's a WIP integration (that we could plan to merge as step1)
- some comment on some changes in the general helper parts

src/account/serialization.ts
emmanuelm41: Yes, of course. I changed that one to avoid using scientific notation, as it mentions [here](https://github.com/MikeMcl/bignumber.js/#use).

image

src/families/filecoin/bridge/utils/serializer.ts
emmanuelm41: I am not sure about this change. Could you explain me more about it? I tried to look for buffer lib on the rest of the code but I could not find anything.

src/bridge/jsHelpers.ts
emmanuelm41: fixed!

src/account/serialization.ts
gre: wow good to know. thanks.

src/families/filecoin/bridge/utils/addresses.ts
emmanuelm41: I added the missing dependency on package.json

src/derivation.ts
emmanuelm41: This FIXME is not needed anymore!

emmanuelm41:
> robust integration so far.
> 
> * I notice some "FIXME" in the code but i guess it's a WIP integration (that we could plan to merge as step1)
> * some comment on some changes in the general helper parts

Thanks for your review. It is a WIP integration indeed. We wanted to let you check the integration in general while we make the last few changes.

src/families/filecoin/bridge/utils/serializer.ts
gre: Ok, it's probably fine for now. On my end, I assume Buffer is a legacy that we'll eventually replace by native arrays (but it's not easy) because it's only really available on Node.
It's not available on web, it's not available in React Native – unless we add a polyfill. Which we did, so it should work! https://github.com/LedgerHQ/ledger-live-mobile/blob/develop/src/polyfill.js#L3

but essentially `import { Buffer } from "buffer"` maybe be needed if we run that code on web.

emmanuelm41: I see. Now I understand it.

src/families/filecoin/bridge/account.ts
```diff
@@ -0,0 +1,307 @@
...
+    gasLimit: new BigNumber(0),
+    gasFeeCap: new BigNumber(0),
+    gasPremium: new BigNumber(0),
+    recipient: "",
+    useAllAmount: false,
+  };
+};
+
+const updateTransaction = (t: Transaction, patch: Transaction): Transaction => {
+  log("debug", "[updateTransaction] patching tx");
```
gre, line 56: please remove some of the tracing logs debug on this file, was probably interesting during development but it's better we try to keep it concise. (typically no need to log on hooks because we look it in different ways in LLD / no need to log "loading fees" because we see a log for the network call, etc..)

src/families/filecoin/bridge/account.ts
```diff
@@ -0,0 +1,307 @@
...
+    errors.recipient = new InvalidAddressBecauseDestinationIsAlsoSource();
+  else if (!validateAddress(recipient).isValid)
+    errors.recipient = new InvalidAddress();
+  else if (!validateAddress(address).isValid)
+    errors.sender = new InvalidAddress();
+
+  if (
+    gasFeeCap.toNumber() === 0 ||
+    gasPremium.toNumber() === 0 ||
+    gasLimit.toNumber() === 0
```
gre, line 87: maybe be more concise with `.eq(0)`

gre (commented):
minor comments. all good overall.

src/families/filecoin/bridge/account.ts
emmanuelm41: Great! Good improvement!

src/families/filecoin/bridge/account.ts
emmanuelm41: I have commented all log lines of that kind. I think they are useful for debugging, so I left them commented so that I can use them again if we have to make some fix.

emmanuelm41:
@gre I think the tests are failing because of ethereum tests use .toString on amount, and I changed that to .toFixed(). I guess that is happening [here](https://github.com/Zondax/ledger-live-common/blob/12f8258ea4811b5023fc16899221f45612180694/src/families/ethereum/modules/compound.ts#L100). Could you confirm me that? So I will change it.

gre:
```
  ● filecoin currency bridge › scanAccounts › filecoin seed 1

    Invariant Violation: no Ledger explorer for filecoin

      49 | ): LedgerExplorer => {
      50 |   const res = findCurrencyExplorer(currency);
    > 51 |   invariant(res, `no Ledger explorer for ${currency.id}`);
         |   ^
      52 |   return <LedgerExplorer>res;
      53 | };
      54 |

      at invariant (node_modules/invariant/invariant.js:40:15)
      at getCurrencyExplorer (src/api/Ledger.ts:51:3)
      at blockchainBaseURL (src/api/Ledger.ts:56:37)
      at getFilecoinURL (src/families/filecoin/bridge/utils/api.ts:26:19)
      ```
      
      looks like you use blockchainBaseURL from getFilecoinURL but as it's not defined in https://github.com/LedgerHQ/ledger-live-common/blob/12f8258ea4811b5023fc16899221f45612180694/src/api/explorerConfig/index.ts and not really hosted by Ledger today, it can't work, it sounds like you need to not use blockchainBaseURL but maybe can return an env you define in env.js
      
like for example we did for tezos https://github.com/LedgerHQ/ledger-live-common/blob/master/src/env.ts#L113-L117

File: src/families/filecoin/bridge/utils/api.ts [hunk 8]
```diff
@@ -0,0 +1,107 @@
...
+} from "./types";
+import network from "../../../../network";
+import { blockchainBaseURL } from "../../../../api/Ledger";
+import { getCryptoCurrencyById } from "../../../../currencies";
+import { getEnv } from "../../../../env";
+
+const root = "http://127.0.0.1:8888/blockchain/filecoin";
+
+const getFilecoinURL = (path?: string): string => {
+  if (getEnv("DEVELOPMENT_MODE")) return root + (path ? path : "");
```
gre, line 23: I would prefer not to introduce a "DEVELOPMENT_MODE" but instead have explicit envs. like a "API_FILECOIN_EXPLORER".

gre (changes requested):
as seen by the CI, something is not working in prod, I don't know if there is today a filecoin api but this would block the PR to be merged.
until then, please find some suggestion on what could be done to prepare this.
basically I think we don't want too abstract env like "DEVELOPMENT_MODE" but instead we prefer to have explicit flags or envs. in that case, it's a blockchain url for filecoin we could have.

src/families/filecoin/bridge/utils/api.ts
emmanuelm41: Perfect! I used that env var because I did not want to create a new one. But I can change it, no problem. I think your suggestion is better.

src/families/filecoin/test-dataset.ts
```diff
@@ -0,0 +1,219 @@
...
+                nonce: 100,
+                gasFeeCap: "0",
+                gasLimit: 100,
+                gasPremium: "10000",
+                recipient: ACCOUNT_1,
+                amount: "100000000",
+              }),
+              expectedStatus: {
+                errors: {
+                  gas: new FeeNotLoaded(),
```
gre, line 129: it seems you expected "FeeNotLoaded" error in this case? it seems the CI sees there are no error in this case.

src/families/filecoin/test-dataset.ts
```diff
@@ -0,0 +1,219 @@
...
+    filecoin: {
+      FIXME_ignoreAccountFields: [],
+      scanAccounts: [
+        {
+          name: "filecoin seed 1",
+          apdus: `
+          => 0600000000
+          <= 0000120500311000049000
+          => 06010000142c000080cd010080000000800000000000000000
+          <= 041e10b3a453db1e7324cd37e78820d7d150c13ba3bf784be204c91afe495816a19a836e85cb89b0e4ff36d06a71a9ca02947de79e16e66dacc645e46dcdf7d9091501cdf9b117d00fa23e938e52feb25da9a32e4ecdd02966317a78343363663671623672643565346f6b6c376c65786e6a756d786535746f716a3676747233699000
```
gre, line 29: Today, I see the CI fails on 
```
    > 14 |     store.ensureQueueEmpty();
```

and

```
    65535 - RecordStoreQueueEmpty: EOF: no more APDU to replay
```

which may indicate a problem around here.

---

you may want to try this with a static seed:

```
ledger-live generateTestScanAccounts -c filecoin
```

locally with your device and will generate the exact APDUs that shall be "stable".

src/families/filecoin/test-dataset.ts
```diff
@@ -0,0 +1,219 @@
...
+          `,
+        },
+      ],
+      accounts: [
+        {
```
gre, line 34: if you think "balance is sum of ops" can't be true, you can add here a

```
          FIXME_tests: [
            "balance is sum of ops",
          ],
```

like done at https://github.com/LedgerHQ/ledger-live-common/blob/9dbfe82ca0511b197d779924ec52c2efa36cb122/src/families/tezos/test-dataset.ts#L63

for CI to ignore this to test.

src/families/filecoin/test-dataset.ts
```diff
@@ -0,0 +1,219 @@
...
+                nonce: 100,
+                gasFeeCap: "1000",
+                gasLimit: 10,
+                gasPremium: "10000",
+                recipient: ACCOUNT_1,
+                amount: "10000000000000",
+              }),
+              expectedStatus: {
+                errors: {
+                  amount: new NotEnoughBalance(),
```
gre, line 169: one test also seem to see a problem that the transaction you express is not > the account balance or that there are not error happening. 

you may want to check if this happens when you do:

```
ledger-live getTransactionStatus --id filecoin:t1d2xrzcslx7xlbbylc5c3d5lvandqw4iwl6epxba --amount 100000 --recipient t1d2xrzcslx7xlbbylc5c3d5lvandqw4iwl6epxba
```

gre (commented):
some hot feedbacks regarding what I saw happening on the CI.

gre (approved):

gre (approved):
this work is passing. let's merge it! and we can have more iteration in further PRs.

PR #1369, opened 2021-09-01
LL-5352 Hardcode removal of Bitcoin dependency for Decred

juan-cortes:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-5352

## Description / Usage
There is an issue on backend side where Decred is listed as depending directly on the Bitcoin app, this means that we are currently blocking installations of the Decred app for users that don't have enough space to have the Bitcoin app too. This pr addresses the issue by hardcoding the removal of the dependency, but it's **worth noting that the fix should be done on backend** at the app store level and this piece of code should be removed once that's confirmed to be the case.

image
Only Decred without Bitcoin installed, correct storage size.

## Expectations


src/apps/mock.ts
```diff
@@ -92,7 +92,11 @@ export function mockListAppsResult(
     .map((a) => a.trim())
     .filter(Boolean)
     .map((name, i) => {
-      const dependencies = getDependencies(name);
+      // FIXME LL-5352 until backend fixes the mistake, remove the dependencies for Decred.
+      const dependencies =
+        name === "Decred" || name === "Decred Testnet"
```
gre, line 97: please put these in a whitelist object to avoid the double code.

gre (changes requested):

src/apps/mock.ts
juan-cortes: Sure, I'll see what I can do boss

juan-cortes: Putting comments allows the reviewer to understand the reasoning for the development, it's all part of the plan 🧠

PR #1383, opened 2021-09-13
Provide Currency to wallet-btc:isValidAddress()

kallerosenbaum:
## Context (issues, jira)

LL-7116

## Description / Usage

The method signature `isValidAddress()` in project `wallet-btc` has changed to include an optional argument of type Currency. This PR adds that argument to the call to `isValidAddress()`.

If the provided argument is not an actual Currency literal type (as defined in `wallet-btc`), an error, eg "Currency 'ontology' doesn't exist" will be shown.

This change is dependent on unreleased wallet-btc code. So before we merge this, we need to merge and release the PR in `wallet-btc`: https://github.com/LedgerHQ/wallet-btc/pull/53

### DIFF
--- a/src/families/bitcoin/logic.ts
+++ b/src/families/bitcoin/logic.ts
@@ -61,9 +61,9 @@ export const isValidRecipient = async (params: {
   try {
     // Optimistically assume params.currency.id is an actual Currency
     valid = isValidAddress(params.recipient, <Currency>params.currency.id);
-  } catch (e) {
+  } catch (e: any) {
     // isValidAddress() will throw Error if c is not an actual Currency
-    return Promise.reject(new CurrencyNotSupported(e.message));
+    return Promise.reject(new InvalidAddress(e.message));
   }
   if (!valid) {
     return Promise.reject(
--- a/src/families/bitcoin/logic.ts
+++ b/src/families/bitcoin/logic.ts
@@ -1,7 +1,11 @@
 import cashaddr from "cashaddrjs";
 import bchaddr from "bchaddrjs";
-import { isValidAddress } from "@ledgerhq/wallet-btc";
-import { RecipientRequired, InvalidAddress } from "@ledgerhq/errors";
+import { Currency, isValidAddress } from "@ledgerhq/wallet-btc";
+import {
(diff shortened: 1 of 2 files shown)


src/families/bitcoin/logic.ts
```diff
@@ -52,7 +56,15 @@ export const isValidRecipient = async (params: {
...
   }
-  const valid = isValidAddress(params.recipient);
+
+  let valid: boolean;
+  try {
+    // Optimistically assume params.currency.id is an actual Currency
+    valid = isValidAddress(params.recipient, <Currency>params.currency.id);
+  } catch (e) {
+    // isValidAddress() will throw Error if c is not an actual Currency
+    return Promise.reject(new CurrencyNotSupported(e.message));
```
haammar-ledger, line 66: Actually you can use `InvalidAddress` here, because it will translate appropriately on the frontend, see [here](https://github.com/LedgerHQ/ledger-live-desktop/blob/b116d532f3dc4f919ad98042b48ace8f135ec715/static/i18n/en/app.json#L3688-L3690),
where `CurrencyNotSupported`'s [translation](https://github.com/LedgerHQ/ledger-live-desktop/blob/b116d532f3dc4f919ad98042b48ace8f135ec715/static/i18n/en/app.json#L3433-L3435) would be misleading.

ghost (changes requested):

src/families/bitcoin/logic.ts
```diff
@@ -52,7 +56,15 @@ export const isValidRecipient = async (params: {
   if (!params.recipient) {
     return Promise.reject(new RecipientRequired(""));
   }
-  const valid = isValidAddress(params.recipient);
+
+  let valid: boolean;
+  try {
+    // Optimistically assume params.currency.id is an actual Currency
+    valid = isValidAddress(params.recipient, <Currency>params.currency.id);
+  } catch (e) {
```
haammar-ledger, line 64: I think TypeScript will require you to use `e: any` here.

valpinkman (approved):

ghost (approved):

PR #1385, opened 2021-09-14
Introduce initial support for Hedera

mehcode:
## Description / Usage

- Support for scanning and synching hedera accounts
- No support yet for sending or receiving hbar

## Expectations


src/data/icons/svg/HBAR.svg
```diff
@@ -0,0 +1,25 @@
+<?xml version="1.0" encoding="UTF-8" standalone="no"?>
+<svg
+   id="Layer_1"
```
gre, line 3: we need cleaner .svg that don't have "id" and "defs" in it, because it wouldn't work when converted to components (like React Native components)

src/families/hedera/api/mirror.ts
```diff
@@ -0,0 +1,108 @@
...
+import { getAccountBalance } from "./network";
+import { Operation, OperationType } from "../../../types";
+import { serializeSignedTransaction } from "@ledgerhq/live-app-sdk";
+import { encodeOperationId } from "../../../operation";
+
+const fetch = (path, query = {}) =>
+  network({
+    type: "get",
+    url: URL.format({
+      pathname: `https://mainnet-public.mirrornode.hedera.com/api/v1${path}`,
```
gre, line 13: the url domain left part need to be an env in env.js

src/families/hedera/js-scanAccounts.ts
```diff
@@ -0,0 +1,126 @@
...
+}: {
+  currency: CryptoCurrency;
+  deviceId: DeviceId;
+  scheme?: DerivationMode | null | undefined;
+  syncConfig: SyncConfig;
+  preferredNewAccountScheme?: DerivationMode;
+}): Observable<ScanAccountEvent> {
+  var derivationMode = scheme ?? getDerivationModesForCurrency(currency)[0];
+
+  return new Observable((o) => {
```
gre, line 44: In future, it will be preferrable to use jsHelpers which should help you generalizing the 'scanAccounts' and 'sync'. Is there any reason it was not used yet?

src/types/account.ts
```diff
@@ -133,6 +133,9 @@ export type Account = {
   // the iterated number to derive the account in a given derivationMode config
   // in context of bip44, it would be the account field of bip44 ( m/purpose'/cointype'/account' )
   index: number;
+  // a "virtual" address is a non-verified address
+  // setting this triggers the receive flow to show a warning that it can't verify the address
+  virtualAddress?: string;
```
gre, line 138: adding new fields at top level is costly. If there is really a need to have this, we can discuss, but generally speaking we would just introduce a namespaced `hederaResources` object (like for tronResources, bitcoinResources, etc..)

gre (commented):
hot feedbacks on this first iteration.

src/data/icons/svg/HBAR.svg
FabriceDautriat: @gre we decided to have it designed internally
(cc @rbeyon-ledger )

gre (changes requested):
some feedback on our side.
do you think it's also possible to add some minimalistic tests? we can guide you through it if necessary, but see https://developers.ledger.com/docs/coin/run-ledger-live/

it's necessary that our coins are tested so we can have non regression automated from the beginning.

src/families/hedera/utils.ts
```diff
@@ -0,0 +1,29 @@
...
+
+export function calculateAmount({
+  account,
+  transaction,
+}: {
+  account: Account;
+  transaction: Transaction;
+}): { amount: BigNumber; estimatedFees: BigNumber; totalSpent: BigNumber } {
+  // NOTE: Hedera declares stable fees in USD
+  //       If we can get the current USD/HBAR price here..
```
haammar-ledger, line 13: Question for @gre: do you think we can get the countervalue here?

src/families/hedera/js-estimateMaxSpendable.ts
```diff
@@ -0,0 +1,28 @@
+import BigNumber from "bignumber.js";
+import type { Account, AccountLike } from "../../types";
+import type { Transaction } from "./types";
+
+export default function estimateMaxSpendable({
```
haammar-ledger, line 5: This code is very similar to the `calculateAmount` function in `utils.ts`, can you maybe use it there?

src/families/hedera/js-receive.ts
```diff
@@ -0,0 +1,65 @@
+import { WrongDeviceForAccount } from "@ledgerhq/errors";
+import { Observable } from "rxjs";
+import { close, open } from "../../hw";
+import { Account, isSegwitDerivationMode } from "../../types";
+import getAddress from "../../hw/getAddress";
+
+export default function receive(
```
haammar-ledger, line 7: Same concern as for the scan & sync account, the default receive should be used preferably... but it's on us to help you find a way to fit our model.

src/families/hedera/js-transaction.ts
```diff
@@ -0,0 +1,46 @@
...
+/**
+ * Update a base property of the transaction.
+ *
+ * @returns  {Transaction}
+ */
+export function updateTransaction(
+  transaction: Transaction,
+  patch: Partial<Transaction>
+): Transaction {
+  return { ...transaction, ...patch } as unknown as Transaction;
```
haammar-ledger, line 28: I think you don't need this double cast?

src/families/hedera/js-sync.ts
```diff
@@ -0,0 +1,49 @@
+import { Observable } from "rxjs";
+import { Account, SyncConfig } from "../../types";
+import { getOperationsForAccount } from "./api/mirror";
+import { getAccountBalance } from "./api/network";
+
+export default function sync(
```
haammar-ledger, line 6: This implementation is not optimal as the full list of operations is replaced every time.
You can consider using `mergeOps`, see examples in other coin families.

There may be other concerns with not using the generic helpers, as already raised by @gre. We need to discuss and help you on that topic.

src/families/hedera/js-getTransactionStatus.ts
```diff
@@ -0,0 +1,55 @@
...
+  } else {
+    const senderAccountId = hedera.AccountId.fromString(account.seedIdentifier);
+
+    try {
+      const recipientAccountId = hedera.AccountId.fromString(
+        transaction.recipient
+      );
+
+      if (senderAccountId.equals(recipientAccountId)) {
+        errors.recipient = new InvalidAddressBecauseDestinationIsAlsoSource("");
```
haammar-ledger, line 30: Is it forbidden to self-send with Hedera?

src/derivation.ts
```diff
@@ -221,6 +221,9 @@ const modes = Object.freeze({
   filecoin: {
     overridesDerivation: "44'/461'/0'/0/<account>",
   },
+  hederaBip44: {
+    overridesDerivation: "44/3030/0/0/<account>",
```
gre, line 225: 44'/3030' ?

src/families/hedera/hw-app-hedera/index.ts
```diff
@@ -0,0 +1,103 @@
...
+  SIGN_TRANSACTION: 0x04,
+};
+
+const STATUS = {
+  OK: 0x9000,
+  USER_CANCEL: 0x6985,
+};
+
+/** Hedera BOLOS API */
+export default class Hedera {
```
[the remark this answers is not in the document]
gre: i think this can be done in second time once the work is validated into our source code. topic for us later. not blocker.

src/env.ts
```diff
@@ -125,6 +125,11 @@ const envDefinitions = {
     def: "https://tron.coin.ledger.com",
     desc: "proxy url for trongrid API",
   },
+  API_HEDERA_MIRROR: {
+    def: "https://mainnet-public.mirrornode.hedera.com",
```
steven-sheehy, line 129: This should be updated to use the unthrottled mirror node:

```suggestion
    def: "https://mainnet.mirrornode.hedera.com",
```

steven-sheehy (changes requested):

src/families/hedera/js-getTransactionStatus.ts
jackthta: To the best of my knowledge, yes it is forbidden to self-send.

src/families/hedera/api/mirror.ts
```diff
@@ -0,0 +1,136 @@
...
+    // NOTE: earlier addresses are the "fee" addresses
+    recipients.reverse();
+    senders.reverse();
+
+    operations.push({
+      value,
+      date: timestamp,
+      blockHeight: 1,
+      // NOTE: not sure what to put here, there are no "blocks" in hedera
+      blockHash: raw.transaction_hash,
```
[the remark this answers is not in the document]
jackthta: 👍🏻

src/families/hedera/js-estimateMaxSpendable.ts
jackthta: 👍🏻

src/families/hedera/js-transaction.ts
jackthta: 👍🏻

src/families/hedera/api/network.ts
```diff
@@ -0,0 +1,54 @@
...
+
+  return {
+    balance: accountBalance.hbars.to(hedera.HbarUnit.Tinybar),
+  };
+}
+
+let _hederaClient: hedera.Client | null = null;
+
+function getClient(): hedera.Client {
+  _hederaClient ??= hedera.Client.forMainnet().setMaxNodesPerTransaction(1);
```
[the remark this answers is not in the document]
jackthta: The client may be targeting a range from node `0.0.3` to `0.0.20` and I don't believe there is a way to provide a custom URL.

I may need an expert opinion to confirm my assumptions though.

src/families/hedera/api/mirror.ts
```diff
@@ -0,0 +1,136 @@
...
+
+export interface Account {
+  accountId: AccountId;
+  balance: BigNumber;
+}
+
+export async function getAccountsForPublicKey(
+  publicKey: string
+): Promise<Account[]> {
+  const r = await fetch("/accounts", { "account.publicKey": publicKey });
```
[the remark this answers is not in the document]
jackthta: I just want to double check if I'm understanding this correctly, but would it be something like this?

`const r = await fetch("/accounts", { "account.publicKey": publicKey, "balance": false });`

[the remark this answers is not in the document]
steven-sheehy: Correct. We have [swagger docs](https://mainnet.mirrornode.hedera.com/api/v1/docs/#/) you can also check. Setting this param will speed up the response time.

Could also put a `limit` param on it if you know you don't care about multiple accounts being associated with the same key. That would also speed it up.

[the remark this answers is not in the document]
jackthta: Thank you! I will make this edit 👍🏻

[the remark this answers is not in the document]
haammar-ledger: It seems from our tests that the `account.balance=false` path param currently doesn't work?

```
$ curl "https://mainnet-public.mirrornode.hedera.com/api/v1/accounts?account.publicKey=9e92a312233d5fd6b5a723875aeea2cea81a8e48ffc00341cff6dffcfd3ab7f2&account.balance=false"
{"_status":{"messages":[{"message":"Invalid parameter: account.balance"}]}}
```

```
$ curl "https://mainnet-public.mirrornode.hedera.com/api/v1/accounts?account.publicKey=9e92a312233d5fd6b5a723875aeea2cea81a8e48ffc00341cff6dffcfd3ab7f2"
{"accounts":[],"links":{"next":null}}
```

src/cross.ts
```diff
@@ -20,6 +25,8 @@ export type AccountData = {
   name: string;
   index: number;
   balance: string;
+  // chain-specific resources
+  hederaResources?: HederaResourcesRaw;
```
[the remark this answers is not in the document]
haammar-ledger: @jackthta According to my understanding, the `accountId` could be retrieved on live-mobile side, in the same way it is retrieved on live-desktop, from the public key, which should be already provided in the AccountData. If you can please confirm, this would be a way to remove the need for `hederaResources` here.

src/families/hedera/api/network.ts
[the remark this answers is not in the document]
haammar-ledger: Topic still pending, we should discuss it during our sync meeting. Either a reverse proxy isn't considered mandatory on this part, or you could maybe target the network directly, same way as it's done in `mirror.ts`.

src/families/hedera/js-sync.ts
haammar-ledger: To follow-up on this topic, I _think_ we are ok with having custom `sync` and `scanAccount` logics, to be confirmed by @gre.

But at minimum we need to make sure the performance is acceptable. The optimization issue that I mentioned above should be addressed.

src/families/hedera/api/mirror.ts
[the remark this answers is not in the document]
steven-sheehy: It's just `balance`, not `account.balance`. Again, please reference the [swagger](https://mainnet.mirrornode.hedera.com/api/v1/docs/#/) site.

src/families/hedera/js-sync.ts
haammar-ledger: Digging a bit more, I'm wondering if there is a way to use the standard mechanism, by using `ChildAccount`... I need to think more about it.

src/families/hedera/api/mirror.ts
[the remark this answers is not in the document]
jackthta: That was my mistake, I looked at the swagger doc and assumed by context that it would be `account.balance`. It should be fixed now.

src/env.ts
jackthta: Changing the mirror api url to `https://mainnet.mirrornode.hedera.com` makes the app stall indefinitely on the synchronizing step (where the url is queried to check for account associated with the ledger device). I changed it back to the previous url for the time being until this is addressed.

src/families/hedera/api/network.ts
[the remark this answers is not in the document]
jackthta: Apparently we _can_ set a proxy via `client.setNetwork()`, but I wanted to ask whether the proxy has been set up? If so, can it be provided so I can try it out to see if it can work please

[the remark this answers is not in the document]
haammar-ledger: Cool, yes we now have a proxy up and running : `https://hedera.coin.ledger.com`

src/env.ts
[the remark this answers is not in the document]
jackthta: Using the ledger reverse-proxy here works 👍🏻. I'll push up the edit once I figure out how to set the reverse-proxy in `network.ts` (running into some difficulty using the aforementioned `client.setNetwork()` function).

src/families/hedera/js-sync.ts
haammar-ledger: Never mind previous comment, probably a dead-end.

src/families/hedera/utils.ts
haammar-ledger: As discussed during our sync meeting, please keep displaying fees in coin unit, as Ledger Live isn't designed to display fees in fiat unit.

jackthta: 👍🏻

src/derivation.ts
jackthta: I'm not sure why it's the case, but `44'/3030'/0/0/<account>` doesn't seem to work or when trying to add an account, will not return the associated list of accounts. `44/3030'/0/0/<account>` _does_ work, but I'm not sure how much of an improvement it would be over what's currently there.

src/families/hedera/js-sync.ts
jackthta: 👍🏼

src/env.ts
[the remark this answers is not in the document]
jackthta: 👋🏻 @haammar-ledger

I've tried setting the `hedera.Client` to target `https://hedera.coin.ledger.com`, but it doesn't seem to work. I'm not sure why, but perhaps because `hedera.Client.forMainnet()` sets into the client a network of nodes (which are the mainnet nodes). 

This is the structure/how it looks:
> MAINNET: {
        "35.237.200.180:50211": new AccountId(3),
        "35.186.191.247:50211": new AccountId(4),
        "35.192.2.25:50211": new AccountId(5),
        "35.199.161.108:50211": new AccountId(6),
        "35.203.82.240:50211": new AccountId(7),
        "35.236.5.219:50211": new AccountId(8),
        "35.197.192.225:50211": new AccountId(9),
        "35.242.233.154:50211": new AccountId(10),
        "35.240.118.96:50211": new AccountId(11),
        "35.204.86.32:50211": new AccountId(12),
        "35.234.132.107:50211": new AccountId(13),
        "35.236.2.27:50211": new AccountId(14),
        "35.228.11.53:50211": new AccountId(15),
        "34.91.181.183:50211": new AccountId(16),
        "34.86.212.247:50211": new AccountId(17),
        "172.105.247.67:50211": new AccountId(18),
        "34.89.87.138:50211": new AccountId(19),
        "34.82.78.255:50211": new AccountId(20),
    }

👆🏻 [Source](https://github.com/hashgraph/hedera-sdk-js/blob/f48dcb2bb2a596fae5fca7eea6c76a862a446ee4/src/client/NodeClient.js#L37-L56)

I'm having a difficult time seeing how to target the network directly.. but if you have any suggestions, please do let me know! Otherwise, I hope it will be okay to leave it as is.

src/families/hedera/utils.ts
```diff
@@ -0,0 +1,34 @@
...
+}> {
+  // NOTE: Hedera declares stable fees in USD
+  //       If we can get the current USD/HBAR price here..
+  //       > transfer fee is 0.0001 USD
+  const estimatedFees = new BigNumber("83300"); // 0.000833 ℏ (as of 2021-09-20)
+
+  const amount =
+    transaction.useAllAmount == true
+      ? // as fees are based on a currency conversion, we stay
+        // on the safe side here and double the estimate for "max spendable"
```
haammar-ledger, line 25: obsolete comment, please remove

haammar-ledger:
Replaced with https://github.com/LedgerHQ/ledger-live-common/pull/1887

PR #1386, opened 2021-09-15
update testnet urls for crypto.com

thomas-nguy:
## Context (issues, jira)

Fixing integration tests

## Expectations


gre:
thanks. i'll work with QA to restore some dataset with our testing seed.

PR #1389, opened 2021-09-16
Elrond fees to high warning

AlexandruPislariu:
## Expectations


gre (approved):

gre:
@AlexandruPislariu would be good to be in sync with our upstream commits, anyway, not a problem as i squash&merged here. but more for the list of commits display :)

PR #1397, opened 2021-09-20
LL-7257 - tezos fix send max

machard:
https://ledgerhq.atlassian.net/browse/LL-7257


gre (changes requested):
all good. just some lint issue.

PR #1404, opened 2021-09-21
LL-7357 Merge wallet-btc into Live Common

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-7357

## Expectations


package.json
```diff
@@ -38,6 +38,9 @@
     "react-dom": ">=16",
     "react-redux": "7"
   },
+  "react-native": {
+    "https": false
+  },
```
haammar-ledger, line 43: @gre We needed this in wallet-btc to be compatible with react-native.
Hope it's ok to have it in live-common.

PR #1405, opened 2021-09-22
LL 7340 & LL-6671 - Sync ERC721 & ERC1155 NFT

ghost:
## Context (issues, jira)

[LL-6671](https://ledgerhq.atlassian.net/browse/LL-6671)
[LL-7340](https://ledgerhq.atlassian.net/browse/LL-7340)

## Context (issues, jira)
This PR is a proposal for the implementation of NFTs and synchronization of ERC721 & ERC1155 tokens in live-common.

### _**This PR is done using the staging server for the `/transactions` call and a mock for the NFTs' metadata**_
Source for the metadata mock: https://github.com/klambert-ledger/mock-nft-metadata-api

## Description / Usage

Account should now have a NFT property filled with owned NFTs:

[Code Preview](https://carbon.now.sh/?bg=rgba%28171%2C+184%2C+195%2C+1%29&t=seti&wt=none&l=application%2Fjson&ds=true&dsyoff=20px&dsblur=68px&wc=true&wa=true&pv=56px&ph=56px&ln=false&fl=1&fm=Hack&fs=14px&lh=133%25&si=false&es=2x&wm=false&code=%257B%250A%2520%2520%2522type%2522%253A%2520%2522Account%2522%252C%250A%2520%2520%2522id%2522%253A%2520%2522js%253A2%253Aethereum%253A0x689c61783100654dc72f2825b8e6050ee65a5419%253A%2522%252C%250A%2520%2520%252F%252F%2520...%250A%2520%2520%2522NFT%2522%253A%2520%255B%250A%2520%2520%2520%2520%257B%250A%2520%2520%2520%2520%2520%2520%2522id%2522%253A%2520%2522js%253A2%253Aethereum%253A0x689c61783100654dc72f2825b8e6050ee65a5419%253A%252B0x1a92f7381b9f03921564a437210bb9396471050c%252B7825-0xe2853d8781de8036df03d2a43e0c4cb14d6f4c9c4b4c458475f7fddcf439e6a7-IN%2522%252C%250A%2520%2520%2520%2520%2520%2520%2522amount%2522%253A%2520%25221%2522%252C%250A%2520%2520%2520%2520%2520%2520%2522tokenId%2522%253A%2520%25227825%2522%252C%250A%2520%2520%2520%2520%2520%2520%2522nftName%2522%253A%2520%2522Cool%2520Cat%2520%25237825%2522%252C%250A%2520%2520%2520%2520%2520%2520%2522picture%2522%253A%2520%2522https%253A%252F%252Fipfs.io%252Fipfs%252FQmUEb1royzdUP9WYp26nBn96X4KSKi3EGGoQjy2PpDpsAM%2522%252C%250A%2520%2520%2520%2520%2520%2520%2522description%2522%253A%2520%2522Cool%2520Cats%2520is%2520a%2520collection%2520of%25209%252C999%2520randomly%2520generated%2520and%2520stylistically%2520curated%2520NFTs%2520that%2520exist%2520on%2520the%2520Ethereum%2520Blockchain.%2520Cool%2520Cat%2520holders%2520can%2520participate%2520in%2520exclusive%2520events%2520such%2520as%2520NFT%2520claims%252C%2520raffles%252C%2520community%2520giveaways%252C%2520and%2520more.%2520Remember%252C%2520all%2520cats%2520are%2520cool%252C%2520but%2520some%2520are%2520cooler%2520than%2520others.%2522%252C%250A%2520%2520%2520%2520%2520%2520%2522properties%2522%253A%2520%255B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522key%2522%253A%2520%2522body%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522value%2522%253A%2520%2522blue%2520cat%2520skin%2522%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257D%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522key%2522%253A%2520%2522face%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522value%2522%253A%2520%2522angry%2520cute%2522%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257D%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522key%2522%253A%2520%2522hats%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522value%2522%253A%2520%2522astro%2522%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257D%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522key%2522%253A%2520%2522shirt%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522value%2522%253A%2520%2522buttondown%2520green%2522%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257D%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522key%2522%253A%2520%2522tier%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%2522value%2522%253A%2520%2522wild_1%2522%250A%2520%2520%2520%2520%2520%2520%2520%2520%2520%2520%257D%250A%2520%2520%2520%2520%2520%2520%255D%252C%250A%2520%2520%2520%2520%2520%2520%2522collection%2522%253A%2520%257B%250A%2520%2520%2520%2520%2520%2520%2520%2520%2522contract%2522%253A%2520%25220x1a92f7381b9f03921564a437210bb9396471050c%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2522tokenName%2522%253A%2520%2522Cool%2520Cats%2522%252C%250A%2520%2520%2520%2520%2520%2520%2520%2520%2522standard%2522%253A%2520%2522ERC721%2522%250A%2520%2520%2520%2520%2520%2520%257D%250A%2520%2520%2520%2520%257D%252C%250A%2520%2520%255D%250A%257D)
## Expectations


klambert-ledger:
This PR is a **proposal**, and it needs some insights for the architecture and need some help to have a more coherent testing  since right now I'm only comparing our results with what OpenSea's API is responding (which doesn't contain the _amount_ of ERC 1155 for example).

src/api/Ethereum.ts
```diff
@@ -178,6 +203,26 @@ export const apiForCurrency = (currency: CryptoCurrency): API => {
...
+      const { data }: { data: NFTMetadata[] } = await network({
+        method: "POST",
+        /** @important FIXME: metadata are mocked while we are waiting for the API */
+        url: "https://mock-nft-metadata-api.vercel.app/v1/chains/eth/contracts/tokens/infos",
+        data: input,
+      });
+
+      return data.reduce(
+        (acc, curr) =>
+          _.setWith(
```
LFBarreto, line 216: could you not use lodash for simple things like this :grin: 
But if you really need it cause its easier for you try and import just the function you use such as this
```javascript
import setWith from "lodash/setWith";
```

src/families/ethereum/synchronisation.ts
```diff
@@ -159,6 +178,7 @@ export const getAccountShape: GetAccountShape = async (
     lastSyncDate: new Date(),
     balanceHistory: undefined,
     syncHash,
+    NFT,
```
LFBarreto, line 181: naming would be a bit off keep it camel case if you can.

src/families/ethereum/synchronisation.ts
```diff
@@ -296,6 +323,136 @@ const txToOps =
...
+          if (!sending && !receiving) {
+            return [];
+          }
+
+          const all: NFTOperation[] = [];
+
+          if (sending) {
+            const type = "OUT";
+            all.push({
+              id: `${nftId}-${hash}-${type}`,
```
LFBarreto, line 348: make sure the ending op Id includes the accountID as we do and keep the order maybe like 
accountID hash nftId type
So we can still parse it easily or maybe nftId at the end to debate :thinking:

src/families/ethereum/synchronisation.ts
```diff
@@ -362,6 +523,7 @@ const txToOps =
         extra: {},
         internalOperations,
         subOperations,
+        nftOperations,
```
LFBarreto, line 526: if these are considered as contract operations shouldnt they be put/merged inside the subOperations maybe :thinking: 
it doesnt make much sense to me to separate the two

src/families/ethereum/test-specifics.test.ts
```diff
@@ -0,0 +1,110 @@
...
+        .sync(account, {
+          paginationConfig: {},
+          blacklistedTokenIds,
+        })
+        .pipe(reduce((a, f) => f(a), account))
+        .toPromise();
+
+      const getPaginatedOpenSeaRes = async (assets = []) => {
+        const { data } = await axios.get(
+          `https://api.opensea.io/api/v1/assets?offset=${assets.length}&limit=50&owner=${nftAccount.freshAddress}`
```
LFBarreto, line 65: we shouldn't rely on an api call for a test dataset :no_entry_sign:

src/families/ethereum/test-specifics.ts
```diff
@@ -0,0 +1,110 @@
...
+          .sync(account, {
+            paginationConfig: {},
+            blacklistedTokenIds,
+          })
+          .pipe(reduce((a, f) => f(a), account))
+          .toPromise();
+
+        const getPaginatedOpenSeaRes = async (assets = []) => {
+          const { data } = await axios.get(
+            `https://api.opensea.io/api/v1/assets?offset=${assets.length}&limit=50&owner=${nftAccount.freshAddress}`
```
LFBarreto, line 64: same here no api call within dataset creation either you copy paste what you need to compare or you mock this

src/families/ethereum/test-specifics.ts
```diff
@@ -0,0 +1,110 @@
...
+
+        const ourList =
+          synced.NFT?.map((n) => ({
+            contract: n.collection.contract,
+            tokenId: n.tokenId,
+          })) || [];
+
+        expect(ourList?.length).toEqual(openSeaList?.length);
+        expect(
+          _.isEqual(
```
LFBarreto, line 102: import isEqual from "lodash/isEqual";
if you really need to compare this in JS but i would use jest for it :grin:

src/types/account.ts
```diff
@@ -214,6 +215,8 @@ export type Account = {
   swapHistory: SwapOperation[];
   // Hash used to discard tx history on sync
   syncHash?: string;
+  // Array of NFTs computed by diffing NFTOperations
+  NFT?: NFT[];
```
LFBarreto, line 219: would you be so kind and rename this in camel case for the variable :grin: 
you can keep use pascal case for types too

LFBarreto (changes requested):
Some stuff you could do to improve, it looks fine by me just different from what should be done i think :+1:

src/families/ethereum/test-specifics.ts
LFBarreto: https://jestjs.io/docs/expect#toequalvalue

src/api/Ethereum.ts
klambert-ledger: Yeah that's just a one line solution (if prettier didn't force the new lines) to the infamous
```javascript
data.reduce((acc, curr) => {
  const contract = curr?.contract?.toLowerCase();
  const tokenId = curr?.tokenId?.toLowerCase();

  if (!acc[contract]) {
    acc[contract] = {};
  }

  acc[contract][tokenId] = curr;
}, {});
```

I can put that if you prefer 😄 

But yeah if you're ok with Lodash, I'll use the tree shaking 👌

src/families/ethereum/synchronisation.ts
klambert-ledger: Ok will do, I just followed the example given by @gre a month ago 👍

src/families/ethereum/synchronisation.ts
klambert-ledger: Will do 👌

src/families/ethereum/synchronisation.ts
klambert-ledger: As discussed already with you on slack, I think we need that in order to be non breaking since subOps and nftOps are different types. But I really don't enough about the usage in LLD/LLM so it might be _safe—ish_ to concat them all. I need to dig a little bit more into that 🤷‍♂️

src/families/ethereum/test-specifics.test.ts
klambert-ledger: This file is just for Juan and Hugo to test what I did, but yeah that's the same file as `test-specifics.ts` just under.

src/families/ethereum/test-specifics.ts
klambert-ledger: Yeah that's what my comment (https://github.com/LedgerHQ/ledger-live-common/pull/1405#issuecomment-925025413) was about, this is an ugly test, and I need some help to create a decent one. 😞

src/families/ethereum/test-specifics.ts
klambert-ledger: This test will disappear when I do a better integration test, but yeah forgot to remove the _.isEqual that I used to console.log after having to find the _.sortBy to make Jest accept to compare 2 arrays' content... 

Will do the tree shaking for _.sortBy and remove the _.isEqual 👌

juan-cortes:
My understanding was that the review wasn't so much about code specifics like but more about the architectural approach and how it fits in the schema of our codebase. NFTs are by definition a token and if we follow the convention we have in live-common they actually would fit in the `TokenAccount` definition, making them be inside the parent `Account`.

image


Keeping them as `TokenAccount` allows us to keep track of the movements we had for them. The only peculiarity is that the `ERC-721` ones would only have a balance of 1 and there would only be the incoming transaction, but we could actually have the outgoing one if we wanted to track the tokens we once had. It's also compatible with the `tokenId` structure we keep where `ethereum/erc721/tokenIdentifier`

We would need to build a list of `TokenCurrency` on sync dynamically, as opposed to based on a hardcoded list like we do for ERC-20. This dynamic list would come from the operations for the user and would be used to know what they can send, I don't think we need to support a list of what we can receive since that's an immense amount of different tokens.

The `TokenCurrency` type already was thought of with the concept of a `tokenType` in mind, we currently use it for erc20, trc10 and trc20 from what I see in the code. If we need some extra fields they should added to this type instead of creating a whole new monster.

This fits 100% the UI we saw on figma, and it would require minimal code changes (asset distribution would need to filter by tokenType to not show NFTs on the table for example) but I might be missing something.

------

As for how to handle the sync and the sending we already talked about it but I think it should be a module inside the families/ethereum/modules folder, the same way we do for erc20 and compound with any peculiarities we might have for them handled there.

IAmMorrow:
as far as I remember @gre was opposed to the idea of reusing "TokenAccount" which are used everywhere in the app (and it would break everywhere).

juan-cortes:
I actually initially wrote having another type of account, NonFungibleTokenAccount or whatever, that can residen alongside the other types of subaccounts, but I guess I changed my mind midway.

If TokenAccount breaks things, then for sure, another type. Just not fields on Account :shrug:

src/families/ethereum/synchronisation.ts
gre: no, subOperations are strictly for "operations of sub accounts". I believe here is the correct choice.

src/types/account.ts
gre: `nfts`

gre:
> We would need to build a list of TokenCurrency on sync dynamically, as opposed to based on a hardcoded list like we do for ERC-20. This dynamic list would come from the operations for the user and would be used to know what they can send, I don't think we need to support a list of what we can receive since that's an immense amount of different tokens.

I would prefer not to touch `Currency` type which is not designed to be dynamic today. indeed it's going to happen in future (dynamic tokens) but i'm not sure we should put this idea in the NFT implementation yet. Lot of parts works in this "load at boot" mode and are not ready for this dynamically.

Also, I feel "Currency" type don't match a lot the idea of an NFT. it is not a currency (look at all the fields of currency, not a lot of them makes sense, like what's the ticker of a NFT? are you going to try to get a countervalues from countervalues api? because with our current logic it's going to try this if you make it such an account =D anyway, just to say that I believe it will be more work to hack that way rather than just making a new type for "NFTs" ).

gre:
For me, the general architecture is good to go

gre (approved):

PR #1410, opened 2021-09-22
LL-7348 - SWAP - Test useProviderRates hook 

jelbaz-ledger:
## Context (issues, jira)

[LL-7348]

## Expectations

[LL-7348]: https://ledgerhq.atlassian.net/browse/LL-7348

### DIFF
--- a/src/exchange/swap/getExchangeRates.ts
+++ b/src/exchange/swap/getExchangeRates.ts
@@ -1,7 +1,7 @@
 import type { Exchange, GetExchangeRates } from "./types";
 import type { Transaction } from "../../types";
 import { getAccountCurrency, getAccountUnit } from "../../account";
-import type { Unit } from "../../types";
+import type { Unit, TokenCurrency, CryptoCurrency } from "../../types";
 import { formatCurrencyUnit } from "../../currencies";
 import { mockGetExchangeRates } from "./mock";
 import network from "../../network";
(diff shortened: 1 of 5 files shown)


src/exchange/swap/getExchangeRates.ts
```diff
@@ -16,16 +16,19 @@ import {
 const getExchangeRates: GetExchangeRates = async (
   exchange: Exchange,
   transaction: Transaction,
-  userId?: string // TODO remove when wyre doesn't require this for rates
+  userId?: string, // TODO remove when wyre doesn't require this for rates
```
juan-cortes, line 19: this is no longer required for Wyre to give us the rates, you can verify (connected to a US vpn)
```
curl -L -X POST 'https://swap.ledger.com/v3/rate' \
-H 'Content-Type: application/json' \
--data-raw '{"from": "bitcoin", "to": "ethereum", "amountFrom": "1"}'
```

src/exchange/swap/getExchangeRates.ts
```diff
@@ -16,16 +16,19 @@ import {
...
+  currencyTo?: TokenCurrency | CryptoCurrency | undefined | null
```
juan-cortes, line 20: yikes

jelbaz-ledger: Yeah this is not ideal but I tried to keep it compatible with the old swap :/. 
I guess the alternative would have been to copy/paste the function and change the args which is not super nice either.

src/exchange/swap/getExchangeRates.ts
jelbaz-ledger: I'll double check 👍

src/exchange/swap/getExchangeRates.ts
juan-cortes: nah, this is good, it's just ugly 🤣

juan-cortes (approved):

imqdee (approved):

PR #1417, opened 2021-09-23
LL-7380 (Elrond): update api endpoint for ledger proxy

LFBarreto:
## Context (issues, jira)
[LL-7380]

## Description / Usage

update api endpoint for elrond

## Expectations

[LL-7380]: https://ledgerhq.atlassian.net/browse/LL-7380

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -82,7 +82,7 @@ const envDefinitions = {
   },
   ELROND_API_ENDPOINT: {
     parser: stringParser,
-    def: "https://api.elrond.com",
+    def: "https://elrond.coin.ledger.com",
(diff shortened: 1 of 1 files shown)


src/env.ts
```diff
@@ -82,7 +82,7 @@ const envDefinitions = {
   },
   ELROND_API_ENDPOINT: {
     parser: stringParser,
-    def: "https://api.elrond.com",
+    def: "https://elrond.coin.ledger.com/ ",
```
gre, line 85: extra `/` should be dropped?

gre (changes requested):

gre (approved):

PR #1420, opened 2021-09-24
LL-7324 Build account ID for non-account-based coins

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-7324

## Description / Usage

Until now, all JS implementations were account-based coins, which ID was built on the unique address of the account.
With bitcoin-like coins, we can't rely on an address we need to use the xpub instead as unique identifier (as done by libcore).

This PR moves responsibility of generating the account ID, from the generic code in jsHelpers, to each coins' implementation of `getAccountShape`.

## Expectations

Existing tests on bitcoin-like coins ensure the ID is the same as the libcore impl.

### DIFF
--- a/src/bridge/jsHelpers.ts
+++ b/src/bridge/jsHelpers.ts
@@ -24,6 +24,7 @@ import {
   emptyHistoryCache,
   generateHistoryFromOperations,
   recalculateAccountBalanceHistories,
+  encodeAccountId,
 } from "../account";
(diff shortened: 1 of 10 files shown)


gre (commented):
lgtm.
I think we also need to fix cross.ts file
see also some lint failure

gre (approved):

PR #1421, opened 2021-09-27
[DRAFT] LL-6675 - See NFT in the CLI

ghost:
## Context (issues, jira)
Jira Link: [LL-6675](https://ledgerhq.atlassian.net/browse/LL-6675)

## Description / Usage

screenshot_2021-09-08_at_17 43 44

## Expectations

### DIFF
--- a/cli/src/commands/sync.ts
+++ b/cli/src/commands/sync.ts
@@ -1,5 +1,6 @@
-import { map } from "rxjs/operators";
+import { map, switchMap } from "rxjs/operators";
 import { accountFormatters } from "@ledgerhq/live-common/lib/account";
+import { metadataCallBatcher } from "@ledgerhq/live-common/lib/nft";
 import { scan, scanCommonOpts } from "../scan";
 import type { ScanCommonOpts } from "../scan";
 export default {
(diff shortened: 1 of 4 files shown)


cli/src/commands/sync.ts
```diff
@@ -20,6 +21,23 @@ export default {
...
+              ...account,
+              nfts: await Promise.all(
+                account.nfts.map(async (nft) => {
+                  const { result: metadata } = await metadataCallBatcher.load({
+                    contract: nft.collection.contract,
+                    tokenId: nft.tokenId,
+                  });
+
+                  return { ...nft, metadata };
+                })
```
gre, line 36: mmh i have a feeling that logic should be generalized in a function back into live-common to "resolve all lazy fields of an account" because that logic feels specific

gre (changes requested):
lgtm but one concern that deserve more general thoughts.
curious how we deal with this on LLM and LLD too. we resolve lazily at demand right?

I feel that at least we need a function `resolveNFT` to be generalized back inside live-common, so we just need to `Promise.all(account.nfts.map(resolveNFT))` if that makes sense

gre:
we'll address the notes later

PR #1423, opened 2021-09-27
LL-7453 accept 4xx to be retried like this was the case in libcore impl

gre:


src/families/bitcoin/wallet-btc/explorer/index.ts
```diff
@@ -54,7 +54,15 @@ class BitcoinLikeExplorer extends EventEmitter implements IExplorer {
...
     // 3 retries per request
-    axiosRetry(client, { retries: 3 });
+    axiosRetry(client, {
+      retries: 3,
+      retryCondition: (e) =>
+        isNetworkOrIdempotentRequestError(e) ||
+        // workaround for explorers v3 that sometimes returns 4xx instead of 5xx
+        (e.code !== "ECONNABORTED" &&
+          (!e.response ||
+            (e.response.status >= 400 && e.response.status <= 499))),
```
haammar-ledger, line 64: Are 5xx errors still covered?

gre: yes. it's part of isNetworkOrIdempotentRequestError

PR #1424, opened 2021-09-27
Adding logo for index-cooperative products

ripples3:
## Expectations


src/data/icons/svg/index-coop/DPI.svg
```diff
@@ -0,0 +1,27 @@
+<!-- Generator: Adobe Illustrator 25.2.1, SVG Export Plug-In . SVG Version: 6.00 Build 0)  -->
+<svg version="1.1" id="Ebene_1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink" x="0px" y="0px"
+	 viewBox="0 0 283.5 283.5" style="enable-background:new 0 0 283.5 283.5;" xml:space="preserve">
+<style type="text/css">
```
gre, line 4: unfortunately we can't have style, id, or "html extended"  SVG properties because have to export it to React Native,  refer to existing icon, it needs to be clean/"normalized" the same way.

example https://github.com/LedgerHQ/ledger-live-common/blob/6bdb4a3978c9316779d2093cec1f9fd8b6f6435f/src/data/icons/svg/ABT.svg?short_path=d2c4b79

it also need to be at the top level by ticker.

gre (changes requested):

src/data/icons/svg/index-coop/DPI.svg
ripples3: Hey gre, thank you for checking.

One question, what do you mean by "the top level by ticker"? Should I remove /index-coop/? path?

gre: yes, that way our importer will see it https://github.com/LedgerHQ/ledger-live-common/blob/12f8258ea4811b5023fc16899221f45612180694/scripts/buildReactIcons.js#L70

ripples3: @gre, I uploaded a new set of .svg files. 
Let me know if this is right. Thank you.

ripples3: @gre, just wanted to follow up on this. Please let me know if this is good.

gre:
Thanks for the contribution.
could you merge master in? we have extra test that do some checks to ensure icons are fine. Thanks

very sorry for the latency on this.

PR #1433, opened 2021-10-04
LL-7233 - SWAP - Fix the sort algorithm from usePickDefaultAccount hook

imqdee:
## Context (issues, jira)

[LL-7233]

Use bignumber methods to compare two bignumbers.

## Expectations

[LL-7233]: https://ledgerhq.atlassian.net/browse/LL-7233


src/exchange/swap/hooks/usePickDefaultAccount.ts
```diff
@@ -1,5 +1,6 @@
 import { useEffect } from "react";
 import { Account, TokenAccount } from "../../../types";
+import { BigNumber } from "bignumber.js";
 
 // Pick a default source account if none are selected.
 export const usePickDefaultAccount = (
```
LFBarreto, line 6: the reeduce logic seems a bit overcomplicated to me if you're just trying to filter and sort the accounts :shrug: 

```js
const highestbalanceAccount = accounts.filter(acc => acc.balance.gt(0))
                               .sort((a,b) => a.balance.gt(b.balance)))
                               .find(Boolean);
```

should do the trick in that case

LFBarreto (changes requested):

LFBarreto (approved):
LGTM :sweat_smile:

valpinkman (approved):

gre (approved):

PR #1439, opened 2021-10-05
 LL-7593 Fix inconsistent result of mergeOps

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-7593

## Description / Usage

Small change of comparison rule in `mergeOps` to ensure the output is consistent between all `return` cases.

## Expectations


haammar-ledger:
We should wait for CI before merging, to ensure no regression on other coins.

PR #1451, opened 2021-10-08
 LL-7655 taproot derivation of accounts

gre:
## Context (issues, jira)

LL-7655

## Description / Usage

work in progress

## Expectations


src/families/bitcoin/js-synchronisation.ts
```diff
@@ -29,6 +29,7 @@ import { getCryptoCurrencyById } from "@ledgerhq/cryptoassets";
 const toWalletDerivationMode = (
   mode: DerivationMode
 ): WalletDerivationModes => {
+  // FIXME really bad to maintain this. we should either add it into derivation.ts or use better way like "isSegwit"
```
haammar-ledger, line 32: Tech debt due to wallet-btc using different types than LLC. I will change this.

PR #1458, opened 2021-10-13
feat: Replaced old DATA token logo with new

jtakalai:
## Context (issues, jira)

Streamr DATA went through a token migration.

See official token migration page: https://streamr.network/token-migration/

## Description / Usage

Only logo files were updated here.

XDATA is the name of the "old" DATA token that is being migrated from. 

## Expectations

- [X] **No impact: The changes of this PR have ZERO impact on the userland.** Meaning, we can use these changes without modifying LLD/LLM at all. It will be a "noop" and the maintainers will be able to bump it without changing anything.

Test coverage not checked because: Not sure how you test for correct icons :D
No impact checked because: hopefully new icons will Just Work...


src/data/icons/svg/XDATA.svg
```diff
@@ -0,0 +1,10 @@
+<?xml version="1.0" encoding="UTF-8"?>
+<svg width="64px" height="64px" viewBox="0 0 64 64" version="1.1" xmlns="http://www.w3.org/2000/svg" xmlns:xlink="http://www.w3.org/1999/xlink">
+    <!-- Generator: Sketch 49.3 (51167) - http://www.bohemiancoding.com/sketch -->
+    <title>64px</title>
+    <desc>Created with Sketch.</desc>
+    <defs></defs>
+    <g id="64px" stroke="none" stroke-width="1" fill="none" fill-rule="evenodd">
```
gre, line 7: Thanks for the contribution. could you merge master in? we have extra test that do some checks to ensure icons are fine.

Unfortunately, we can't afford to have `id=`, `class=` or `<style>` in these SVG because we will convert them back into react class.

gre (changes requested):

PR #1486, opened 2021-10-27
[LL-8099][POC] sell flow in SDK context

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-8099
https://ledgerhq.atlassian.net/browse/LL-8107
https://ledgerhq.atlassian.net/browse/LL-8863

🔗  https://github.com/LedgerHQ/ledger-live-desktop/pull/4361

## Description / Usage

- Add of the exchange Sell and Fund flow in the context of the [live-app-sdk](https://github.com/LedgerHQ/live-app-sdk).
- Enable exchange app end to end testing through env variable and test config and provider signed with Ledger test private key linked to **[this](https://github.com/LedgerHQ/app-exchange/blob/master/src/init.c#L6-L17)** Ledger `TEST_PUBLIC_KEY `

## Expectations


IAmMorrow (approved):

src/exchange/platform/startExchange.ts
```diff
@@ -23,8 +23,8 @@ const startExchange = (
       await withDevicePromise(deviceId, async (transport) => {
         log("exchange", `attempt to connect to ${deviceId}`);
         if (unsubscribed) return;
-        const swap = new Exchange(transport, exchangeType, 0x00); // TODO Pass fixed/float for UI switch ?
-        const nonce: string = await swap.startNewTransaction();
+        const exchange = new Exchange(transport, exchangeType); // TODO Pass fixed/float for UI switch ?
```
juan-cortes, line 26: You could probably solve that todo if you wanted now, isn't it a matter of adding an optional rateType to the `StartExchagneInput` and passing it in?

juan-cortes (approved):
lgtm, im sorry this is still in poc status

src/exchange/hw-app-exchange/Exchange.ts
```diff
@@ -2,17 +2,18 @@ import Transport from "@ledgerhq/hw-transport";
...
-export const TRANSACTION_TYPES = {
-  SWAP: 0x00,
-  SELL: 0x01,
-  FUND: 0x02,
-};
-type TransactionRate = typeof TRANSACTION_RATES[keyof typeof TRANSACTION_RATES];
-type TransactionType = typeof TRANSACTION_TYPES[keyof typeof TRANSACTION_TYPES];
+
+export const enum RateTypes {
+  FIXED = 0x00,
```
JunichiSugiura, line 7: enum property is usually defined by UpperCamerlCase

JunichiSugiura: Is const key necessary for enum in flow? It's not a case in TS.

chabroA: > enum property is usually defined in UpperCamelCase

Indeed, I'll update that

> Is const key necessary for enum in flow? It's not a case in TS.

Don't really understand what you mean by "necessary for enum in flow"?
We are not using flow anymore in LLC but TS.
Regarding the `const` keyword, it is not "necessary" per say but makes sense in this case since the enum is made up of const (i.e: non computed) members. cf. https://www.typescriptlang.org/docs/handbook/enums.html#const-enums

gre (approved):
lgtm. impact looks safe.

PR #1497, opened 2021-11-03
[LL-7761]Bitcoin js perf fix

hzheng-ledger:
improve the performance of btc-wallet for sync an account
https://ledgerhq.atlassian.net/browse/LL-7761

### DIFF
--- a/src/families/bitcoin/wallet-btc/xpub.ts
+++ b/src/families/bitcoin/wallet-btc/xpub.ts
@@ -7,6 +7,7 @@ import { ICrypto } from "./crypto/types";
 import { PickingStrategy } from "./pickingstrategies/types";
 import * as utils from "./utils";
 import { TransactionInfo, InputInfo, OutputInfo } from "./types";
+import Base from "./crypto/base";
 
 // names inside this class and discovery logic respect BIP32 standard
 class Xpub extends EventEmitter {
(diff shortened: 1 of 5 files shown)


src/families/bitcoin/wallet-btc/xpub.ts
```diff
@@ -440,18 +415,25 @@ class Xpub extends EventEmitter {
...
-      { address, account, index },
-      lastTx
-    );
-    const inserted = await this.storage.appendTxs(txs);
+      txs = await this.explorer.getAddressTxsSinceLastTxBlock(
+        this.txsSyncArraySize,
+        { address, account, index },
+        lastTx
+      );
+      pendingTxs = pendingTxs.concat(txs.filter((tx) => !tx.block));
```
machard, line 433: no concat because the ledger explorer always return all the pending tx on every page.
and a regular explorer would probably just return the pending tx after it returned the confirmed txs

machard: however they would not have been added twice, storage is careful about that

machard:
for posterity i have to say that the receive address generation has been available easily from the wallet.ts wrapper from the beginning...
https://github.com/LedgerHQ/wallet-btc/blob/master/src/wallet.ts#L164

PR #1512, opened 2021-11-08
[LL-7913] fix confirmation status for btc operation

hzheng-ledger:


hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LL-7913

gre (approved):

PR #1513, opened 2021-11-08
LL-7972 Forbid receive to mainnet taproot address before activation

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-7972
https://ledger.slack.com/archives/C02DR5Y5EDN/p1636381753312700

## Expectations


gre:
per discussion. probably not needed. we keep the branch for a bit in case we still need it.

PR #1522, opened 2021-11-15
LL-8133 Improve tx size estimation function

kallerosenbaum:
* rename estimateTxSize -> maxTxSize
* Provide output addresses to maxTxSize
* Provide outputs to UTXO picking strategies

## Expectations

### DIFF
--- a/src/families/bitcoin/wallet-btc/utils.ts
+++ b/src/families/bitcoin/wallet-btc/utils.ts
@@ -6,6 +6,7 @@ import { DerivationModes } from "./types";
 import { Currency, ICrypto } from "./crypto/types";
 import cryptoFactory from "./crypto/factory";
 import { fallbackValidateAddress } from "./crypto/base";
+import { UnsupportedDerivation } from "../../../errors";
 
 export function parseHexString(str: any) {
   const result: Array<number> = [];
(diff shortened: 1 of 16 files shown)


gre (approved):
I can't really comment on the logic but great to see this is covered by test 🚀 . If you want it to run against the bot, we can put this in a bitcoin-js branch.

kallerosenbaum:
Why is CI failing @gre ? What's with the src/__tests__/__snapshots__/all.libcore.ts.snap file? It's probably because I accidentally pushed a change on all.libcore.ts, and then reverted the change and force-pushed again. But I'm not sure exactly why it happens. What can I do to fix this?

src/families/bitcoin/wallet-btc/utils.ts
```diff
@@ -91,58 +92,129 @@ export function byteSize(count: number) {
...
+ * maximum.
+ *
+ * @param inputCount Number of inputs of the imaginary transaction
+ * @param outputAddrs The addresses of the outputs, excluding change.
+ * @param includeChange Indicates whether we should add a change output. The
+ * change output will have the same derivation mode as the inputs. See
+ * derivationMode parameter.
+ * @param currency The currency for which the calculation is done.
+ * @param derivationMode The derivation mode used for calculating the size of
+ * the inputs.
```
haammar-ledger, line 176: Appreciated :clap: :+1: thanks.

ghost (approved):

PR #1528, opened 2021-11-16
LL-7930 Ensure bech32 version 1.1.x and 2.0.0 are not used together

ghost:
## Context (issues, jira)

**Error while adding taproot account in LLM**
https://ledgerhq.atlassian.net/browse/LL-7930

## Description / Usage

Issues happen on LLM everywhere bech32 is used (scan of accounts but also recipient address validation).
Root cause is that different versions of bech32 currently coexist on LLM, as shown by:
```
$ yarn list bech32
├─ @ledgerhq/live-common@21.16.0
│  └─ bech32@2.0.0
├─ bech32@1.1.4
└─ bitcore-lib@8.25.10
   └─ bech32@1.1.3
```

Workaround: fallback using only bech32 `^1.1.3` instead of `2.0.0`, and copy some new bech32m code from 2.0.0 (used only for tests).
Fixes all bech32 issues on LLM.

## Expectations


haammar-ledger:
Regression on LLD at the moment.

src/families/bitcoin/wallet-btc/crypto/bitcoin.ts
```diff
@@ -140,7 +141,12 @@ class Bitcoin extends Base {
   }
 
   private tryBech32(address: string): boolean {
-    const result = fromBech32(address);
+    let result;
+    try {
+      result = fromBech32(address);
+    } catch {
+      return false;
```
kallerosenbaum, line 148: I think it's wrong to return false here. If you look at Bitcoin.validateAddress, it expects an exception if bech32 decoding fails, so that it can try base58 instead:

```
    try {
      // This prefix check is to avoid returning false in cases where a valid base58 address also happens
      // to be a valid bech32(m) string (but invalid segwit address).
      if (address.toLowerCase().startsWith(`${this.network.bech32}1`)) {
        return this.tryBech32(address);
      }
    } catch {
      /* Try base58 instead */
    }
```
I'm surprised that this wasn't caught by tests. I'm almost certain it would be caught by "toOutputScript p2sh and p2pkh work as expected" in bitcoin.test.ts.

kallerosenbaum (changes requested):

src/families/bitcoin/wallet-btc/crypto/bitcoin.ts
kallerosenbaum: Ok, it's only when the address starts with bc1... that tryBech32 is called, so it's only in extreme (weird) edge cases this is a problem. But I still think throwing is the correct thing to do as my comment suggests.

gre:
@haammar-ledger can you explain why 2.0.0 is broken in context of LLM? I'm not sure to get it. if there is a bug in bech32 we should report to them.

haammar-ledger:
@gre It's not 2.0.0 that's an issue, it's having both 1.1.x and 2.0.0 at the same time. But I don't know why they can't coexist, unlike usual JS dependencies that can have multiple versions included...

We can't use 2.0.0 because all these dependencies currently pull 1.1.x:
- bitcoinjs-lib
- @cosmjs/encoding
- @crypto-com/chain-jslib
- @ethersproject/providers

So the only other option seems to keep using 1.1.x for now.

gre:
you can ignore the compound related test. i'll have to drop these unstable tests really soon. meanwhile the CI pass on the rest.
@haammar-ledger do you think this is mergeable?

haammar-ledger:
Issue on LLD has disappeared, this is mergeable.

kallerosenbaum (approved):

PR #1529, opened 2021-11-18
LL-8185 add hardcoded dependencies between plugins and eth app

ghost:


gre:
in the future, please ask a dev, could be more productive 👍

PR #1536, opened 2021-11-24
Add Celo coin with Sync and Send features

pawelnguyen:
## Context (issues, jira)

This adds Celo coin with Sync, Send and Receive features with the intention to add it to Ledger Live Desktop and add Staking features later.

I haven't added a Celo icon, docs state that it should be handled by the design team

## Description / Usage

To fetch account information and transactions/operations, we're using an indexer. For the rest, we're using ContractKit which is a Celo SDK.

You'll need these env variables set up `API_CELO_NODE` `API_CELO_INDEXER`, ping me on slack and I'll help set it up

Example usage via CLI:

```
ledger-live sync -c celo
```
CleanShot 2021-11-24 at 14 41 57@2x

```
EXPERIMENTAL_CURRENCIES=celo ledger-live getTransactionStatus -c celo -i 0 --amount 0  --recipient 0x2a38Fdf0A79Ad366628C83bfdA4036dD64358BA0 --use-all-amount
```

CleanShot 2021-11-24 at 14 44 31@2x

```
EXPERIMENTAL_CURRENCIES=celo ledger-live send -c celo -i 0 --amount 0.001  --recipient 0x2a38Fdf0A79Ad366628C83bfdA4036dD64358BA0
```
CleanShot 2021-11-24 at 14 49 11@2x

https://hubble.figment.io/celo/chains/celo/transactions/0x8c1ee3cbb4e432b5f9fa673ac24eff599260fbbee78d9d51bbb7f640f964d657

## Screenshots from LLD

Zrzut ekranu 2021-11-24 o 13 25 03
Zrzut ekranu 2021-11-24 o 13 25 13
Zrzut ekranu 2021-11-24 o 13 25 33
Zrzut ekranu 2021-11-24 o 13 25 53
Zrzut ekranu 2021-11-24 o 13 26 48
Zrzut ekranu 2021-11-24 o 13 26 58
Zrzut ekranu 2021-11-24 o 13 27 08
Zrzut ekranu 2021-11-24 o 13 27 28
Zrzut ekranu 2021-11-24 o 13 27 38
Zrzut ekranu 2021-11-24 o 13 27 43
Zrzut ekranu 2021-11-24 o 13 27 47
Zrzut ekranu 2021-11-24 o 13 29 55
Zrzut ekranu 2021-11-24 o 13 30 02

## Expectations

Related PR https://github.com/LedgerHQ/ledgerjs/pull/723


cli/src/transaction.ts
```diff
@@ -106,6 +106,8 @@ export async function inferTransactions(
       transaction.amount = transaction.useAllAmount
         ? new BigNumber(0)
         : inferAmount(account, opts.amount || "0");
+      // Most probably there's a bug here and this method should `return transaction`,
+      // but every family overwrites this method in cli-transaction.ts so can't really reproduce
```
gre, line 110: I'm not sure to follow. No bug here to me.

https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/ethereum/cli-transaction.ts#L101-L109

these have correct types on the other end as it's getting exploded back, but indeed the cli-transaction side will remap them into transaction.

```
function inferTransactions(
  transactions: Array<{
    account: AccountLike;
    transaction: Transaction;
    mainAccount: Account;
  }>,
  opts: Record<string, any>,
  { inferAmount }: any
): Transaction[]
```

src/families/celo/hw-app-celo.ts
```diff
@@ -0,0 +1,80 @@
+import Eth from "@ledgerhq/hw-app-eth";
+import { splitPath, foreach } from "@ledgerhq/hw-app-eth/lib/utils";
+import { decode, encode } from "rlp";
+
+/**
+ * Copied from celo-web-wallet, which in turn reused hw-app-eth/src/Eth.js
+ * https://github.com/celo-tools/celo-web-wallet/blob/master/src/features/ledger/CeloLedgerApp.ts
+ */
+export class CeloApp extends Eth {
```
gre, line 9: I'm not sure this is really a good idea to extend `@ledgerhq/hw-app-eth`

any idea what is the reason to extend it in this context? also why ethereum (it's a fork?) and any reason in that case why the "hw-app-eth" itself wouldn't fit?   (thinking out loud here)

yarn.lock
```diff
@@ -7175,6 +9164,246 @@ walker@^1.0.7:
...
+    bn.js "^4.11.9"
+    eth-lib "0.2.8"
+    ethereum-bloom-filters "^1.0.6"
+    ethjs-unit "0.1.6"
+    number-to-bn "1.7.0"
+    randombytes "^2.1.0"
+    underscore "1.12.1"
+    utf8 "3.0.0"
+
+web3@1.3.6:
```
gre, line 9394: this would be the first time we depends on `web3`, we maybe have to study if it's a heavy move and if it's supported everywhere (typically in React Native environment).

gre (commented):
looks a good integration so far.
the thing I can see a bit tricky is the fact the celo package and ecosystem brings a lot of dependency into our project, and notable "web3" which wasn't depended on so far, so there will be a study of impact & making sure it works everywhere, but not necessarily a blocker.

cli/src/transaction.ts
pawelnguyen: Ok, removing the comment if it should work like this, thanks :)

src/families/celo/hw-app-celo.ts
pawelnguyen: I agree, there's a small risk things might stop working when hw-app-eth changes.

The reason for extending it is that yes, Celo is an Ethereum fork but it does have a different form of transactions. I went with your line of thought at first and did try to use `hw-app-eth`, but the differences in transactions mean that most of the code is same/similar, but we have to encode a couple more fields:

https://docs.celo.org/developer-guide/celo-for-eth-devs#key-differences-between-celo-and-ethereum
`Celo transaction objects are slightly different from transaction objects on Ethereum. Ethereum transaction objects include fields to, value, gas, gasPrice, data, nonce, signature (v,r,s). Celo transaction objects include the same fields as Ethereum transaction objects, plus feeCurrency, gatewayFee and gatewayFeeRecipient`

If we don't want to extend `@ledgerhq/hw-app-eth`, I can copy paste it and overwrite relevant pieces or stick to a certain version of hw-app-eth. Any other ideas?

yarn.lock
pawelnguyen: good catch, let me know if I should do anything about it 👍

gre (changes requested):
Please add a `spec.js` that would allow us to automate tests for transaction and synchronisations.

you can find a barebone example of this here:

https://github.com/LedgerHQ/ledger-live-common/pull/1338

and documentation:

https://github.com/LedgerHQ/ledger-live-common/blob/master/docs/bot.md

src/families/celo/hw-app-celo.ts
gre: after double thinking on this topic. I would prefer we don't use inheritence, especially that we're planning some evolution on the ETH signTransaction signature. You can just take the part you need and fork the code.

pawelnguyen: Sure, will do 👍 
I'll add a spec as well

pawelnguyen:
hey @gre, I need some help :)
I started adding a `specs.ts` by setting up that bot you linked, but I'm stuck on an error.

1. I followed Getting Started steps https://github.com/LedgerHQ/ledger-live-common/blob/master/docs/bot.md#getting-started
2. I ran `ledger-live bot` and got stuck on this: https://cln.sh/wse1eQ
3. I added `celo` to `/src/__tests__/test-helpers/setup.ts` but I don't see it trying to run specs for Celo (I'll debug it later but maybe I'm missing something obvious)

I understand that I should set `COINAPPS` variable? I'm not sure where can I get an `.elf` file from for Celo

src/families/celo/hw-app-celo.ts
pawelnguyen: ok, that's done @gre

pawelnguyen:
@gre this is ready for another round of code review

Specs that I ran locally with a bot:
photo_2022-01-11_15-25-25

src/families/celo/js-getTransactionStatus.ts
```diff
@@ -0,0 +1,65 @@
...
+  }
+
+  if (!errors.amount && account.spendableBalance.lt(estimatedFees)) {
+    errors.amount = new NotEnoughBalance();
+  }
+
+  if (!transaction.recipient) {
+    errors.recipient = new RecipientRequired();
+  } else if (!isValidAddress(transaction.recipient)) {
+    errors.recipient = new InvalidAddress();
```
haammar-ledger, line 53: Please use this:
```
errors.recipient = new InvalidAddress("", { currencyName: a.currency.name, });
```
It will display the coin's name in the error message. (we should document or improve this...)

haammar-ledger:
LGTM! Just need to rebase on `master` / solve conflicts.

src/families/celo/js-signOperation.ts
```diff
@@ -0,0 +1,146 @@
...
+
+          const celo = new CeloApp(transport);
+          const unsignedTransaction = await buildTransaction(
+            account,
+            transaction
+          );
+          const { chainId, to } = unsignedTransaction;
+          const rlpEncodedTransaction = rlpEncodedTx(unsignedTransaction);
+
+          const tokenInfo = tokenInfoByAddressAndChainId(to!, chainId!)
```
gre, line 99: some lint issues raised by ci.

gre (changes requested):
could you merge latest master in? Hopefully it should pass the CI. thanks

src/families/celo/js-signOperation.ts
pawelnguyen: oops, thanks for noticing that 👍 fixed

pawelnguyen:
@gre seems like it passed tests in most cases, on Linux it fails with `Error: Error: The 'github-token' input is not specified`

ghost (approved):
:clap:

pawelnguyen:
yay! thanks 🎉

PR #1545, opened 2021-11-30
[LL-7227][LL-7539]Bitcoin js

hzheng-ledger:
Fix peercoin and BCH for js implementation


src/api/Ledger.ts
```diff
@@ -25,7 +26,15 @@ export const findCurrencyExplorer = (
...
   if (config.experimental && getEnv("EXPERIMENTAL_EXPLORERS")) {
-    const { base, version } = config.experimental;
+    const base = config.experimental.base;
+    let version = config.experimental.version;
+    //V2 explorer for doge and bitcoin cash when libcore is used
+    if (
+      (currency.id === "bitcoin_cash" || currency.id === "doge") &&
+      !shouldUseJS(currency)
+    ) {
+      version = "v2";
```
gre, line 36: shouldn't bitcoin_cash + V2 also to set the explorer id to abc ?

hzheng-ledger: 1. Confirmed by backend team, "abc" and "bch" are exactly same for v2/v3. it doesn’t matter whether we use abc/bch.
2. In fact, for the old implementation, the id doesn’t impact the url. (but the version does)
for the old implementation, it seems that the “id” used by explorer url (abc) depends on the libcore
https://github.com/LedgerHQ/lib-ledger-core/blob/fc9d762b83fc2b269d072b662065747a64ab2816/core/src/wallet/bitcoin/networks.cpp#L93
So, it doesn't break anything I think

gre (approved):
lgtm

src/families/bitcoin/js-signOperation.ts
```diff
@@ -73,14 +73,10 @@ const signOperation = ({
 
         const segwit = isSegwitDerivationMode(account.derivationMode);
 
-        // FIXME Call to explorer needed to set timestamp (https://ledgerhq.atlassian.net/browse/LL-7539)
-        // cf. https://github.com/LedgerHQ/lib-ledger-core/blob/fc9d762b83fc2b269d072b662065747a64ab2816/core/src/wallet/bitcoin/transaction_builders/BitcoinLikeUtxoPicker.cpp#L150-L154
-        /*
-        const hasTimestamp = networkParams.usesTimestampedTransaction;
+        const hasTimestamp = currency.id === "peercoin";
```
haammar-ledger, line 76: Why removing this? Could've stayed generic I think, but no big deal.

ghost (approved):

src/families/bitcoin/js-signOperation.ts
gre: I kinda agree with @haammar-ledger here

gre: I agree with @haammar-ledger , ideally (and it's not yet the case), each coin should have its folder that defines all its "specifics" and only this folder should have such "if",
that said it's not the only place where we have these I feel.

haammar-ledger: The btc-js impl already has such mechanisms actually:
- https://github.com/LedgerHQ/ledger-live-common/blob/16bc0befa95687030b0ced54c3db30369836390a/src/families/bitcoin/logic.ts#L149-L194
- https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/bitcoin/networks.ts

They're legacy, could need cleanup for some altcoins.

gre (approved):

PR #1550, opened 2021-12-02
Fixes LL-8516

gre:
- LL-8516 bsc and polygon won't display nfts yet
  - tests added to prove this
- disabling NFT back to false will make .nfts go back to undefined with a full resync
  - tests added to prove this
- enabling it back will also make a full sync & reappear (like before)
  - tests added to prove this
- a clear cache will remove all .nfts for making sure a full resync happens
- a NFT_CURRENCIES env is made available and set at "ethereum" for now


src/api/Ethereum.ts
```diff
@@ -132,10 +133,9 @@ export const apiForCurrency = (currency: CryptoCurrency): API => {
       let { data } = await network({
         method: "GET",
         url: URL.format({
-          pathname:
-            getEnv("NFT") && currency.ticker === "ETH"
-              ? `https://explorers.api-01.live.ledger-stg.com/blockchain/v3/eth/addresses/${address}/transactions`
-              : `${baseURL}/addresses/${address}/transactions`,
+          pathname: isNFTActive(currency)
+            ? `https://explorers.api-01.live.ledger-stg.com/blockchain/v3/eth/addresses/${address}/transactions`
```
haammar-ledger, line 137: Why not an env for this URL?

ghost (approved):

src/api/Ethereum.ts
gre: big temporary hack indeed. the env is "EXPLORER" in practice, the ledger-stg.com api had bugs in some other endpoints which forced us to branch only that endpoint

gre: we have a tech debt task logged

PR #1552, opened 2021-12-02
LL-8508 - Fix NFT merging in incremental sync + tests

ghost:
## Context (issues, jira)

[LL-8508](https://ledgerhq.atlassian.net/browse/LL-8508)

## Description / Usage

Fix an issue that would never remove NFT from your account once transferred. 

## Expectations


src/families/ethereum/nft.test.ts
```diff
@@ -122,6 +197,17 @@ describe("gaspard NFT on ethereum", () => {
     expect(resync.nfts).toEqual(account.nfts);
   });
 
+  test("remove half NFTs will restore them with half operations", async () => {
+    const halfOps = Math.ceil(account.operations.length || 0);
```
gre, line 201:
```suggestion
    const halfOps = Math.ceil(account.operations.length / 2);
```

gre (approved):

PR #1555, opened 2021-12-03
Filter out nft that would have amount<=0

gre:
before this fix, I was seeing NFT i don't have anymore with quantity 0 and even -1 (?!)

### DIFF
--- a/src/nft/helpers.ts
+++ b/src/nft/helpers.ts
@@ -45,7 +45,7 @@ export const nftsFromOperations = (ops: Operation[]): NFT[] => {
       return acc;
     }, {});
 
-  return Object.values(nftsMap);
+  return Object.values(nftsMap).filter((n) => n.amount.gt(0));
 };
 
 export const nftsByCollections = (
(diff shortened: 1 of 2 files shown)


porenes-ledger:
https://ledgerhq.atlassian.net/browse/LL-8540

src/nft/helpers.ts
```diff
@@ -45,7 +45,7 @@ export const nftsFromOperations = (ops: Operation[]): NFT[] => {
       return acc;
     }, {});
 
-  return Object.values(nftsMap);
+  return Object.values(nftsMap).filter((n) => n.amount.gt(0));
```
haammar-ledger, line 48: No risk of empty values in `nftsMap`?
If so should do `Object.values(nftsMap).filter(Boolean).filter((n) => n.amount.gt(0))` otherwise it's fine.

gre: mmh to me no risk. `nft` is never falsy. in that `acc[nftKey] = nft;`

ghost (approved):
I don't have all context but the fix looks safe!

PR #1556, opened 2021-12-03
fix jobs

valpinkman:


.github/workflows/bot2.yml
```diff
@@ -58,6 +58,8 @@ jobs:
         run: npm i -g yarn
       - name: pull docker image
         run: docker pull ghcr.io/ledgerhq/speculos
+      - name: kill apt-get
+        run: sudo killall apt-get apt || echo OK
```
MortalKastor, line 62:
```suggestion
        run: sudo killall -w apt-get apt || echo OK
```

Adding the `-w` flag would be safer here, as it makes `killall` wait for the killed processes to finish and would prevent a potential race condition were you start installing packages before the killed apt releases the lock (confusingly using `kill` or `killall` sends a signal to the process to exit _cleanly_ and doesn't _kill_ it, unless you change the signal to something like `-9` (which I wouldn't recommend in this case))

.github/workflows/bot3.yml
```diff
@@ -60,6 +60,8 @@ jobs:
         run: npm i -g yarn
       - name: pull docker image
         run: docker pull ghcr.io/ledgerhq/speculos
+      - name: kill apt-get
+        run: sudo killall apt-get apt || echo OK
```
MortalKastor, line 64:
```suggestion
        run: sudo killall -w apt-get apt || echo OK
```

MortalKastor (changes requested):

PR #1568, opened 2021-12-08
Solana Integration 2

konoart:
## Context

As part of Solana integration


ghost (approved):

src/families/solana/js-prepareTransaction.ts
```diff
@@ -64,11 +63,14 @@ async function deriveCommandDescriptor(
...
-        const encoder = new TextEncoder();
-        if (encoder.encode(model.uiState.memo).byteLength > MAX_MEMO_LENGTH) {
+        const memoBytes = Buffer.from(model.uiState.memo, "utf-8");
+        if (memoBytes.byteLength > MAX_MEMO_LENGTH) {
           errors.memo = errors.memo = new SolanaMemoIsTooLong(undefined, {
             maxLength: MAX_MEMO_LENGTH,
           });
+          // HACK: LLM expects <transaction> as error key to disable
+          // continue button
+          errors.transaction = errors.memo;
```
haammar-ledger, line 73: Not really a hack, as the memo field is part of the transaction.
I would just set `errors.transaction = new SolanaMemoIsTooLong...`, unless you have a specific use for `errors.memo`?

konoart: To my opinion UI should consider a transaction invalid if there are any keys in `errors`. I think if you need to specify a particular key (which you can not know until you try to integrate with mobile) in `errors` for the UI to understand that the transaction is invalid is less maintainable. Plus, that doesn't work quite nicely for desktop version, since desktop version shows an input field for memo (instead of a separate screen, like on mobile) and that field should show all the warnings and errors related to memo validation. If we set memo validation error as `transaction` in `errors` there are two  consequences: 
a) desktop will not be 100% sure if the error is related to memo or not 
b) we will not be able to add more fields later (if we need to) because they will have to share `transaction` key in `errors`.
I can remove the word `HACK` but I would prefer to have both `memo` and `transaction` keys set and a comment explaining why it is so.

gre (approved):

src/families/solana/js-prepareTransaction.ts
gre: > To my opinion UI should consider a transaction invalid if there are any keys in errors

curious, I thought it was the case already. we'll study this.

PR #1572, opened 2021-12-09
LL-8572 Improve UX for connectApp flows

juan-cortes:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-8572

## Description / Usage
While investigating this issue for both  LLD and LLM integrations I saw several problems that led to situations where we displayed a "Connect and unlock your device" while the device was still connected. 
- After a successful `quitApp` we emitted a disconnect event in order to restart the flow for LLM, this impacted LLD in showing the connect animation when not needed, since the event driven paradigm will detect the re-connection and start the flow again. Since we still need the disconnect event for polling I simply added a `expected` flag so that the reducer will show the `isLoading` animation on LLD instead of the connect one, since we know there will be a connection afterwards.
- There's also a case where for the wired implementation the debounces were conflicting, meaning we would react to the disconnect one for a second and then see the connect one, simply conditionally debouncing the disconnects a bit more like we had for LLM mitigated the issue further.

## Testing
I believe there might still be cases where we _could_ see the connect your device animation as a false positive but this should be better than the current production version. Simply go over flows for both LLM and LLD with different setups and see if it's better than now:
- Without the target app, to trigger inline install
- While being in the wrong app, to trigger the quitApp
- Needing "Allow manager" 
- Confirming a receive
- etc

## Expectations

## Videos
Here is a sample flow that was impacted by the issue. On LLD, while inside the BTC app, start an ETH add account flow. This should be the same in both cases but notice how one of them asks the user twice to connect the device and the other one doesn't.

https://user-images.githubusercontent.com/4631227/145415853-d09bb478-8afe-414c-818a-5d328fb788e3.mov

https://user-images.githubusercontent.com/4631227/145415176-545e53bc-c1b5-4575-a547-1592b42f8ce6.mov

### DIFF
--- a/src/hw/actions/app.ts
+++ b/src/hw/actions/app.ts
@@ -462,13 +462,13 @@ const implementations = {
 
               if (disconnectT) {
                 // any connect app event unschedule the disconnect debounced event
-                clearTimeout(disconnectT);
                 disconnectT = null;
+                clearTimeout(disconnectT);
               }
 
               if (event.type === "unresponsiveDevice") {
                 return; // ignore unresponsive case which happens for polling
-              } else if (event.type === "disconnected" && !event.expected) {
+              } else if (event.type === "disconnected") {
                 // the disconnect event is delayed to debounce the reconnection that happens when switching apps
                 disconnectT = setTimeout(() => {
                   disconnectT = null;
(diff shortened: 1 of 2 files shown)


juan-cortes:
Requesting the review from @gre because I can't review myself

juan-cortes:
HODL until I make sure some connectivity issues are not related to this work

gre (approved):

PR #1574, opened 2021-12-10
Process raw txs to combine send and fees in only one tx

emmanuelm41:
## Context (issues, jira)
The backend service generates two registers per transaction: one for the fee and one for the transaction itself. We must combine both of them on only one register.

## Description / Usage
We create a temporal index by tx hash in order to group fee and send tx together, and after we create one tx with fees included, if they are present. 

## Expectations

### DIFF
--- a/src/families/filecoin/bridge/utils/utils.ts
+++ b/src/families/filecoin/bridge/utils/utils.ts
@@ -14,27 +14,59 @@ import { encodeAccountId } from "../../../../account";
 import flatMap from "lodash/flatMap";
 import { Transaction } from "../../types";
 
+type TxsById = {
+  [id: string]: {
+    Send: TransactionResponse;
+    Fee?: TransactionResponse;
+  };
+};
+
 export const getUnit = () => getCryptoCurrencyById("filecoin").units[0];
+
+export const processTxs = (
+  txs: TransactionResponse[]
+): TransactionResponse[] => {
+  const txsById = txs.reduce((result: TxsById, currentTx) => {
+    const { hash, type } = currentTx;
+    const txById = result[hash] || {};
+
+    if (type == "Send" || type == "Fee") txById[type] = currentTx;
+
+    result[hash] = txById;
+    return result;
+  }, {});
+
+  const processedTxs: TransactionResponse[] = [];
+  for (const txId in txsById) {
+    const { Fee: feeTx, Send: sendTx } = txsById[txId];
+
+    if (feeTx) sendTx.fee = feeTx.amount.toString();
+
+    processedTxs.push(sendTx);
+  }
+
+  return processedTxs;
+};
 
 export const mapTxToOps =
   (id, { address }: GetAccountShapeArg0) =>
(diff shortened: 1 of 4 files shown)


src/families/filecoin/bridge/utils/utils.ts
```diff
@@ -14,27 +14,59 @@ import { encodeAccountId } from "../../../../account";
...
+  for (const txId in txsById) {
+    const { Fee: feeTx, Send: sendTx } = txsById[txId];
+
+    if (feeTx) sendTx.fee = feeTx.amount.toString();
+
+    processedTxs.push(sendTx);
+  }
+
+  return processedTxs;
+};
```
gre, line 49: so if I read this correctly, this will group transaction by hash. is there a possibility that more than one send is received from the same tx hash? for instance, what happens if I send to myself? (maybe this case is forbidden)

gre (commented):
👍

src/families/filecoin/bridge/utils/utils.ts
emmanuelm41: Txs will be grouped by hash, but only as a temporary step. It is not possible, as send to yourself is not allowed.

emmanuelm41: I will check with the team internally to confirm it. It is always better to make a double check.

emmanuelm41: @gre it is possible to send fil to yourself, so I removed that check. However, the backend api service won't retrieve more than one send per tx hash. In case transactions with same to and from address will be filtered.

gre: ok so if you send funds to yourself you will have a send where from and to is the same, possibly a "OUT" tx? if the value of the operation is set and if there is not the equivalent "IN" tx, we will have a problem with the balance history calculation 🤔

emmanuelm41: But it will be an OUT and an IN tx. [This](https://github.com/Zondax/ledger-live-common/blob/22739873a4758719c5db46c18ca061a12e836492/src/families/filecoin/bridge/utils/utils.ts#L59-L95) part of the code takes care of it. Is it possible I am missing something?

emmanuelm41: I would need to know if the value should include the fee or not, because now we have a difference between IN and OUT txs. I should choose one of the two ways haha.

emmanuelm41: Elrond did exactly the same we actually do. The link is [here](https://github.com/Zondax/ledger-live-common/blob/3546a7ee89c30fc3b7483d873498b98ecf4a1d3d/src/families/elrond/api/sdk.ts#L49-L56)

gre: oh ok, i get it now. thanks!

gre (approved):
all is good to me on this iteration, i'm taking it so it can make it as part of incoming live-common bump.

PR #1575, opened 2021-12-13
update libraries

gre:
see diff


gre:
ok bignumber to not upgrade somehow

PR #1585, opened 2021-12-20
Elrond ESDT tokens & EGLD staking integration

AlexandruPislariu:
## Expectations


src/families/elrond/js-buildSubAccounts.ts
```diff
@@ -0,0 +1,113 @@
...
+    swapHistory: [],
+    creationDate:
+      operations.length > 0
+        ? operations[operations.length - 1].date : new Date(),
+    balanceHistoryCache: emptyHistoryCache, // calculated in the jsHelpers
+  };
+  return tokenAccount;
+}
+
+async function elrondBuildESDTTokenAccounts({
```
gre, line 45: I am not sure if this function do this yet, but what would be good is that, if the synchronisation do not find any changes in the sub token accounts, that the object reference is kept. basically if I do two sync in a row, it should often be `oldAccount.subAccounts === newAccount.subAccounts`

we managed to solve this in Ethereum here https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/ethereum/synchronisation.ts#L112-L157 and https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/ethereum/synchronisation.ts#L631

unfortunately it's not part of a "generalized framework" but feel free to grab some code from there still.

gre (changes requested):
overall it's a cool addition, some feedback regarding the reconciliation & incremental sync. it's important we try to keep object reference to avoid the frontend to re-render all the things too much.

src/families/elrond/js-buildSubAccounts.ts
AlexandruPislariu: Currently, we do not verify if there are any changes between two sync. I will add this part.

AlexandruPislariu:
I removed the hw-app-elrond package implementation since it is already implemented in ledger-live-common.
We decided to move forward with that [version](https://github.com/LedgerHQ/ledger-live-common/tree/master/src/families/elrond/hw-app-elrond).

gre:
@AlexandruPislariu can you merge master in? It should include your work from ledgerjs

ghost (commented):
So far so good, just maybe a few parts to adapt to token transactions:
- `estimateMaxSpendable` for token transactions?
- `getTransactionStatus` : no specific errors/warnings for tokens?

gre:
please also confirm if testing against LLD have the feature overall working (you can see token accounts, you can send them over, etc..) – maybe would be great to organize a call / if you have a video of the feature working.
Thanks

AlexandruPislariu:
> please also confirm if testing against LLD have the feature overall working (you can see token accounts, you can send them over, etc..) – maybe would be great to organize a call / if you have a video of the feature working. Thanks

Currently we encounter an issue with esdt token transactions, as soon as it is fixed we can organize a call.
Also, can you please tell me how the token assets(like token log) can be displayed in LLD?

gre (approved):

haammar-ledger:
@AlexandruPislariu Sorry about that, but it seems there are unsigned commits in this branch, can you please make sure all commits are signed? :pray: Thank you.

AlexandruPislariu:
@haammar-ledger these commits are verfied with another GPG key which is no longer in my github account. I can't update the commit signature for a pushed commit.

henri-ly (approved):
LGTM ! Good job !

PR #1595, opened 2021-12-27
Fix dcr

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LL-7352

### DIFF
--- a/src/families/bitcoin/wallet-btc/crypto/decred.ts
+++ b/src/families/bitcoin/wallet-btc/crypto/decred.ts
@@ -60,12 +60,11 @@ class BIP32 {
   hmacSHA512(key, data) {
     return createHmac("sha512", key).update(data).digest();
   }
-  hash160(buffer) {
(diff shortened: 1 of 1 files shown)


src/families/bitcoin/wallet-btc/crypto/decred.ts
```diff
@@ -7,8 +7,7 @@ import createBlakeHash from "blake-hash";
 import RIPEMD160 from "ripemd160";
 import bs58 from "bs58";
 import ecc from "tiny-secp256k1";
-import createHash from "create-hash";
-import createHmac from "create-hmac";
+import crypto from "crypto";
```
haammar-ledger, line 10: Can you please check this works on LLM? Since `crypto` is a node.js package I'm thinking it may not work in react-native.

hzheng-ledger: I've checked with LLM and LLD. The PR is ready to be merged

ghost (approved):

PR #1596, opened 2021-12-29
[LL-8450] Fix custom fees bug in LLM

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LL-8450


ghost (approved):
Thanks, sorry for that oversight :)

PR #1615, opened 2022-01-06
[LL-7852]Btc perf

hzheng-ledger:
Remove unnecessary getAddress with device
https://ledgerhq.atlassian.net/browse/LL-7852


src/bridge/jsHelpers.ts
```diff
@@ -391,16 +390,17 @@ export const makeScanAccounts =
...
-
               if (!res) {
-                res = await getAddress(transport, {
-                  currency,
-                  path: freshAddressPath,
-                  derivationMode,
-                });
+                res =
+                  currency.family === "bitcoin"
+                    ? { path: freshAddressPath }
```
gre, line 396: not super confortable with this branching idea (family specific in generic codebase)

it could be a parameter of the makeScanAccounts as the fact you don't need to derivate the address (as we derivate the xpub i imagine?)

gre (changes requested):

gre:
we'll innovate in future in the way the scan accounts works. making it more parallel etc.
it will be done at the general level, not specific to BTC.
the changes here are too impacting & the specifics should be avoided to be in the global utilities (we must avoid to do "if bitcoin")

we'll reopen the topic later. thanks

PR #1618, opened 2022-01-07
LL-8914 - Remove experimental flag for NFTs

ghost:
## Context (issues, jira)

[LL-8914](https://ledgerhq.atlassian.net/browse/LL-8914)

## Description / Usage

Removes the experimental feature flag for the NFTs.
Make the FW App Ethereum 1.9.17 mandatory.

## Expectations


src/families/ethereum/synchronisation.ts
```diff
@@ -57,8 +53,7 @@ export const getAccountShape: GetAccountShape = async (
     initialAccount &&
     areAllOperationsLoaded(initialAccount) &&
     mostRecentStableOperation &&
-    !outdatedSyncHash &&
-    NFT_VERSION
```
lambertkevin, line 61: Versionning is not useful anymore because we are now regenerating all NFTs from all ops at each sync (to avoid pontetial missing txs from the explorers) instead of doing it incrementaly. 
NFT Versioning was a way to force a full resync in case we changed anything to the NFT type, or if we added new standards not indexed yet for example.

src/apps/support.ts
```diff
@@ -31,6 +31,7 @@ const appVersionsRequired = {
   Algorand: ">= 1.2.9",
   Polkadot: ">= 9.9140.0",
   Elrond: ">= 1.0.11",
+  Ethereum: ">= 1.9.17",
```
lambertkevin, line 34: Eth app 1.9.17 will now be mandatory for all eth tx because we don't have a flow that would allow us to determine 1) what device we're using 2) what the transaction is about, both at the same time and place.
Which means that we can't determine if a LNX has an outdated eth app without the NFT support OR if you're using an LNS that will just not have NFT support (at least for now).

ghost (approved):

PR #1619, opened 2022-01-07
LL-8911 Activate all BTC JS like for all users

gre:


ghost (commented):
We need BCH explorer v3 as stable, safer to do it in that PR too maybe?

ghost (approved):

PR #1621, opened 2022-01-10
peercoin fix

hzheng-ledger:
Peercoin is broken because the peercoin protocol is updated (time stamp is removed)
refer to the proposal from https://github.com/LedgerHQ/ledgerjs/pull/702


gre:
but then does this PR need to also have the work of  LedgerHQ/ledgerjs#702 ?

hzheng-ledger:
No, we can use this PR instead of https://github.com/LedgerHQ/ledgerjs/pull/702
since I think it is more reasonable to handle coin specific logic in bitcoin-js than ledgerjs

gre (approved):

PR #1624, opened 2022-01-12
init cardano integration

pavanvora:
## Context (issues, jira)

Draft PR for Cardano integration

## Description / Usage

- Support sync Accounts, receive and send ADA
- Support receive and list token balance

## Expectations


src/env.ts
```diff
@@ -216,7 +216,7 @@ const envDefinitions = {
     desc: "enable experimental support of Bluetooth",
   },
   EXPERIMENTAL_CURRENCIES: {
-    def: "",
+    def: "cardano",
```
gre, line 219: shouldn't modify this here. default must remain empty, end client configure it.

haammar-ledger, line 219: Please don't add the coin in the default value, it is meant to stay empty and only be set at runtime.

gre (changes requested):
I don't see the bridge/js.ts file in this PR, is it to come later? this is a first step but not mergeable as is because env shouldn't be modified

src/families/cardano/constants.ts
```diff
@@ -1 +1,4 @@
+import { NetworkId } from "./types";
+
 export const STAKING_ADDRESS_INDEX = 0;
+export const CARDANO_NETWORK_ID = NetworkId.testnet as number;
```
haammar-ledger, line 4: Currently targeting testnet (just noting for later reference)

src/families/cardano/env.ts
```diff
@@ -0,0 +1,44 @@
...
+    feesA: "44",
+    feesB: "155381",
+    lovelacePerUtxoWord: "34482",
+    poolDeposit: "500000000",
+    keyDeposit: "2000000",
+    maxTokenValue: "9223372036854775807",
+    networkMagic: 1097911063,
+    priceSteps: "0.0000721",
+    priceMemory: "0.0577",
+    collateralPercent: "150",
```
haammar-ledger, line 37: Same comment here, do you confirm this is not supposed to change over time?

src/families/cardano/env.ts
```diff
@@ -0,0 +1,44 @@
+import { NetworkId } from "./types";
+import { types as TyphonTypes } from "@stricahq/typhonjs";
+
+import testnetCostModels from "./costmodels/testnet-protocol.json";
+
+type ENV_TYPE = {
```
haammar-ledger, line 6: Nitpicking here, but would you agree to reword "env" with something else?
This is to avoid confusion with the global `env.ts` file and actual environment variables.

The data you're handling here closely looks like network parameters in other coins, e.g: https://github.com/LedgerHQ/ledger-live-common/blob/6aebe41709035d227c2682c8617d7f70b374ab10/src/families/bitcoin/types.ts#L76
Maybe you could set it as a `NetworkParameters` type in the types.ts file.

ghost (commented):
Just a first quick review with a couple small comments, didn't have a thorough look yet.

src/families/cardano/costmodels/testnet-protocol.json
```diff
@@ -0,0 +1,170 @@
+{
+  "PlutusScriptV1": {
```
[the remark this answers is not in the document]
pavanvora: These values can be changed at epoch boundary (every 5days) but it's very rare. Do you think it is a good idea to sync those values from preload?

src/families/cardano/serialization.ts
```diff
@@ -0,0 +1,80 @@
...
+  return r;
+}
+
+function toStakeCredentialRaw(r: StakeCredential): StakeCredentialRaw {
+  return r;
+}
+
+function fromStakeCredentialRaw(r: StakeCredentialRaw): StakeCredential {
+  return r;
+}
```
haammar-ledger, line 37: I think these work because by chance `BipPath` and `BipPathRaw` are identical, but it would be safer to do the explicit conversion here.

src/families/cardano/preload.ts
```diff
@@ -0,0 +1,49 @@
...
+}
+
+const updates = new Subject<CardanoPreloadData>();
+
+export function getCurrentCardanoPreloadData(): CardanoPreloadData {
+  return currentPreloadedData;
+}
+
+export function setCardanoPreloadData(data: CardanoPreloadData): void {
+  if (data === currentPreloadedData) return; // TODO: check logic
```
haammar-ledger, line 22: Need to do something here?

src/families/cardano/preload.ts
```diff
@@ -0,0 +1,49 @@
...
+  currentPreloadedData = data;
+  updates.next(data);
+}
+
+export function getCardanoPreloadDataUpdates(): Observable<CardanoPreloadData> {
```
haammar-ledger, line 27: In other implementations (i.e Polkadot) this is only used to provide a React hook for the UI, see here: https://github.com/LedgerHQ/ledger-live-common/blob/develop/src/families/polkadot/react.ts#L16
You'll probably want such a hook to use the preload data.

src/families/cardano/js-broadcast.ts
```diff
@@ -0,0 +1,20 @@
+import type { Operation, SignedOperation } from "../../types";
+import { patchOperationWithHash } from "../../operation";
+import { submitTransaction } from "./api/submitTransaction";
+
+/**
+ * Broadcast the signed transaction
+ * @param {signature: string, operation: string} signedOperation
+ */
+const broadcast = async ({
+  signedOperation: { operation, signatureRaw },
```
haammar-ledger, line 10: You can remove the typing here that is redundant with `SignedOperation` below.

src/families/cardano/js-signOperation.ts
```diff
@@ -0,0 +1,175 @@
...
+        );
+
+        o.next({
+          type: "signed",
+          signedOperation: {
+            operation,
+            signature: signed.payload,
+            signatureRaw: signed,
+            expirationDate: null,
+          },
```
haammar-ledger, line 163: Please type this object `as SignedOperation`, to make sure it has the correct shape. I'm only asking because in other integrations some sneaky bugs were caused by this object not being aligned with what's used in `broadcast`. Unfortunately we lack a way to enforce this implicitly...

src/families/cardano/api/submitTransaction.ts
```diff
@@ -0,0 +1,138 @@
+import BigNumber from "bignumber.js";
+import network from "../../../network";
+import { CARDANO_API_ENDPOINT } from "../constants";
+
+type Token = {
```
haammar-ledger, line 5: Shouldn't all the types here belong in `api-types.ts` ?

src/families/cardano/api/submitTransaction.ts
```diff
@@ -0,0 +1,138 @@
...
+      : undefined,
+    metadata: payload.metadata,
+    certificate: payload.certificate as any,
+    withdrawals: payload.withdrawals as Array<any>,
+  };
+}
+
+export async function submitTransaction(
+  id: string,
+  transaction: string
```
haammar-ledger, line 127: Minor coding style suggestion: it's better to provide a single object param, to avoid silent mismatch errors especially when types are the same.

```
safeSubmitTransaction({
  id,
  transaction
} : {
  id: string,
  transaction: string
}) { ... }
```

ghost (commented):
New round of review, didn't finish yet.

src/families/cardano/js-synchronisation.ts
```diff
@@ -0,0 +1,100 @@
...
+import { encodeAccountId } from "../../account";
+
+import BigNumber from "bignumber.js";
+import Ada from "@cardano-foundation/ledgerjs-hw-app-cardano";
+import { str_to_path } from "@cardano-foundation/ledgerjs-hw-app-cardano/dist/utils";
+import { getCardanoResourseForAccount } from "./api/getAccount";
+import { getOperations } from "./api";
+import { getBaseAddress, getBipPathString } from "./logic";
+
+const postSync = (initial: Account, parent: Account) => parent;
```
haammar-ledger, line 13: This is actually unneeded since it's already provided as a default value [here](https://github.com/LedgerHQ/ledger-live-common/blob/5fa6a917da4a4789d757ae67eb0679f54ce6cd52/src/bridge/jsHelpers.ts).
(It's on us to update the doc and cleanup other integrations)

src/families/cardano/js-getTransactionStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+    false
+  );
+
+  if (!amount.gt(0)) {
+    errors.amount = useAllAmount
+      ? new NotEnoughBalance()
+      : new AmountRequired();
+  } else if (amount.lt(minTransactionAmount)) {
+    errors.amount = new CardanoMinAmountError(
+      `Minimum ${minTransactionAmount.div(1e6)} ADA required`
```
haammar-ledger, line 53: Error messages are defined in live-desktop and live-mobile, e.g:
- https://github.com/LedgerHQ/ledger-live-desktop/blob/615022c001c7858a1c1d2f9d00a716ecd6a41029/static/i18n/en/app.json#L3957-L3959
- https://github.com/LedgerHQ/ledger-live-mobile/blob/3978dfa2cf49e97c1a459ea0943afdb09814b495/src/locales/en/common.json#L396-L398

If you need custom errors than you'll need to open PRs on live-desktop and live-mobile. 
Here you should only provide the data to be used in the message.
From the example in the links you can see that you can provide the `currencyName` param to the `InvalidAddress` error above.

src/families/cardano/js-getTransactionStatus.ts
```diff
@@ -0,0 +1,68 @@
...
+const getTransactionStatus = async (
+  a: Account,
+  t: Transaction
+): Promise<TransactionStatus> => {
+  const errors: Record<string, Error> = {};
+  const warnings: Record<string, Error> = {};
+  const useAllAmount = !!t.useAllAmount;
+
+  const estimatedFees = t.fees || new BigNumber(0);
+  const amount = t.amount;
```
haammar-ledger, line 25: Unless I'm mistaken, I don't see where the `useAllAmount` param is used to get the correct amount.
I'm referring to this kind of logic: https://github.com/LedgerHQ/ledger-live-common/blob/5fa6a917da4a4789d757ae67eb0679f54ce6cd52/src/families/bitcoin/js-getTransactionStatus.ts#L96

src/families/cardano/hw-getAddress.ts
```diff
@@ -0,0 +1,40 @@
...
+
+const resolver: Resolver = async (transport, { path }) => {
+  const spendingPath = getBipPathFromString(path);
+  const stakingPathString = getBipPathString({
+    account: spendingPath.account,
+    chain: StakeChain.stake,
+    index: STAKING_ADDRESS_INDEX,
+  });
+  const ada = new Ada(transport);
+  const r = await ada.deriveAddress({
```
haammar-ledger, line 20: On the issue of displaying the address or not, here's maybe what you can do:
The `Resolver` type has a `GetAddressOptions` param that itself has a `verify` param: https://github.com/LedgerHQ/ledger-live-common/blob/5fa6a917da4a4789d757ae67eb0679f54ce6cd52/src/hw/getAddress/types.ts#L14
This param is already provided as needed by the callers of receive (live desktop and live mobile).

Also from what I can see the ledgerjs-hw-app-cardano helper provides a [showAddress](https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/blob/6ec2cc3fa869ef1ee1e6033792aae80c49962141/src/Ada.ts#L282) function in addition to [deriveAddress](https://github.com/cardano-foundation/ledgerjs-hw-app-cardano/blob/6ec2cc3fa869ef1ee1e6033792aae80c49962141/src/Ada.ts#L265).

So maybe here you can use the `verify` param to just call the appropriate helper function?
I'm just guessing here, to be confirmed. :slightly_smiling_face:

src/families/cardano/hw-getAddress.ts
```diff
@@ -0,0 +1,40 @@
...
+      type: AddressType.BASE_PAYMENT_KEY_STAKE_KEY,
+      params: {
+        spendingPath: str_to_path(path),
+        stakingPath: str_to_path(stakingPathString),
+      },
+    },
+  });
+  return {
+    address: TyphonUtils.getAddressFromHex(r.addressHex).getBech32(),
+    publicKey: "",
```
haammar-ledger, line 35: Does the app provide no pubkey?
This field is used in a few generic places, I think it's best to provide a value if possible.

ghost (commented):
First review finished!

src/families/cardano/hw-getAddress.ts
pavanvora: yup, this is exactly what I did after our last meeting and it is working all good :)

src/families/cardano/hw-getAddress.ts
pavanvora: No, the app only provides the addressHex. We will discuss what can be done there in our next call.

ghost:
Replaced with https://github.com/LedgerHQ/ledger-live/pull/173

PR #1636, opened 2022-01-19
fix peercoin

hzheng-ledger:
Refer to discussion in https://github.com/LedgerHQ/ledgerjs/pull/702

### DIFF
--- a/src/families/bitcoin/wallet-btc/wallet.ts
+++ b/src/families/bitcoin/wallet-btc/wallet.ts
@@ -272,9 +272,7 @@ class BitcoinLikeWallet {
       if (additionals && additionals.includes("peercoin")) {
         // remove timestamp for new version of peercoin input, refer to https://github.com/peercoin/rfcs/issues/5 and https://github.com/LedgerHQ/ledgerjs/issues/701
         const version = i.txHex.substring(0, 8);
-        if (version !== "01000000" && version !== "02000000") {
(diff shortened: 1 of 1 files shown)


src/families/bitcoin/wallet-btc/wallet.ts
```diff
@@ -272,9 +272,7 @@ class BitcoinLikeWallet {
       if (additionals && additionals.includes("peercoin")) {
         // remove timestamp for new version of peercoin input, refer to https://github.com/peercoin/rfcs/issues/5 and https://github.com/LedgerHQ/ledgerjs/issues/701
         const version = i.txHex.substring(0, 8);
-        if (version !== "01000000" && version !== "02000000") {
-          hasTimestamp = false;
-        }
+        hasTimestamp = version === "01000000" || version === "02000000";
```
haammar-ledger, line 275: This always overrides initial value of `params.hasTimestamp` which becomes never used.
Just to mention, but I think this is ok since Peercoin is the only remaining supported coin that's using timestamp...

ghost (approved):

gre (approved):

src/families/bitcoin/wallet-btc/wallet.ts
gre: but only in case of peercoin

PR #1637, opened 2022-01-19
LL-8903 Less blocking lastSeenDevice logic

juan-cortes:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LL-8903

## Description / Usage
 The original work from LL-8903 meant that if we had bad internet connection we would be blocking the openApp action for a few more seconds than needed. This change makes it non blocking, we just request the `getDeviceInfo` from the device, needed for the latest firmware API call, and if there's time to emit the `device-update-last-seen` event we do, if there's no time to emit it then we just continue. This idea of making it non blocking comes from Benito by the way, all kudos to him

Particularly interested in @gre and @valpinkman taking a look at the solution for the rxjs part. I tried with the `takeUntil` we talked about but then it wouldn't emit the inner events of the condition, this seems to work nicely.

## Expectations

### DIFF
--- a/src/hw/connectApp.ts
+++ b/src/hw/connectApp.ts
@@ -1,5 +1,5 @@
 import semver from "semver";
-import { Observable, concat, from, of, throwError, defer } from "rxjs";
+import { Observable, concat, from, of, throwError, defer, merge } from "rxjs";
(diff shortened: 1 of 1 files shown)


gre (approved):
lgtm

PR #1648, opened 2022-01-21
[LL-9060]fix cache bug because of xpub+derivation hash conflict

hzheng-ledger:


gre (approved):
lgtm. bot confirm all still work. CI is :x: but only on ETH like on master.

PR #1651, opened 2022-01-24
LL-9178 Bitcoin js cacherefactoring

hzheng-ledger:
Refactoring for address cache and unit test added


src/families/bitcoin/wallet-btc/crypto/bitcoincash.ts
```diff
@@ -7,16 +7,16 @@ import bchaddr from "bchaddrjs";
 import { toOutputScript } from "bitcoinjs-lib/src/address";
 import { InvalidAddress } from "@ledgerhq/errors";
 import { DerivationModes } from "../types";
-import { ICrypto } from "./types";
 import Base from "./base";
 
 // a mock explorer class that just use js objects
-class BitcoinCash implements ICrypto {
+class BitcoinCash extends Base {
```
haammar-ledger, line 13: Did you check that all the default methods implemented in `Base` are relevant for all the coins you modified?

hzheng-ledger: good point.
I just changed the "Base" class.
In "Base" class, it has following public or protected methods: "constructor", "getPubkeyAt", "getLegacyAddress", "getAddress", "customGetAddress", "getDerivationMode", "toOutputScript", "validateAddress", "isTaprootAddress".
They are relevant for all the coins that inherited "Base" class, IMO

Two private methods in "Base": "getSegWitAddress", "getNativeSegWitAddress". They are not relevant for all the coins. I think it should be OK as they are private.

gre:
make sure the title have a LL- ticket so we can know the context. thanks

ghost (approved):
Hard to assess impact just from the code, but good thing to have added a test :+1: and the bot shows no related errors (only known issues with pending UTXOs)

hzheng-ledger:
> make sure the title have a LL- ticket so we can know the context. thanks
I had a ticket for that
https://ledgerhq.atlassian.net/browse/LL-9178

PR #1654, opened 2022-01-25
Fix tests

alexalouit:
Fix deep equality in tests
```
expect(received).toBe(expected) // Object.is equality
If it should pass with deep equality, replace "toBe" with "toStrictEqual"
```

## Expectations


ghost (approved):
:+1:

PR #1655, opened 2022-01-25
Restore CI

gre:


haammar-ledger:
Need to regen `all.libcore.ts.snap`?
Some solana snapshots will be removed, but there are also some diff in ethereum operations?... :thinking:

haammar-ledger:
Still the ETH operations diff :open_mouth: I don't really get it...

gre:
ETH diff is a backend problem discussed in slack @haammar-ledger

PR #1661, opened 2022-01-28
Revert "LL-8975 bump stellar (#1644)"

gre:
due to bug complication, it is safer to revert this part, making master stable again
while we will need to address bugs and ideally reveal them with test if possible


ghost (changes requested):
CI issue with the `yarn.lock`.

gre:
hodling this because it's better to try to fix possible bugs.

PR #1662, opened 2022-01-28
enable solana tests

gre:
attempt to restore solana tests


gre:
You'll need to Regen snapshot because I didn't put the correct one I think

gre:
will do another PR.

PR #1665, opened 2022-01-31
Cosmos JS integration

ghost:
## Expectations

### DIFF
--- a/src/env.ts
+++ b/src/env.ts
@@ -156,7 +156,7 @@ const envDefinitions = {
     desc: "location of the compound API",
   },
   COSMOS_GAS_AMPLIFIER: {
-    def: 4,
+    def: 1.6,
     parser: intParser,
     desc: "estimate gas multiplier",
   },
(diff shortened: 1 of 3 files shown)


src/env.ts
```diff
@@ -156,7 +156,7 @@ const envDefinitions = {
     desc: "location of the compound API",
   },
   COSMOS_GAS_AMPLIFIER: {
-    def: 1.4,
+    def: 4,
```
haammar-ledger, line 159: Is it wanted to revert this value? Why?

ghost (commented):
Looks good, just one question.

src/env.ts
wa-aal: Fix in 92f4d599d10d0f6dad8c52c0035104a4a296159a.
This is a workarround for mismatched fees estimation calculation (simulation and broadcast are on different endpoints)

gre:
ok let's let the bot runs with this 🙏

PR #1679, opened 2022-02-03
Tezos: add storage limit

henri-ly:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-9137

## Description / Usage

When we didn't set storage limit, so new account with low balance got an error storage_exhaust error that wasn't catch.

https://tezos.stackexchange.com/questions/679/transaction-failing-with-proto-003-psddfki3-storage-exhausted-operation#comment751_680

## Expectations

### DIFF
--- a/src/families/tezos/bridge/js.ts
+++ b/src/families/tezos/bridge/js.ts
@@ -1,6 +1,10 @@
 // @flow
 import { BigNumber } from "bignumber.js";
-import { TezosToolkit, DEFAULT_FEE } from "@taquito/taquito";
+import {
+  TezosToolkit,
+  DEFAULT_FEE,
+  DEFAULT_STORAGE_LIMIT,
+} from "@taquito/taquito";
 import {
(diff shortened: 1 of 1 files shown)


src/families/tezos/bridge/js.ts
```diff
@@ -201,6 +201,7 @@ const prepareTransaction = async (
           mutez: true,
           to: transaction.recipient,
           amount: amount.toNumber(),
+          storageLimit: 257, // https://github.com/TezTech/eztz/blob/master/PROTO_003_FEES.md for originating an account
```
gre, line 204: can you use a constant from taquito instead? like we did for `DEFAULT_FEE.REVEAL`
probably this: https://github.com/ecadlabs/taquito/blob/48a11cfaffe4c6bdfa6f04ebbd0b756f4b135865/packages/taquito/src/constants.ts#L15

gre (approved):

PR #1687, opened 2022-02-04
LL-9121 [Algorand] Update xpub during sync to handle migration from libcore

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LL-9121
_"LLD - Send: Optimistic operation is missing. Need at least one confirmation to see the operation in the history"_

## Description / Usage

When an Algorand account was migrated from libcore to JS, its recorded xpub was not updated, causing a mismatch with the generated account ID, causing the account to be flagged as `needClear` [here](https://github.com/LedgerHQ/ledger-live-common/blob/f8d0ce9be8dd807ecaf872fdd385ff6d4cb63d59/src/bridge/jsHelpers.ts#L165), causing the `pendingOperations` to be erased, causing the issue described in the ticket.
The one-line fix is to update the xpub as well during sync.
(also includes some zero-impact cleanup)

## Expectations

### DIFF
--- a/src/families/algorand/buildTransaction.ts
+++ b/src/families/algorand/buildTransaction.ts
@@ -28,8 +28,7 @@ export const buildTransactionPayload = async (
 
   const note = memo ? new TextEncoder().encode(memo) : undefined;
 
-  // Minor hack: cast the result as a algosdk.SuggestedParams, while it's actually a TransactionParametersResponse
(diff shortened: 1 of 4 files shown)


gre (changes requested):
linter to fix. good otherwise.

gre (approved):

PR #1688, opened 2022-02-07
fix build

hzheng-ledger:
Build was broken because of rlp version 2.2.6 and version 3.0.0 has different encode signature and we should explicitly specify version number in package.jon


gre (approved):

gre:
apart from TRON flawkiness, all good on linux, enough to merge.

PR #1692, opened 2022-02-10
LIVE-893 - Feature flag learn page

thomasrogerlux:
## Context (issues, jira)

[LIVE-893]

## Description / Usage

The new learn section that will be shipped both in LLD and LLM will need feature flagging to simplify its development workflow. In this PR, I'm adding the learn feature to the Typescript types from the feature flagging module of LLC.

I'm also fixing a quick bug in the FeatureToggle component when we don't pass a fallback or children props

## Expectations

[LIVE-893]: https://ledgerhq.atlassian.net/browse/LIVE-893?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/featureFlags/FeatureToggle.tsx
+++ b/src/featureFlags/FeatureToggle.tsx
@@ -17,8 +17,8 @@ export const FeatureToggle = ({
   const feature = useFeature(featureId);
 
   if (!feature || !feature.enabled) {
-    return fallback;
+    return fallback ?? null;
   }
 
-  return children;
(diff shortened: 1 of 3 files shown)


LFBarreto (approved):

juan-cortes (approved):
I trust you

PR #1700, opened 2022-02-14
LIVE-886 (Market): add sparkline variation + fix issue in countervalue initial value

LFBarreto:
- (Market): add svg sparkline variation + fix issue in countervalue initial setting

## Context (issues, jira)

[LIVE-886]

## Expectations

[LIVE-886]: https://ledgerhq.atlassian.net/browse/LIVE-886?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/market/api/api.ts
+++ b/src/market/api/api.ts
@@ -45,11 +45,11 @@ async function setSupportedCoinsList(): Promise<SupportedCoins> {
 
 const matchSearch =
   (search: string) =>
-  (currency: MarketCoin): boolean => {
-    if (!search) return false;
-    const match = `${currency.symbol}|${currency.name}`;
-    return match.toLowerCase().includes(search.toLowerCase());
-  };
+    (currency: MarketCoin): boolean => {
+      if (!search) return false;
+      const match = `${currency.symbol}|${currency.name}`;
+      return match.toLowerCase().includes(search.toLowerCase());
+    };
 
 function distributedCopy(items: number[], n: number): number[] {
   const elements = [items[0]];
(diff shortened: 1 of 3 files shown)


src/market/api/api.ts
```diff
@@ -83,6 +83,7 @@ function sparklineAsSvgData(points: number[]): SparklineSvgData {
       })
       .join(" "),
     viewBox: `0 0 ${totalXSteps} ${sparklineYHeight + 3}`,
+    isPositive: points[0] <= points[points.length - 1],
```
haammar-ledger, line 86: Can `points` be empty here? (noob question I have 0 context on market feature)

LFBarreto: good point should be handled now :+1:

src/market/api/api.ts
```diff
@@ -52,6 +52,8 @@ const matchSearch =
   };
 
 function distributedCopy(items: number[], n: number): number[] {
+  if (!items) return [];
+  if (items.length <= n) return items;
```
haammar-ledger, line 56: This safety looks good for the `distributedCopy` function :+1: but does it help for the `sparklineAsSvgData` function below?

LFBarreto: its always called with data provided by this function so in this case yes :sweat_smile:

LFBarreto: line 185
```js

 sparklineIn7d: currency?.sparkline_in_7d?.price
        ? sparklineAsSvgData(
            distributedCopy(currency.sparkline_in_7d.price, 6 * 7)
          ) // keep 6 points per day
        : null,
```

ghost (approved):

PR #1717, opened 2022-02-19
Stellar: add support for non-native assets + custom fee

quietbits:
Please let us know what tests are missing or we should update.


src/families/stellar/logic.ts
```diff
@@ -216,17 +226,73 @@ export const isAddressValid = (address: string): boolean => {
...
+  id: string | null;
+  isMuxedAccount: boolean;
+  assetIds: string[];
+}> => {
+  if (!address) {
+    return {
+      id: null,
+      isMuxedAccount: false,
+      assetIds: [],
+    };
```
haammar-ledger, line 250: Maybe better to return `null` in these cases?

src/families/stellar/js-signOperation.ts
```diff
@@ -7,21 +7,20 @@ import { open, close } from "../../hw";
...
+
   const operation: Operation = {
     id: `${account.id}--OUT`,
     hash: "",
     type: "OUT",
-    value:
-      transaction.useAllAmount && transaction.networkInfo
-        ? account.balance.minus(transaction.networkInfo.baseReserve).minus(fees)
-        : transaction.amount.plus(fees),
+    value: getAmountValue(account, transaction, fees),
```
haammar-ledger, line 23: It seems this changes the computation in the native case?

src/families/stellar/js-getTransactionStatus.ts
```diff
@@ -32,86 +42,170 @@ const getTransactionStatus = async (
...
+  maxAmount: BigNumber;
   totalSpent: BigNumber;
 }> => {
   const errors: Record<string, Error> = {};
   const warnings: Record<string, Error> = {};
   const useAllAmount = !!t.useAllAmount;
 
+  const destinationNotExistMessage =
+    new NotEnoughBalanceBecauseDestinationNotCreated("", {
+      minimalAmount: "1 XLM",
```
haammar-ledger, line 54: Don't you want to use a variable instead of hardcoding 1? (maybe BASE_RESERVE if it's related?)

src/families/stellar/js-getTransactionStatus.ts
```diff
@@ -124,6 +218,7 @@ const getTransactionStatus = async (
     warnings,
     estimatedFees,
     amount,
+    maxAmount,
```
haammar-ledger, line 221: Sorry but you can't do this, `getTransactionStatus` is part of the `AccountBridge` contract and has a specified signature: https://github.com/LedgerHQ/ledger-live-common/blob/ef304028b3ce0c021b6c75f4a5ade38ccaec29c1/src/types/bridge.ts#L106

From what I can see you only use `maxAmount` in the `estimateMaxSpendable` function, but it already calls `getTransactionStatus` with a fake transaction that has `useAllAmount: true` : https://github.com/LedgerHQ/ledger-live-common/blob/ef304028b3ce0c021b6c75f4a5ade38ccaec29c1/src/families/stellar/js-estimateMaxSpendable.ts#L25
So why not just use the returned `amount` as done currently?

src/families/stellar/js-getTransactionStatus.ts
```diff
@@ -32,86 +42,170 @@ const getTransactionStatus = async (
...
+    } else {
+      // Native payment
+      maxAmount = nativeAmountAvailable;
+      amount = useAllAmount ? maxAmount : t.amount || 0;
+
+      if (amount.gt(maxAmount)) {
+        errors.amount = new NotEnoughBalance();
+      }
+
+      // TODO: ??? do we need to include fee in total?
```
haammar-ledger, line 166: Yes `totalSpent` is intended to reflect the total amount that is taken from the account.
It just needs to be used consistently in subsequent tests, which I think is the case in the current implementation.

src/families/stellar/js-estimateMaxSpendable.ts
```diff
@@ -25,7 +25,7 @@ const estimateMaxSpendable = async ({
     useAllAmount: true,
   });
   const s = await getTransactionStatus(mainAccount, t);
-  return s.amount;
+  return s.maxAmount;
```
haammar-ledger, line 28: Why not just `s.amount`? See previous comment in js-getTransactionStatus.ts.

src/families/stellar/api/horizon.ts
```diff
@@ -40,21 +49,46 @@ StellarSdk.HorizonAxiosClient.interceptors.response.use((response) => {
 });
 
 const getFormattedAmount = (amount: BigNumber) => {
-  return amount
-    .div(new BigNumber(10).pow(currency.units[0].magnitude))
-    .toString(10);
+  // TODO: ??? not sure why we need this conditional, but it is more consistent
+  // with it. Otherwise, some transactions would just freeze on broadcasting at
+  // the very end.
```
haammar-ledger, line 54: Very weird, not something we are seeing currently with the existing code.
Would be to investigate a bit the broadcast failure to understand the link with this.

ghost (commented):
Looking good so far! :+1: 

- My main concern is the removal of the incremental sync mechanism, see related comment.
- The `helpers` folder doesn't do harm, but many helpers are already defined in the logic.ts file, maybe you want to keep them all in a single place for consistency?

haammar-ledger:
> Please let us know what tests are missing or we should update.

Regarding tests, there are 2 main areas:
- You should update the [test-dataset.ts](https://github.com/LedgerHQ/ledger-live-common/blob/develop/src/families/stellar/test-dataset.ts) file. This file provides data to be tested in a static way, you can add new account(s) at the end, they should contain tokens and new types transactions to be tested, but they should also be frozen accounts that never change, so the test can reliably run on them.
- You should update the [specs.ts](https://github.com/LedgerHQ/ledger-live-common/blob/develop/src/families/stellar/specs.ts) file which contains the data used by the "bot" i.e our automated transaction testing tool. Here you should only define new transactions cases that are relevant to test with this new implementation.

If this isn't very clear, please refer to the (perfectible) doc we have at the moment: https://developers.ledger.com/docs/coin/run-ledger-live/#test-dataset and be sure to check examples in the many other coins.

src/families/stellar/logic.ts
```diff
@@ -131,16 +142,12 @@ const getValue = (
...
-          ? parseCurrencyUnit(currency.units[0], operation.amount)
-          : new BigNumber(0);
-
-      if (type === "OUT") {
-        value = value.plus(transaction.fee_charged);
-      }
+      return parseCurrencyUnit(currency.units[0], operation.amount);
 
-      return value;
+    // TODO: ??? do we need to include fee in the amount?
```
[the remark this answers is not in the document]
quietbits: :+1: The thing to note about the Stellar network fee is that it’s more like a max user is willing to pay. For example, if someone sets the fee to 1 XLM, but it takes only 0.00001 XLM to process, the final fee will be 0.00001 XLM.

src/families/stellar/logic.ts
```diff
@@ -216,17 +226,73 @@ export const isAddressValid = (address: string): boolean => {
...
   addr: string,
   accountId: string
 ): Promise<Operation[]> => {
+  const supportedOperationTypes = [
+    "create_account",
+    "payment",
+    "path_payment_strict_send",
+    "path_payment_strict_receive",
+    "change_trust",
+  ];
```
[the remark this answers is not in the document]
quietbits: TS type casting is tricky when using enum for supported values in this case. It seemed cleaner to use an array. If you feel strongly about this, we can use the enum and cast operation type to `any` (that might be the only solution to make TS happy).

src/families/stellar/js-signOperation.ts
quietbits: The native amount is calculated differently, because of the required reserve and fees. Non-native asset balances don’t have these requirements (for non-native transactions we need to check asset amount + native (XLM) amount to make sure there is enough to cover the transaction). Was that your question?

src/families/stellar/api/horizon.ts
quietbits: Couldn’t reproduce this again, it must have been caused by work in progress.

src/families/stellar/js-signOperation.ts
haammar-ledger: My concern is mainly avoiding regressions, I was just asking about this change:
```diff
-     value:
-      transaction.useAllAmount && transaction.networkInfo
-        ? account.balance.minus(transaction.networkInfo.baseReserve).minus(fees)
-        : transaction.amount.plus(fees)

compared to getAmountValue:
+   return transaction.useAllAmount && transaction.networkInfo
+    ? BigNumber.max(account.spendableBalance.minus(fees), 0)
+   : transaction.amount;
```

src/families/stellar/logic.ts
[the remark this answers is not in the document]
haammar-ledger: Nope it's fine :ok_hand:

src/families/stellar/js-synchronization.ts
```diff
@@ -1,35 +1,114 @@
...
+  ).toString();
+
+  const newOperations =
+    (await fetchOperations({
+      accountId,
+      addr: address,
+      // To fetch new records, use "desc" + "now"
+      // To fetch latest records, use "asc" + paging token
+      order: lastPagingToken === "0" ? "desc" : "asc",
+      cursor: lastPagingToken === "0" ? "now" : lastPagingToken,
```
haammar-ledger, line 75: Wouldn't `asc + paging token` work in all situations? Unless I missed something `desc + now` returns the same operations as `asc + 0` right? With only the order being reversed, which doesn't matter if you are fetching everything.

src/families/stellar/logic.ts
```diff
@@ -97,11 +106,14 @@ export const formatOperation = async (
...
-      ? {
-          memo,
-        }
-      : {},
+    extra: {
+      pagingToken: rawOperation.paging_token,
+      // @ts-expect-error check transaction_successful property
+      assetCode: rawOperation?.asset_code,
+      // @ts-expect-error check transaction_successful property
+      assetIssuer: rawOperation?.asset_issuer,
```
haammar-ledger, line 114: No way to address these TS errors?

src/families/stellar/js-getTransactionStatus.ts
```diff
@@ -38,80 +48,165 @@ const getTransactionStatus = async (
...
+        !recipientAccount.assetIds.includes(`${t.assetCode}:${t.assetIssuer}`)
+      ) {
+        errors.recipient = new StellarAssetNotAccepted("", {
+          assetCode: t.assetCode,
+        });
+      }
+
+      const assetBalance = asset?.balance || new BigNumber(0);
+
+      // @ts-expect-error check spendableBalance property
```
haammar-ledger, line 151: This doesn't seem required? Is there really a TS error here?

ghost (commented):
Can you please consider moving the helpers function into logic.ts ? We don't really need 1 single file for each.

src/families/stellar/js-synchronization.ts
quietbits: You're right. I missed that we're fetching the whole operations history on the initial load. Thanks for catching this!

src/families/stellar/js-signOperation.ts
quietbits: I don't think we need to add the fee to the amount for XLM because it's submitted separately. If I'm sending someone 1 XLM and set the fee 0.5 XLM, the total amount I'm sending should still be 1 XLM, and max fee paid will be 0.5 XLM (could be less). Is that your concern?

ghost:
Replaced with https://github.com/LedgerHQ/ledger-live/pull/210

PR #1721, opened 2022-02-19
Cosmos JS

ghost:
## Expectations


src/families/cosmos/js-synchronisation.ts
```diff
@@ -42,56 +42,87 @@ const txToOps = (info: any, id: string, txs: any): Operation[] => {
...
-          op.extra.validators.push({
-            amount: attributes["amount"].replace(currency.units[1].code, ""),
-            address: attributes.destination_validator,
-          });
-          op.extra.cosmosSourceValidator = attributes.source_validator;
+          if (
+            attributes["amount"] &&
+            attributes["amount"].indexOf(currency.units[1].code) != -1 &&
+            attributes.destination_validator &&
+            attributes.source_validator
```
haammar-ledger, line 101: We should avoid mixing styles, `attributes.xxx` should be preferred over `attributes["xxx"]` where possible.

src/families/cosmos/js-signOperation.ts
```diff
@@ -158,19 +163,27 @@ const signOperation = ({
             ? "REWARD"
             : "OUT";
 
+        const senders = [] as any;
+        const recipients = [] as any;
```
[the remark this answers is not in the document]
wa-aal: fix in 744633738c87a036ba1ee816ddec272b4c81e333

src/families/cosmos/js-synchronisation.ts
wa-aal: fix in 892148b6dc81a8928bb10c76699f4bd6f00c2bf9

PR #1723, opened 2022-02-21
LIVE-825 - Update market API to have an AND logic on filters instead of OR logic

thomasrogerlux:
## Context (issues, jira)

[LIVE-825]

## Description / Usage

Before you couldn't do a star filtering + live compatible filtering at the same time (same with search), because the different filters were acting like OR not AND.

## Expectations

No need to update LLD/LLM but it fixes the filtering logic of the market

[LIVE-825]: https://ledgerhq.atlassian.net/browse/LIVE-825?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/market/api/api.ts
+++ b/src/market/api/api.ts
@@ -105,18 +105,26 @@ async function listPaginated({
 }: MarketListRequestParams): Promise<CurrencyData[]> {
   let ids = _ids;
 
-  if (liveCompatible) {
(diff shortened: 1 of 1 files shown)


henrily-ledger:
LGTM

henri-ly (approved):

PR #1724, opened 2022-02-21
LIVE-1443 LL-1170 Improve Bitcoin performance

hzheng-ledger:
1. use react-native-secp256k1 lib to calculate elliptic curve in async mode
2. get rid of unnecessary lib for zcash
3. add index to improve getLastTx


src/families/bitcoin/wallet-btc/crypto/bip32.ts
```diff
@@ -28,7 +29,13 @@ class BIP32 {
     const I = createHmac("sha512", this.chainCode).update(data).digest();
     const IL = I.slice(0, 32);
     const IR = I.slice(32);
-    const Ki = Buffer.from(publicKeyTweakAdd(this.publicKey, IL));
+    const Ki = Buffer.from(
+      typeof navigator !== "undefined" && navigator.product === "ReactNative"
```
haammar-ledger, line 33: Just wondering, would it not be good to have the new code on all platforms? Desktop and mobile?

src/families/bitcoin/wallet-btc/crypto/secp256k1.ts
```diff
@@ -0,0 +1,420 @@
+/* eslint-disable */
+const BigInt = require('big-integer');
```
haammar-ledger, line 2: `big-integer` is currently available only through other dependencies, maybe we should add it to package.json?

Also I think we often had issues with bigint polyfills in the past... can you please make sure to test on both Android and iOS?

src/families/bitcoin/wallet-btc/crypto/bip32.ts
hzheng-ledger: no, the lib "secp256k1" is faster in LLD because of C++ but it is very slow in LLM as it takes js implementation as fallback plan：
so the lib "noble-secp256k1"(the new one added by me) is used by LLM only

gre: the way we "feature detect" ReactNative here is not really safe to me. it could be better to introduce a "implementation" concept and a "setImplementation" that the client side will have to call.

gre: this idea of having a separate setImplementation could come when we will really need it for when we inject `react-native-secp256k1` – meanwhile it's acceptable to take this quick win with the embedded branching – it could be good to add a comment on the line of code with:

```
// FIXME LIVE-1443
```

( see https://ledgerhq.atlassian.net/browse/LIVE-1443 )

src/families/bitcoin/wallet-btc/crypto/zec.ts
```diff
@@ -74,7 +71,20 @@ class ZCash extends Base {
 
   // eslint-disable-next-line class-methods-use-this
   validateAddress(address: string): boolean {
-    return zec.Address.isValid(address, "livenet");
+    try {
+      const version = Number(
+        "0x" + bs58check.decode(address).slice(0, 2).toString("hex")
```
haammar-ledger, line 76: You can add a comment to document this

ghost (commented):
Please update yarn.lock

src/families/bitcoin/wallet-btc/storage/index.ts
```diff
@@ -183,14 +206,23 @@ class BitcoinLikeStorage implements IStorage {
...
     addressCache: { [key: string]: string };
   }) {
     this.txs = data.txs;
     this.primaryIndex = data.primaryIndex;
+    this.accountIndex = data.accountIndex;
     this.unspentUtxos = data.unspentUtxos;
     this.addressCache = data.addressCache;
     Base.addressCache = { ...Base.addressCache, ...this.addressCache };
+    if (
+      (!this.accountIndex || _.isEmpty(this.accountIndex)) &&
```
haammar-ledger, line 220: Can you please try import isEmpty instead of _ from lodash ?

haammar-ledger:
Looks good :+1: but needs to test manually all BTC family on iOS + Android.

package.json
```diff
@@ -68,6 +68,7 @@
     "@ledgerhq/json-bignumber": "^1.1.0",
     "@ledgerhq/live-app-sdk": "^0.2.0",
     "@ledgerhq/logs": "6.10.0",
+    "@ledgerhq/react-native-secp256k1": "^1.0.6",
```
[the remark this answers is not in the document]
gre: here is a basic idea of what i mean (quick draft):

LLM

```js
import { setCryptoImpl } from "@ledgerhq/live-common/lib/some/path/crypto-impl";
import Something from "@ledgerhq/react-native-secp256k1";

setCryptoImpl({
  // well i don't know the contract of the crypto function you need, but basically give this impl an object OR just a function that would use Something. obviously it's async
})
```

LL Common

```js
// file src/some/path/crypto-impl.js

let impl = ...; here there need to be a "default" implementation, that JS would use basically. so our impl technically works in web for instance

// this allows to override default impl. LLM can inject its own impl, but LLD could have its own too (node.js one for instance)
export function setCryptoImpl (other) {
  impl = other;
}

export function yourCryptoFunction (...) {
// just use impl to "proxy" whatver you need
}
```

src/families/bitcoin/wallet-btc/crypto/secp256k1.ts
```diff
@@ -0,0 +1,13 @@
...
+let secp256k1Instance = secp256k1;
+
+export function setSecp256k1Instance(secp256k1: any): void {
+  secp256k1Instance = secp256k1;
+}
+
+export async function publicKeyTweakAdd(
+  publicKey: Uint8Array,
+  tweak: Uint8Array
+): Promise<any> {
```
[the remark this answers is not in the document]
hzheng-ledger: I've updated to Promise<number[]>. thanks

package.json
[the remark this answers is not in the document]
hzheng-ledger: fixed

src/families/bitcoin/wallet-btc/storage/index.ts
hzheng-ledger: fixed

src/families/bitcoin/wallet-btc/crypto/zec.ts
hzheng-ledger: fixed

gre (approved):

src/families/bitcoin/wallet-btc/crypto/zec.ts
haammar-ledger: where? :smile:

src/families/bitcoin/wallet-btc/storage/index.ts
```diff
@@ -20,23 +23,35 @@ class BitcoinLikeStorage implements IStorage {
...
     // eslint-disable-next-line @typescript-eslint/ban-ts-comment
     // @ts-ignore
-    const tx: TX | undefined = findLast(this.txs, (t) => {
-      return (
-        (!has(txFilter, "account") || t.account === txFilter.account) &&
-        (!has(txFilter, "index") || t.index === txFilter.index) &&
-        (!has(txFilter, "address") || t.address === txFilter.address) &&
-        (!has(txFilter, "confirmed") ||
+    const tx: TX | undefined = findLast(
+      this.accountIndex[`${txFilter.account}-${txFilter.index}`].map(
```
gre, line 36: please avoid all the repetition maybe with a `hashAccountIndex(tx)` funciton shared across every parts.

gre, line 36: please avoid all the repetition maybe with a `hashAccountIndex(tx)` funciton shared across every parts.

src/families/bitcoin/wallet-btc/crypto/zec.ts
gre: why is there something specific to zcash on the "performance" topic? how should QA test this? thanks

gre: why is there something specific to zcash on the "performance" topic? how should QA test this? thanks

PR #1725, opened 2022-02-21
Cosmos JS

ghost:
## Expectations


src/families/cosmos/js-synchronisation.ts
```diff
@@ -112,21 +108,26 @@ const txToOps = (info: any, id: string, txs: any): Operation[] => {
...
           }
           break;
       }
     });
 
+    if (!["IN", "OUT"].includes(op.type)) {
+      op.senders = [];
+      op.recipients = [];
+    }
+
```
haammar-ledger, line 130: Isn't it equivalent to initializing senders and recipients to [] when creating the op? (L28 & L29)
(Unless I'm missing some case where it's not correct?)

PR #1735, opened 2022-02-24
LIVE-590 Usage of setAllowAutoDisconnect

juan-cortes:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LIVE-590

Please refer to https://github.com/LedgerHQ/ledger-live-desktop/pull/4764 for a deeper explanation. This latest version is decoupled from a release cycle of the transport itself and instead relies of the concept of transport modules introduced by @gre so we only need to make sure that LLD uses the right version that transport and a version of live-common including this work for it to work.

### DIFF
--- a/src/hw/deviceAccess.ts
+++ b/src/hw/deviceAccess.ts
@@ -14,7 +14,7 @@ import {
   DeviceHalted,
 } from "@ledgerhq/errors";
 import { getEnv } from "../env";
-import { open, close } from ".";
+import { open, close, setAllowAutoDisconnect } from ".";
 
 export type AccessHook = () => () => void;
 
@@ -88,12 +88,14 @@ export const withDevice =
       let sub;
       const deviceQueue = deviceQueues[deviceId] || Promise.resolve();
 
-      const finalize = (transport, cleanups) =>
-        close(transport, deviceId)
+      const finalize = (transport, cleanups) => {
+        setAllowAutoDisconnect(transport, deviceId, true);
+        return close(transport, deviceId)
           .catch(() => {})
           .then(() => {
             cleanups.forEach((c) => c());
           });
+      };
 
       // when we'll finish all the current job, we'll call finish
       let finish;
@@ -105,6 +107,7 @@ export const withDevice =
       deviceQueue
         .then(() => open(deviceId)) // open the transport
         .then(async (transport) => {
+          setAllowAutoDisconnect(transport, deviceId, false);
           if (unsubscribed) {
             // it was unsubscribed prematurely
             return finalize(transport, [finish]);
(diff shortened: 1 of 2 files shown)


src/hw/deviceAccess.ts
```diff
@@ -88,12 +88,14 @@ export const withDevice =
       let sub;
       const deviceQueue = deviceQueues[deviceId] || Promise.resolve();
 
-      const finalize = (transport, cleanups) =>
-        close(transport, deviceId)
+      const finalize = (transport, cleanups) => {
+        transport.setAllowAutoDisconnect(true);
```
gre, line 92: not all transport have this method 🤔 sounds like it will break for some transport implementation then. i'm also curious why it passes the type system too

src/hw/deviceAccess.ts
```diff
@@ -105,6 +107,7 @@ export const withDevice =
       deviceQueue
         .then(() => open(deviceId)) // open the transport
         .then(async (transport) => {
+          transport.setAllowAutoDisconnect(false);
```
gre, line 110: same

gre (changes requested):

gre (approved):

juan-cortes:
Requesting review again since we changed the paradigm @gre

tlaforge-ledger (approved):

gre (approved):

src/hw/index.ts
```diff
@@ -77,6 +84,22 @@ export const close = (
...
+  for (let i = 0; i < modules.length; i++) {
+    const m = modules[i];
+    const p =
+      m.setAllowAutoDisconnect &&
+      m.setAllowAutoDisconnect(transport, deviceId, allow);
+    if (p) return p;
+  }
+
+  // fallback on an actual close
+  return transport.close();
```
gre, line 101: i'm not sure about this. it changes the previous behavior: 

in the 

```
 deviceQueue
        .then(() => open(deviceId)) // open the transport
        .then(async (transport) => {
```
        
        case we were NOT doing a close(). it is important not to close() the ble transport of mobile for perf.

gre (dismissed):
actually maybe an issue

src/hw/index.ts
juan-cortes: No, you are absolutely right, this is not correct.
I copy pasted the code from `close` as a base and left that there. This method should only do something if the transport is implementing it, if not, it should be innocuous. Good catch

juan-cortes:
You were right @gre that was leftover code from copying `close` implementation 🤦🏼 
Please take a look again

juan-cortes:
Re @gre

juan-cortes:
🙏🏼 @gre

gre (approved):

gre:
seen with @juan-cortes . QA is ok. we can merge.

PR #1741, opened 2022-02-24
LIVE-1476 Add unit conversion for Polkadot getMinimumBondBalance

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1476

## Description / Usage

Without warning Polkadot changed the unit of the data returned by the minNominatorBond query, from PLANCK to DOT:

- Before: 1600000000000
- Now: 160

Add a dumb conversion so that the data remains consistent with other fetched values, which are all still in PLANCK...

## Expectations


haammar-ledger:
Issue fixed by Polkadot, no need for that fix on our side anymore.

PR #1742, opened 2022-02-25
HOTFIX LIVE-1469 Handle onboarded devices in device actions

valpinkman:
Hotfix provided by @juan-cortes #1738

### DIFF
--- a/package.json
+++ b/package.json
@@ -8,7 +8,7 @@
     "type": "git",
     "url": "https://github.com/LedgerHQ/ledger-live-common"
   },
-  "version": "21.33.0",
+  "version": "2.33.1",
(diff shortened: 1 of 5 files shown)


package.json
```diff
@@ -8,7 +8,7 @@
     "type": "git",
     "url": "https://github.com/LedgerHQ/ledger-live-common"
   },
-  "version": "21.33.0",
+  "version": "2.33.1",
```
gre, line 11: - bad version
- we should only bump from master not in a PR ?

gre (changes requested):
the fix itself is good but the live-common version that i see in the PR is glitched.

PR #1745, opened 2022-02-25
increase the CI tests timeout

gre:
CI can be slow to run these test for a few reasons and we need to increase it

may help some failing cases of bitcoin/ethereum timeout in dataset tests


gre:
something really fishy is going on with Jest tbh.

gre:
not helping, the timeout makes things works and it's likely a jest parallelisation problem

PR #1746, opened 2022-02-25
Cosmos JS

ghost:
## Expectations


src/families/cosmos/js-prepareTransaction.ts
```diff
@@ -121,9 +121,10 @@ const prepareTransaction = async (
   if (
     transaction.memo !== memo ||
     transaction.fees !== fees ||
-    transaction.gas !== gas
+    transaction.gas !== gas ||
+    transaction.amount !== amount
```
gre, line 125: correct me if i'm wrong but I think amount is a big number so the `!==` should be `! .eq()` no?

gre (approved):

PR #1747, opened 2022-02-25
LIVE-1157 New scan accounts mode

ghost:
## Context (issues, jira)

:information_source: **This PR makes changes on the coin integration common framework, a review is especially needed to validate the design.**

https://ledgerhq.atlassian.net/browse/LIVE-1157

## Description / Usage

As described in the JIRA issue above, Hedera has 2 specificities that are not supported by LL right now:
- The mapping from a public key (retrieved from the Nano app) to addresses can't be done programmatically – addresses need to be fetched from a backend.
- 1 public key is associated to n Hedera addresses

To address these needs, this PR extends the way `scanAccounts` works, by allowing to provide a `getAddressesFromPubkey` function which is responsible for deducing up to n addresses, from the pubkey that's retrieved from the Nano.

We also need a different loop over accounts to be added, than the current one:
- current mode : discovery of up to 255 accounts
- new mode : create an account for every already fetched address

Finally, because of the special design of Hedera described above, the check for the right device can't be done on the freshAddress alone, because the address can't be deduced from the device alone.
So this PR relaxes the test to match either the account's freshAddress or seedIdentifier, instead of just the freshAddress.

## Breaking changes & how to migrate after this is merged
Bascially this, in your js-synchronisation.ts file:
```diff
- const postSync = (initial: Account, parent: Account) => parent;

- export const scanAccounts = makeScanAccounts(getAccountShape);
- export const sync = makeSync(getAccountShape, postSync);
+ export const scanAccounts = makeScanAccounts({ getAccountShape });
+ export const sync = makeSync({ getAccountShape });
```

## Expectations


src/bridge/jsHelpers.ts
```diff
@@ -42,6 +42,8 @@ import type { Result, GetAddressOptions } from "../hw/getAddress/types";
 import { open, close } from "../hw";
 import { withDevice } from "../hw/deviceAccess";
 
+export type GetAddressesFromPubkey = (pubkey: string) => Promise<Array<string>>;
```
gre, line 45:
```suggestion
export type GetAddressesFromPubkey = (pubkey: string) => Promise<string[]>;
```

src/bridge/jsHelpers.ts
```diff
@@ -362,97 +420,104 @@ export const makeScanAccounts =
...
+              const addresses: string[] = await getAddressesFromPubkey(
+                result.publicKey
+              );
 
-              if (finished) {
-                log("debug", `new account scanning process has been finished`);
-                break;
-              }
+              // 2. stepAccount for each address
+              for (let index = 0; index < addresses.length; index++) {
```
gre, line 434: So this only allows to derivate ONE account only? I mean, it only will derivate on `/0'` (account path) but then allows to have multiple address on it? 🤔 

I have a feeling we need to find a way to externalize the logic of "iteration" and the current account scan of `/n'` is one existing and default implementation. because we should avoid the `if` of this proposal imho.

but indeed it may not be that trivial. I would like to try helping on a second iteration on top of this.

haammar-ledger: > So this only allows to derivate ONE account only? I mean, it only will derivate on /0' (account path) but then allows to have multiple address on it?

Yup, Hedera spec exactly... 1 pubkey --> n LL accounts, that's the new use-case that we don't support right now.

> I have a feeling we need to find a way to externalize the logic of "iteration" and the current account scan of /n' is one existing and default implementation. because we should avoid the if of this proposal imho.
> 
> but indeed it may not be that trivial. I would like to try helping on a second iteration on top of this.

I agree I don't like having 2 loop mechanisms in parallel, just depending on that `if`... this is kindof a first working proposal to just meet the new use-case of Hedera and likes.

If we want to externalize the logic of iteration, maybe we can provide the current default one and the new Hedera one as standalone functions of some sort, then families would provide one or the other, and then here in `jsHelpers` we would use a `perFamily` mechanism. Does this match what you had in mind?

src/hw/actions/app.ts
```diff
@@ -634,7 +634,8 @@ export const createAction = (
       ...state,
       inWrongDeviceForAccount:
         state.derivation && appRequest.account
-          ? state.derivation.address !== appRequest.account.freshAddress
+          ? state.derivation.address !== appRequest.account.freshAddress &&
+            state.derivation.address !== appRequest.account.seedIdentifier // Use-case added for Hedera
```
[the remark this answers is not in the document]
juan-cortes: Unless we start introducing many more exceptions like this, I think this is ok

haammar-ledger:
Dark times for CI... a lot of red but only known issues & false negatives.

gre (approved):

gre:
src/families/tezos/synchronisation.test.ts to drop
weird src/hw changes diff

gre (changes requested):
See comment above. Bad merge happened unfortunately

gre:
@haammar-ledger please merge develop in, it means you are outdated to me.

gre:
if the error persists, it reveals a regression to me. this error means Tezos have the "address" style of account id.

gre:
actually as said in my first comment on this PR, you must remove `src/families/tezos/synchronisation.test.ts`

src/hw/actions/app.ts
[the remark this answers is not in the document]
gre: the diff is still bad imho. why does this PR even refer this.

gre (changes requested):
still two merge issues. the .test.ts and the actions/app.ts diff

haammar-ledger:
My bad for the tezos test file.
I don't see an issue with actions/app.ts though? The change in this PR is intentional and I could merge develop without conflict.

src/hw/actions/app.ts
haammar-ledger, line 638: @gre please review and confirm this is safe in your opinion :pray:

gre: afaik this is not safe. at first glance i think this is breaking all implementations as `address` is not `seedIdentifier`. I don't understand why the address != freshAddress is not enough to detect "it's wrong account for device". but what's sure is that seedIdentifier is never address in all our implementations

gre: what's bad is, that means freshAddressPath+freshAddress idea that both are coupled is broken and it might break in many part that expect this to match.

haammar-ledger: Do you mean it can have an impact on existing coins? I doesn't seem possible because as we discussed no current integration have `address from device === seedIdentifier`.
If it can have an impact for Hedera only, then it can be addressed with their own integration, or we can consider a further step.
I manually tested that this part works for Hedera, and Bitcoin as an example of other coin.
I can test other parts but I don't know where it could break.

gre: it might be safe but it will require QA to be sure imo

gre (approved):

PR #1750, opened 2022-02-25
LIVE-1449 Use Ledger's reverse proxy for Crypto.org (CRO)

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1449

## Description / Usage

Crypto.org was released without using a reverse proxy from Ledger, this PR fixes that.

## Expectations


src/env.ts
```diff
@@ -176,7 +176,7 @@ const envDefinitions = {
     desc: "location of the crypto.org indexer testnet API",
   },
   CRYPTO_ORG_RPC_URL: {
-    def: "https://rpc.mainnet.crypto.org",
+    def: "https://cryptoorg-rpc-node.coin.ledger",
```
gre, line 179: .com ?

juan-cortes, line 179:
```suggestion
    def: "https://cryptoorg-rpc-node.coin.ledger.com",
```

juan-cortes (dismissed):

henri-ly (approved):

gre (approved):

PR #1762, opened 2022-02-28
LIVE-1341: Remove xrp lib

henri-ly:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1341

## Description / Usage

Replace Websocket of XRP with JSON-RPC should be more stable cause the websocket was not handle properly.

We should probably retest XRP and all these use case properly

## Expectations

### DIFF
--- a/src/families/ripple/bridge/js.ts
+++ b/src/families/ripple/bridge/js.ts
@@ -33,27 +33,26 @@ import {
 } from "../../../account";
 import getAddress from "../../../hw/getAddress";
 import { open, close } from "../../../hw";
-import {
-  parseAPIValue,
-  parseAPICurrencyObject,
-  formatAPICurrencyXRP,
-} from "../../../api/Ripple";
 import type { CurrencyBridge, AccountBridge } from "../../../types/bridge";
 import signTransaction from "../../../hw/signTransaction";
 import type { Transaction, NetworkInfo } from "../types";
 import { makeAccountBridgeReceive, mergeOps } from "../../../bridge/jsHelpers";
 import {
-  preparePayment,
   submit,
   getAccountInfo,
   getServerInfo,
   getTransactions,
+  parseAPIValue,
 } from "../../../api/Ripple";
+
+const NEW_ACCOUNT_ERROR_MESSAGE = "actNotFound";
 
 // true if the error should be forwarded and is not a "not found" case
 const checkAccountNotFound = (e) => {
   return (
-    !e.data || (e.message !== "actNotFound" && e.data.error !== "actNotFound")
+    !e.data ||
+    (e.message !== NEW_ACCOUNT_ERROR_MESSAGE &&
+      e.data.error !== NEW_ACCOUNT_ERROR_MESSAGE)
   );
 };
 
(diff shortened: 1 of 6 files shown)


henrily-ledger:
Still have some work on CI to do

src/families/ripple/bridge/js.ts
```diff
@@ -297,33 +249,36 @@ const txToOperation =
     }
 
     const op: Operation = {
-      id: `${account.id}-${id}-${type}`,
-      hash: id,
+      id: `${account.id}-${hash}-${type}`,
+      hash: hash,
```
haammar-ledger, line 253: Have you tried this on an account with existing operations? Won't there be duplicates with different ids? :thinking:

src/families/ripple/bridge/js.ts
```diff
@@ -647,7 +608,9 @@ const prepareTransaction = async (a: Account, t: Transaction) => {
   if (!networkInfo) {
     try {
       const info = await getServerInfo(a.endpointConfig);
-      const serverFee = parseAPIValue(info.validatedLedger.baseFeeXRP);
+      const serverFee = parseAPIValue(
+        String(info.info.validated_ledger.base_fee_xrp)
```
haammar-ledger, line 612: `String(null)` will return `"null"`, maybe safer to use `toString`?

src/families/ripple/bridge/js.ts
```diff
@@ -677,7 +640,9 @@ const getTransactionStatus = async (a, t) => {
     feeTooHigh?: Error;
   } = {};
   const r = await getServerInfo(a.endpointConfig);
-  const reserveBaseXRP = parseAPIValue(r.validatedLedger.reserveBaseXRP);
+  const reserveBaseXRP = parseAPIValue(
+    String(r.info.validated_ledger.reserve_base_xrp)
```
haammar-ledger, line 644: (also here)

ghost (commented):
This is largely a rework of the XRP impl, good job!
I'm thinking it will probably need some non-reg in addition to testing the linked issue...

Also maybe that rework is the occasion to split that huge `js.ts` bridge file into appropriate `js-*.ts` files. But not really mandatory. :smile:

src/families/ripple/bridge/js.ts
haammar-ledger: Nevermind I guess id and hash have the same value?

henrily-ledger:
Well I think the issues was related to websocket, since it doesn't open properly so I don't know if I can reproduce it with test right now :/
I'll try to split with js think right now, but will deliver build to QA right now since there's no functional modification by changing this

src/families/ripple/bridge/js.ts
henrily-ledger: yeah, they are the same in the previous implem I just reuse it

gre:
I think we need to set up a bot for it and do our own QA first. 🙏

gre:
@henrily-ledger @haammar-ledger I have changed the base to be `family/ripple` which have a XRP run-per-commit on Mere Denis bot. We can merge to it when you are ready.

PR #1764, opened 2022-03-01
Feature Branch for Ripple (XRP) family

gre:
branch dedicated to XRP implementation that will run a dedicated bot on each commit.

- Rework XRP implementation for LIVE-1341 see https://github.com/LedgerHQ/ledger-live-common/pull/1762


.github/workflows/bot-xrp-mere-denis.yml
```diff
@@ -0,0 +1,79 @@
+name: Bot 'XRP on Mère Denis'
+on:
+  push:
+    branches:
+      - family/ripple
```
haammar-ledger, line 5: Could you please change to `family/ripple-*`?
In accordance with proposal here: https://github.com/LedgerHQ/ledger-live-common/pull/1771

Also while I do see the benefit of the `family` tag to have dedicated bots, it doesn't really fit with the official tags in our proposed gitflow, cf. https://docs.google.com/presentation/d/1tgb2ikDaelddjglTR94Rq3h8-wBExeoMqm9iAWNstg0/edit#slide=id.g10d93822343_1_0
Maybe we should discuss with team so that everyone is on the same page on how to use that `family` tag.

gre: I don't think it's a good idea actually. This will make bot runs in parallel when 2 person work on 2 diff things and will create race conditions.

However, we can do this as soon as we figure out a way to "lock" (fail if running) on a given seed

valpinkman (approved):

ghost (approved):
Awesome rework :muscle: 
A couple comments but nothing blocking I think.

src/families/ripple/js-synchronization.ts
```diff
@@ -0,0 +1,182 @@
...
+  const accountId = encodeAccountId({
+    type: "js",
+    version: "2",
+    currencyId: currency.id,
+    xpubOrAddress: address,
+    derivationMode,
+  });
+  const accountInfo = await getAccountInfo(address);
+
+  if (!accountInfo || accountInfo.error === NEW_ACCOUNT_ERROR_MESSAGE) {
```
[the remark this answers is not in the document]
henrily-ledger: When an account is not created or exist, the api will return this NEW_ACCOUNT_ERROR_MESSAGE and we will not get any info so yeah, in case of scan I guess we should return a shallow account

ghost (approved):

haammar-ledger:
Tests failures are random **unrelated** timeouts, only on Windows and MacOS.
Sorry about that :pray: but doesn't prevent merging.

PR #1765, opened 2022-03-01
fix error revelead by bot

henri-ly:
## Context (issues, jira)

Found a bug thanks to to test about issues that wasn't ordered correctly

## Description / Usage

The problem was when you wanted to send more than what you can spent, it's mean above your existential deposit (which is 0.275 XTZ). So to be sure if it's fix, try to send between your `account.balance - 0.275 XTZ` it should return an error

## Expectations


gre:
I'm going to be merging this into family/tezos as it's evaluating tezos. but i'm waiting the previous bot have finished on it.

PR #1771, opened 2022-03-02
Make coin bots target all <coin>-* branches

ghost:
## Context (issues, jira)

Instead of having bots running on a generic `<coin>` branch, make them run on any branch named `<coin>-*` e.g. `bitcoin-add-cache`, `ripple-rework-websocket`, etc...
This allows a bit more context on different PRs instead of recycling the same name on a same PR that's a bit opaque.

## Expectations

### DIFF
--- a/.github/workflows/bot4.yml
+++ b/.github/workflows/bot4.yml
@@ -2,8 +2,7 @@ name: Bot 'Mooncake'
 on:
   push:
     branches:
-      - bitcoin-js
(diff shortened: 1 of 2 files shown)


gre:
we shouldn't do this until we figure out a way to make it work in parallel and prevent a bot seed to run at same time , using some sort of lock / fail an action if an instance of same seed+family runs.

PR #1777, opened 2022-03-04
Generalize the concept of IterateResult

gre:
proposal to generalize the part inside the loop.

this add an optional attribute `buildIterateResult` on `makeScanAccounts` helper.

There is a default implementation of that "buildIterateResult" that will implement the regular behavior of `iterateResultWithAddressDerivation` (using the device) but this mecanism allow to do whatever you want.

but you can also overrides it

example of "fetching" the results in one pass:

```js
const buildIterateResult: IterateResultBuilder = async ({ result }) => {
  const results = await fetchResults(result.publicKey);
  return async ({ index }) => {
    return results[index];
  };
};
```

overall, this simplifies the implementation of makeScanAccounts.

### DIFF
--- a/src/bridge/jsHelpers.ts
+++ b/src/bridge/jsHelpers.ts
@@ -42,6 +42,7 @@ import type { Result, GetAddressOptions } from "../hw/getAddress/types";
 import { open, close } from "../hw";
 import { withDevice } from "../hw/deviceAccess";
 
+// Customize the way to iterate on the keychain derivation
 type IterateResult = ({
   transport: Transport,
   index: number,
--- a/src/bridge/jsHelpers.ts
+++ b/src/bridge/jsHelpers.ts
@@ -52,7 +52,7 @@ type IterateResult = ({
   currency: CryptoCurrency,
 }) => Promise<Result | null>;
 
-type IterateResultBuilder = ({
+export type IterateResultBuilder = ({
   result: Result, // derivation on the "root" of the derivation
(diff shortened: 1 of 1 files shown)


src/bridge/jsHelpers.ts
```diff
@@ -42,7 +42,21 @@ import type { Result, GetAddressOptions } from "../hw/getAddress/types";
...
-export type GetAddressesFromPubkey = (pubkey: string) => Promise<string[]>;
+// Customize the way to iterate on the keychain derivation
+type IterateResult = ({
+  transport: Transport,
+  index: number,
+  derivationsCache: Object,
+  derivationScheme: string,
+  derivationMode: DerivationMode,
+  currency: CryptoCurrency,
+}) => Promise<Result>;
```
haammar-ledger, line 53:
```suggestion
}) => Promise<Result | null>;
```
Hedera needs to stop looping when no address is found, otherwise `getAccountShape` gets called with a `null` address.

src/bridge/jsHelpers.ts
```diff
@@ -42,7 +42,21 @@ import type { Result, GetAddressOptions } from "../hw/getAddress/types";
...
+
+type IterateResultBuilder = ({
```
haammar-ledger, line 55:
```suggestion
export type IterateResultBuilder = ({
```
This type needs to be visible for other coins to customize it.

ghost (changes requested):

PR #1780, opened 2022-03-04
Update Ethereum related libraries

gre:
we must try to merge this in priority so other dependencies to update will benefit the update of ledgerjs (tokens update notably).

This PR only update the Ethereum related libraires

- ethereumjs-tx
- walletconnect

a new bot spec runs on this branch to evaluate the work (note that develop have bot issues on the operation optimistic so this won't be fixed here)

I have also updated **axios** because it was updated in ledgerjs, and it's good to dedup libs in general.

This is necessary to also ship:
- LL-8941 Conditionally release usb connection hold after timeout (**loosely** dependent tho, but it's better this gets in first, so all ledgerjs lib are in "sync")
- LL-9254 bi-weekly CAL update from form


ghost (approved):
Code LGTM but needs QA approval.

PR #1785, opened 2022-03-08
Live 719/new live apps providers

IAmMorrow:
## Context (issues, jira)

Added three new data providers to be used for live apps:
- Live App Provider: Provide data about live apps and their manifests
- Discover Provider: Provide data about the discover section (catalog)
- Ramp Provider: Provide data about the multibuy catalog

## Description / Usage

This is non-breaking because the old all-in-one provider is still available. Clients are expected to migrate to this new system with the multibuy and discover v2 implementations.

## Expectations


gre (approved):

haammar-ledger:
Considered top prio and merged on behalf of @IAmMorrow.

PR #1792, opened 2022-03-09
blake2b on dependency

henri-ly:
## Context (issues, jira)

https://ledger.slack.com/archives/C02LACPH3U3/p1646834398860669

## Description / Usage

missing dependency shouldn't be major because blake it's already on .lock but it's still better to have it on main dependency

## Expectations


ghost (approved):

juan-cortes (approved):

gre (changes requested):
ci shows a problem, are you sure there are no diff when running yarn ?

PR #1800, opened 2022-03-10
Cosmos JS

ghost:
## Expectations


src/families/cosmos/api/Cosmos.ts
```diff
@@ -281,6 +282,15 @@ export const broadcast = async ({
     );
   }
 
+  log(
+    "info",
+    "debug operation: ",
+    patchOperationWithHash(
+      { ...operation, blockHeight: data.tx_response.height },
+      data.tx_response.txhash
+    )
```
gre, line 291: the type `operation` is possibly not serializable, typically it have "BigNumber" types, so it would be better to do `toOperationRaw(...` of this which will transform it to something json serializable.

you may also factorize an object as you do a `return` after it.

PR #1805, opened 2022-03-12
[LIVE-1174] - Feature: Upgrade NFT Architecture

lambertkevin:
# ⚠️ WARNING ⚠️
# LLC CLI DEPS HAVE BEEN CHANGED IN ORDER TO PASS THE CI, THIS WILL NEED TO BE REVERTED BEFORE MERGING THIS PR.

## Context (issues, jira)

In order to allow for "synchronous" metadata and cross-chain NFT resolution, we updated the NFT model to integrate it in the CurrencyBridge as an optional method.

## Description / Usage

Screenshot 2022-03-25 at 11 01 59


gre:
Can you also work on the mocks so we can deliver this rework with mocks that can be tested? see with @aulric-ledger too

src/nft/helpers.ts
```diff
@@ -49,30 +54,99 @@ export const nftsFromOperations = (ops: Operation[]): NFT[] => {
...
 
 export const nftsByCollections = (
-  nfts: NFT[] = [],
+  nfts: Array<ProtoNFT | NFT> = [],
   collectionAddress?: string
-): CollectionWithNFT[] => {
-  const filteredNfts = collectionAddress
-    ? nfts.filter((n) => n.collection.contract === collectionAddress)
-    : nfts;
+): Record<string, Array<ProtoNFT | NFT>> | Array<ProtoNFT | NFT> => {
```
haammar-ledger, line 59: This return type is a bit hard to read, maybe add a small comment on this function to explain what it does.

src/nft/helpers.ts
```diff
@@ -49,30 +54,99 @@ export const nftsFromOperations = (ops: Operation[]): NFT[] => {
...
+  tokenId: string,
+  currencyId: string
+): string => {
+  return `${currencyId}-${contract}-${tokenId}`;
 };
 
-export const getNftKey = (contract: string, tokenId: string): string => {
-  return `${contract}-${tokenId}`;
+const batchersMap = new Map();
+const makeBatcher = (api: API, chainId: number) =>
```
haammar-ledger, line 83: Same here, could use a one-liner to explain what the function does (esp. since it doesn't have an explicit return type).
(But the content itself is very well commented ! :+1: )

src/nft/helpers.ts
```diff
@@ -49,30 +54,99 @@ export const nftsFromOperations = (ops: Operation[]): NFT[] => {
...
+      }): Promise<NFTMetadataResponse> {
+        return new Promise((resolve, reject) => {
+          batch.push({ couple: { contract, tokenId }, resolve, reject });
+          timeoutBatchCall();
+        });
+      },
+    };
+  })();
+
+export const metadataCallBatcher = (currency: CryptoCurrency) => {
```
haammar-ledger, line 139: Same here, esp. since it's an exported function + no return type.

ghost (commented):
CI tests didn't run it seems :thinking: maybe need to restart them manually.
Mostly nitpicking otherwise :stuck_out_tongue:

haammar-ledger:
> Can you also work on the mocks so we can deliver this rework with mocks that can be tested? see with @aulric-ledger too

@lambertkevin also this earlier question from Gaetan.

lambertkevin:
> > Can you also work on the mocks so we can deliver this rework with mocks that can be tested? see with @aulric-ledger too
> 
> @lambertkevin also this earlier question from Gaetan.

@haammar-ledger Yes I had a meeting with @aulric-ledger, and it turns out the issues that the QA have with the NFT mocks are not really related to the new model... They don't really care about the image being legit or not, nor even coming from the actual nfts of the account or not. 

Basically we have 2 scenarii:
- Metadata for an NFT you own: For that case we have the new model that *could* provide some sync metadata and allow for an image or strings to be displayed in an E2E test. This is usable for some views (Your gallery, your collection or the NFT viewer itself)
- Metadata for an NFT you *don't* own anymore: For that case we will always rely on a backend to get the metadata since we can't have the information anywhere else without storing even more stuff inside our `Operation` model (which would mean having even more storage issues). For these views (basically any Operation related view), the new model is not going to provide anything new here and we'll have the same problems for the E2E tests. 

So since we only have "half" a solution here with the new model, my idea would be to simply override the metadata provided by LLC in the `useNftMetadata` react hook by something non flaky that would make the UI testable. 
And, to me, that can definitely be done in another PR to avoid delaying this one 🤷‍♂️

cli/package.json
```diff
@@ -39,7 +39,7 @@
     "@ledgerhq/hw-transport-node-hid": "6.24.1",
     "@ledgerhq/hw-transport-node-speculos": "6.24.1",
     "@ledgerhq/ledger-core": "6.14.5",
-    "@ledgerhq/live-common": "^21.34.0",
+    "@ledgerhq/live-common": "https://github.com/LedgerHQ/ledger-live-common.git#31177aae3a559e17f8452ed8dc6e010cd1d6f41e",
```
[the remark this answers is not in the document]
lambertkevin: Added a gigantic message in the PR description to avoid forgetting it 👌

cli/src/commands/sync.ts
```diff
@@ -1,6 +1,11 @@
 import { map, switchMap } from "rxjs/operators";
-import { accountFormatters } from "@ledgerhq/live-common/lib/account";
+import {
+  accountFormatters,
+  decodeAccountId,
+} from "@ledgerhq/live-common/lib/account";
 import { metadataCallBatcher } from "@ledgerhq/live-common/lib/nft";
```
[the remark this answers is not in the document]
lambertkevin: Done 👌

src/account/formatters.ts
```diff
@@ -138,32 +138,34 @@ const cliFormat = (account, level?: string) => {
...
 
-    str += NFTCollections.map(
-      // nfts are set to any because there not just NFT, we added a metadata prop on the fly
-      // in the first step of the Rxjs flow to avoid having some async code here
-      ({ contract, nfts }: { contract: string; nfts: any[] }) => {
-        const tokenName = nfts?.[0]?.metadata?.tokenName;
-        const { bold, magenta, cyan, reverse } = styling;
+    str += Object.entries(NFTCollections)
+      .map(
+        // nfts are set to any because there not just NFT, we added a metadata prop on the fly
```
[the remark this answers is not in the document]
lambertkevin: Done 👌

src/nft/helpers.ts
lambertkevin: Done 👌

src/nft/helpers.ts
lambertkevin: Done 👌

src/families/ethereum/nftMetadataResolver.ts
```diff
@@ -0,0 +1,44 @@
...
+}) => {
+  // This is for test/mock purposes
+  if (typeof metadata !== "undefined") {
+    return {
+      status: 200,
+      result: {
+        contract,
+        tokenId,
+        ...metadata,
+      },
```
haammar-ledger, line 25: Good catch, maybe explicitly set the function's return type to `Promise<NFTMetadataResponse>`, cause I think TS would have seen that.

lambertkevin: Yeah I don't know why it was catched only now by my VSCode... TS should have told me right away but it wasn't the case when I added the "sync" part of the resolver at the time... 🤔 
And I can't actually be more precise than `CurrencyBridge["nftMetadataResolver"]` here cause it already contains the response part...

haammar-ledger: Yeah idk why but happened to me also that the function's _type_ isn't enough for TS to deduce the _return type_... no idea why, and sometimes it even works.

However I think you could add this to the function's signature, right?
```diff
-  }) => {
+  }): Promise<NFTMetadataResponse> => {
```

haammar-ledger:
(Red CI because there's a known issue with Cosmos backend at the moment...)

ghost (approved):

PR #1812, opened 2022-03-14
Fixes LIVE-1123

gre:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1123

I tested it locally with a 

```js
ledger-live getTransactionStatus --id myaccount --recipient anemptyaccount --amount 0.02
```

but unfortunately it's not yet covered by automation. we'll need to see how we'll write detected tests for error cases in future.


src/families/tezos/bridge/js.ts
```diff
@@ -43,9 +43,7 @@ import api from "../api/tzkt";
 
 const receive = makeAccountBridgeReceive();
 
-const EXISTENTIAL_DEPOSIT = new BigNumber(
-  DEFAULT_STORAGE_LIMIT.ORIGINATION * 1000
-);
+const EXISTENTIAL_DEPOSIT = new BigNumber(275000);
```
gre, line 46: this was a mistake I did, ORIGINATION is not 275*

src/families/tezos/bridge/js.ts
```diff
@@ -159,6 +147,19 @@ const getTransactionStatus = async (
     errors.amount = new NotEnoughBalance();
   }
 
+  // Catch a specific case that requires a minimum amount
+  if (
+    !errors.amount &&
+    t.mode === "send" &&
+    t.amount.lt(EXISTENTIAL_DEPOSIT) &&
+    (await api.getAccountByAddress(t.recipient)).type === "empty"
+  ) {
```
gre, line 156: I had to move it here because it's actually not inside a "tequito error" (weirdly) but one more edge case to add on top of taquito

src/families/tezos/bridge/js.ts
```diff
@@ -159,6 +147,19 @@ const getTransactionStatus = async (
...
+    amountMustReset = true;
+    errors.amount = new NotEnoughBalanceBecauseDestinationNotCreated("", {
+      minimalAmount: "0.275 XTZ",
```
haammar-ledger, line 159: I think there's an util to get the value from EXISTENTIAL_DEPOSIT + the currency's unit, right? But I can't find it...
Could be a small improvement instead of hardcoding it.

ghost (approved):

src/families/tezos/bridge/js.ts
henrily-ledger: I thought it was the same, didn't paid attention also

henri-ly (approved):

src/families/tezos/bridge/js.ts
gre: definitely. and one other issue is that it's hardcoded but may change based on user's language.

PR #1815, opened 2022-03-15
added dummy app to remoteLiveAppProvider

ggilchrist-ledger:
## Description / Usage

Adding the Dummy Live app to the `RemoteLiveAppProvider` so it is available for the `discover.spec.ts` test.


IAmMorrow (approved):
LFG 🚀

PR #1820, opened 2022-03-20
(MarketProvider): updated for better ux and seperation of data sections single vs list market data

LFBarreto:
## Context (issues, jira)

required for V3 mobile as it fixes most remaining UX issues due to poor data context isolation between market data lists vs single coin data points.

this is breaking :warning:  i'll create a PR for desktop as well to use the updated API

## Expectations


gre:
I set a HODL to explicily block the merge until we are really ready to organize the merge, because live-common historically rarely do a break, and when it does, it's a new major bump :) if we still follow semver 🙏

PR #1831, opened 2022-03-24
Remove all libcore codebase

gre:
**This fully remove the "libcore" codebase on Ledger Live stack.**

## feature impact for QA

at best we want these taken out of this PR into separate PRs so we have a more "non impacting" PR here.

- Tezos `validateAddress` was still using libcore's, even in the JS implementation. It has been changed to use `taquito/utils`. This is the validation of the Recipient field in the Send flow.
- Cosmos `calculateFees` was still using libcore's too. This impacts **canClaimRewards** which is uses on LLD & LLM to determine if the Claim Reward is possible.
- LLD & LLM logic of "Reset" and "Clear Cache" is impacted as we removed libcore logic in these.

I believe there are no other impact for the users, because apart from these points, we already moved away from libcore and this is mostly code deletion.

## impact for userland (users of live-common)

- **all import using libcore on user end have to be removed**. This includes notably the usage of `withLibcore` or `implementLibcore`. User end can also completely remove dependency to lib-ledger-core but it's independent of this.
- The `libcore:*` account ids will still "work" (be accepted by the implementation) as we will still have this implicit and automatic migration by using the JS bridge.

## impact for live-common contributors

The "dataset tests" that runs from `all.libcore.test.ts` has been modularized. If you are working in a new coin integration, you will now need to create a `bridge.test.ts` in the family that do this (replace bitcoin by your family):

```js
import { setup } from "../../__tests__/test-helpers/libcore-setup";
import { testBridge } from "../../__tests__/test-helpers/bridge";
import dataset from "./test-dataset";

setup("bitcoin");
testBridge("bitcoin", dataset);
```

in `types.ts` you can also remove these:

```diff
- export type CoreStatics = Record<string, never>;
- export type CoreAccountSpecifics = Record<string, never>;
- export type CoreOperationSpecifics = Record<string, never>;
- export type CoreCurrencySpecifics = Record<string, never>;
- export const reflect = (_declare: any) => {};
```


ghost (approved):

haammar-ledger:
QA approved by @aulric-ledger and @mcoursier-ledger

haammar-ledger:
CI failures are only timeouts... known and unrelated issue.

PR #1835, opened 2022-03-27
LL-1589

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LIVE-1589

Description:
when there is an error was raised during the btc sync, we need to release the axios client instance firstly instead of throwing the exception directly. This cause the "retry sync" doesn't work


ghost (approved):
:clap:

haammar-ledger:
Tests failing because of an ongoing situation with Cosmos backend... unrelated to this PR.

PR #1844, opened 2022-04-01
Solana staking

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1751
https://ledgerhq.atlassian.net/browse/LIVE-1757

## Description / Usage

Solana staking integration, developed by 3rd party.
I approved their initial PR :heavy_check_mark: 

## Expectations


src/families/solana/validator-app/index.ts
```diff
@@ -0,0 +1,105 @@
...
+  voteAccount: string;
+  name?: string;
+  avatarUrl?: string;
+  wwwUrl?: string;
+};
+
+const URLS = {
+  validatorList: (cluster: Extract<Cluster, "mainnet-beta" | "testnet">) => {
+    const clusterSlug = cluster === "mainnet-beta" ? "mainnet" : cluster;
+    return `https://www.validators.app/api/v1/validators/${clusterSlug}.json?order=score`;
```
haammar-ledger, line 58: @konoart At the same time you change the value to `http://validators-solana.coin.ledger.com`, it would be useful for us to get that server's URL from an env var, similar to `API_SOLANA_PROXY` for the node. So we could fallback to the direct endpoint for testing, especially while the new endpoint finishes being set up. Thanks. :slightly_smiling_face:

lambertkevin (approved):
LGTM

PR #1855, opened 2022-04-05
bugfix/LIVE-1560 : algo add min balance to transaction test

henri-ly:
## Context (issues, jira)

Bugfix

## Description / Usage

https://ledgerhq.atlassian.net/browse/LIVE-1560

## Expectations

### DIFF
--- a/src/families/bitcoin/logic.ts
+++ b/src/families/bitcoin/logic.ts
@@ -1,5 +1,4 @@
 import cashaddr from "cashaddrjs";
-import bchaddr from "bchaddrjs";
 import { Currency, isValidAddress } from "./wallet-btc";
 import { RecipientRequired, InvalidAddress } from "@ledgerhq/errors";
 import type { Account, CryptoCurrency, CryptoCurrencyIds } from "./../../types";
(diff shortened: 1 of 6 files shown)


src/families/bitcoin/logic.ts
```diff
@@ -143,8 +142,8 @@ type CoinLogic = {
   injectGetAddressParams?: (arg0: Account) => any;
 };
 
-const bchToCashaddrAddressWithoutPrefix = (recipient) =>
-  recipient ? bchaddr.toCashAddress(recipient).split(":")[1] : recipient;
+export const bchToCashaddrAddressWithoutPrefix = (recipient) =>
+  recipient ? recipient.substring(recipient.indexOf(":") + 1) : recipient;
```
gre, line 146: these are not equivalent. toCashAddress can sometimes also convert legacy addr.

```
toCashAddress('1B9UNtBfkkpgt8kVbwLN9ktE62QKnMbDzR') // bitcoincash:qph5kuz78czq00e3t85ugpgd7xmer5kr7c5f6jdpwk
```

and as we use bchToCashaddrAddressWithoutPrefix maybe for diff reasons, we will have to double check its impact. hopefully covered by test snapshot?

src/families/bitcoin/specs.ts
```diff
@@ -257,7 +257,9 @@ const bitcoinLikeMutations = ({
     test: ({ account }) => {
       expect(
         account.bitcoinResources?.utxos
-          .filter((u) => u.blockHeight && u.blockHeight < account.blockHeight) // Exclude pending UTXOs and the Utxos just written into new block
+          .filter(
+            (u) => u.blockHeight && u.blockHeight < account.blockHeight - 10
+          ) // Exclude pending UTXOs and the Utxos just written into new block (10 blocks time)
```
gre, line 262: - why 10. the comment should refer to what was the problem solved otherwise someone coming later to this fix will not understand anything.
- make this a env please

PR #1856, opened 2022-04-05
Replace all deprecated uses of open(deviceId) with withDevice

ghost:
## Context (issues, jira)

It's all in the title.
Required for Solana staking https://github.com/LedgerHQ/ledger-live-common/pull/1844 as it fixes this blocking issue: https://ledgerhq.atlassian.net/browse/LIVE-1890

## Description / Usage

No change perceptible from the user, the device is just used how it's meant to be.

## Expectations


juan-cortes (approved):
lgtm

PR #1862, opened 2022-04-06
[LIVE-1909] - Bugfix: OpenSea Lazy mint quantity fix

lambertkevin:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1909

## Description / Usage

This should fix the OpenSea issue with the lazy minting giving negative balance of NFTs making them invisible in the live even after receiving them again.

## Expectations


ghost (approved):
Super well documented, thanks ;)

PR #1863, opened 2022-04-06
[LIVE-1911][LIVE-1912] Feature - Add collection name resolver to eth familly

lambertkevin:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LIVE-1911
https://ledgerhq.atlassian.net/browse/LIVE-1912

## Description / Usage

Screenshot 2022-04-06 at 23 09 07

Changing the bridge architecture for NFTs to allow for a new call for collection names.
Adding the `useNftCollectionMetadata` hook.

## Expectations


src/nft/NftMetadataProvider/types.ts
```diff
@@ -42,33 +47,37 @@ export type NFTMetadataContextAPI = {
...
+        }
+  ) => Promise<NFTMetadataResponse | NFTCollectionMetadataResponse>;
 };
 
 export type BatchElement = {
-  couple: {
-    contract: string;
-    tokenId: string;
-  };
+  element: any;
```
haammar-ledger, line 74: This could be typed, in an another iteration later.

src/types/nft.ts
```diff
@@ -47,3 +51,11 @@ export type NFTMetadataResponse = {
     links: Record<NFTMetadataLinksProviders, string>;
   } | null;
 };
+
+export type NFTCollectionMetadataResponse = {
+  status: 200 | 404 | 500;
```
haammar-ledger, line 56: Just wondering, the status can never be something else?...
(same for NFTMetadataResponse above)

ghost (approved):

src/types/nft.ts
lambertkevin: I don't think so no, and it's not in the Swagger so I'll ask Rémi for more precision 👌

ghost (approved):

PR #1876, opened 2022-04-11
[LL-8775] added platform ERC20 token logic 

IAmMorrow:
## Context (issues, jira)

This PR adds ERC20 tokens support to the Platform API

## Expectations


src/platform/converters.ts
```diff
@@ -1,13 +1,39 @@
...
+} from "../types";
 import type {
   PlatformAccount,
   PlatformCurrency,
   PlatformTransaction,
 } from "./types";
 
-export function accountToPlatformAccount(account: Account): PlatformAccount {
+export function accountToPlatformAccount(
+  account: Account | TokenAccount,
```
JunichiSugiura, line 17: What about the case with `ChildAccount`? I see `AccountLike[]` is passed to this function on desktop side which includes child account type. I believe `ChildAccount` doesn't have `freshAddress` field so I guess `parentAccount.freshAddress` should be passed instead? 🤔

JunichiSugiura:
moved to https://github.com/LedgerHQ/ledger-live/pull/81

PR #1886, opened 2022-04-12
Fix bch bot and change hash into txid from explorer response

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LIVE-1931
1. we read the "hash" field from explorer for each transaction, now we use id field instead.
2. fix bch bot


gre (approved):
ideally we should split the two PRs (fix the bot and hash change) because the bot fix can be merged as is VS the other changes are less safe to merge 🙏

PR #1887, opened 2022-04-12
LIVE-1004 Hedera integration

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1004

## Description / Usage

Hedera integration in Ledger Live - add accounts / sync / send / receive

## Expectations


src/families/bitcoin/__snapshots__/bridge.test.ts.snap
```diff
@@ -19844,6 +19844,9 @@ Array [
       "type": "OUT",
       "value": "210486",
     },
+  ],
+  Array [],
```
henrily-ledger, line 19848: Why is there an empty Array ?

henri-ly: Something is weird it shouldn't modifiy bitcoin snapshot

src/families/hedera/test-dataset.ts
```diff
@@ -0,0 +1,92 @@
+import type { DatasetTest } from "../../types";
+import type { Transaction } from "./types";
+import {
+  NotEnoughBalance,
+  InvalidAddressBecauseDestinationIsAlsoSource,
+  AmountRequired,
+} from "@ledgerhq/errors";
+import { fromTransactionRaw } from "./transaction";
+
+const dataset: DatasetTest<Transaction> = {
```
henri-ly: We will probably need to find a frozen seed

henri-ly:
LGTM !

src/families/hedera/test-dataset.ts
ghost: Good point, we'll ask Hedera about it.

PR #1897, opened 2022-04-15
Cosmos JS

alexalouit:
## Context (issues, jira)
Ledger-live-common <> Cosmos:
- Move from CosmJS/ledger-amino to LedgerHQ/hw-app-cosmos signer

## Expectations


desirendr:
Hey @alexalouit, can you add some context into the PR and maybe screenshots so we know what it's about ? Thanks a lot 🙏

ghost (commented):
Since @cosmjs is still used in the end, could you consider updating it in package.json?
Thanks.

src/families/cosmos/js-signOperation.ts
```diff
@@ -32,52 +31,40 @@ const signOperation = ({
...
-        });
-
         o.next({ type: "device-signature-requested" });
 
-        const accounts = await ledgerSigner.getAccounts();
-
-        let pubkey;
+        const { publicKey } = await hwApp.getAddress(
+          // is valid?
+          account.freshAddresses[0].derivationPath,
```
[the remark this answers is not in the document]
alexalouit: fix in 828fec72b

alexalouit:
> Since @cosmjs is still used in the end, could you consider updating it in package.json? Thanks.

I need time to understand how AminoTypes works, there's a breaking change.

haammar-ledger:
> > Since @cosmjs is still used in the end, could you consider updating it in package.json? Thanks.
> 
> I need time to understand how AminoTypes works, there's a breaking change.

If it's too much trouble it's ok we can skip, as this will be removed later when migrating to protobuf signing right?

src/families/cosmos/js-signOperation.ts
```diff
@@ -32,52 +31,39 @@ const signOperation = ({
...
         o.next({ type: "device-signature-requested" });
 
-        const accounts = await ledgerSigner.getAccounts();
-
-        let pubkey;
+        const { publicKey } = await hwApp.getAddress(
+          account.freshAddressPath,
+          "cosmos",
+          false
+        );
```
haammar-ledger, line 48: Sorry I didn't think of this before, but can't the public key be deduced from `account.freshAddress`?
Same as done [here](https://github.com/LedgerHQ/ledger-live-common/blob/7ff03d79cd5c812a840a15c55ab8e97a3a2c7faf/src/families/cosmos/js-synchronisation.ts#L167).
This would save time without the extra call to the Nano app, if we can deduce the info.

alexalouit: I don't think it is available without the full signature package, decodeBech32Pubkey is exclusively for a bech32 format.
By the way, it is the reverse operation that is done [here](https://github.com/LedgerHQ/ledger-live-common/blob/7ff03d79cd5c812a840a15c55ab8e97a3a2c7faf/src/families/cosmos/js-synchronisation.ts#L167) (pubkey to address) which is much simpler.
Not using the hwApp will add CosmJS dependency cost, isn't this contradictory with protobuf switching?

alexalouit:
> > > Since @cosmjs is still used in the end, could you consider updating it in package.json? Thanks.
> > 
> > 
> > I need time to understand how AminoTypes works, there's a breaking change.
> 
> If it's too much trouble it's ok we can skip, as this will be removed later when migrating to protobuf signing right?

Yes, that is true.

PR #1908, opened 2022-04-25
fixing an odd crash caused by the getProviders function

IAmMorrow:
## Context (issues, jira)

[LIVE-2126](https://ledgerhq.atlassian.net/browse/LIVE-2126)

## Description / Usage

Fixed an odd bug where the Error() what returned instead of being throw.

## Expectations


gre (approved):

gre (changes requested):

gre (approved):
we'll need to increase our test coverage and provide a wider solution to "mock" the networking in future. ok to merge in the fix for now.

PR #1914, opened 2022-04-26
Support/add ci for explorer staging

henri-ly:
## Context (issues, jira)

Support for CI

## Description / Usage

We add the ability to run multiple coins in the bot 

## Expectations

### DIFF
--- a/.github/workflows/bot-staging-explorer-btc.yml
+++ b/.github/workflows/bot-staging-explorer-btc.yml
@@ -1,0 +1,90 @@
+name: Bot Staging Explorer 'Bitcoin'
+on:
+  schedule:
+  - cron: 0 8,12,16 * * 1-5
+  
+  workflow_dispatch:
+    inputs:
+      ref:
+        description: "Which ref (branch or tag) to pull from ?"
+        required: false
+        default: master
+
+jobs:
+  start-runner:
+    name: "start ec2 instance (Linux)"
+    if: ${{ always() }}
+    uses: ledgerhq/actions/.github/workflows/start-linux-runner.yml@main
+    secrets:
+      CI_BOT_TOKEN: ${{ secrets.CI_BOT_TOKEN }}
+
+  stop-runner:
+    name: "stop ec2 instance (Linux)"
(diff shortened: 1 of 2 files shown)


src/bot/index.ts
```diff
@@ -58,11 +58,11 @@ export async function bot({ currency, family, mutation }: Arg = {}) {
   const maybeCurrency = currency
     ? findCryptoCurrencyByKeyword(currency)
     : undefined;
-  const maybeFilterOnlyFamily = family;
+  const maybeFilterOnlyFamily = family ? family.split(",") : undefined;
 
-  for (const family in allSpecs) {
-    const familySpecs = allSpecs[family];
-    if (maybeFilterOnlyFamily && maybeFilterOnlyFamily !== family) {
+  for (const fam in allSpecs) {
```
haammar-ledger, line 63: More `spec` than `fam` ideally ^^

src/bot/index.ts
```diff
@@ -58,11 +58,11 @@ export async function bot({ currency, family, mutation }: Arg = {}) {
...
+    const familySpecs = allSpecs[fam];
+    if (maybeFilterOnlyFamily && !maybeFilterOnlyFamily.includes(fam)) {
```
haammar-ledger, line 65: From L61, it seems `maybeFilterOnlyFamily` can be `undefined` right?

ghost (changes requested):

src/bot/index.ts
henrily-ledger: I think it shouldn't be a problem we got if (maybeFilterOnlyFamily) covering the case it's undefined or maybe i misunderstood the point ?

src/bot/index.ts
henrily-ledger: it's family because allSpecs is something like etc.... 
```
{ algorand, bitcoin, ethereum, cosmos }
```

But that shouldn't matter a lot because fam is just use in these two lines (and family was use before as an arg)

ghost (approved):

PR #1918, opened 2022-04-28
Fixes LIVE-2166: nftsFromOperations was mutating account.operations

gre:
ops.sort()  is mutating ops, so we must protect this to happen.

It was the cause of a bad reordering operations issue on Ethereum.

A test was brought to also ensure we don't regress.


valpinkman (approved):
with the test ❤️

lambertkevin (approved):

juan-cortes (commented):
:)

PR #1922, opened 2022-04-28
fix: hedera spec bot

jackthta:


src/families/hedera/utils.ts
```diff
@@ -11,23 +17,28 @@ export async function calculateAmount({
...
-    transaction.useAllAmount == true
-      ? await estimateMaxSpendable({ account })
-      : transaction.amount;
+    if (txAmount.isGreaterThan(account.balance)) {
+      // NOTE: work-around for 'NotEnoughBalance' edge case
+      // since the addition of estimated fees overflows account balance,
+      // set amount w/o the added estimated fees.
+      amount = transaction.amount;
+    } else {
+      amount = transaction.amount.plus(estimatedFees);
```
haammar-ledger, line 35: This looks really weird, like you're ignoring the fees just because you can't pay them. :thinking: seems it could lead to other issues e.g. signing an amount different from what the user set for the transaction.

Let's merge and see what the bot has to say about it. :slightly_smiling_face:

PR #1926, opened 2022-05-02
LIVE-2160 Bugfix/bitcoin js

hzheng-ledger:
https://ledgerhq.atlassian.net/browse/LIVE-2160


src/families/bitcoin/wallet-btc/utils.ts
```diff
@@ -7,6 +7,7 @@ import { Currency, ICrypto } from "./crypto/types";
 import cryptoFactory from "./crypto/factory";
 import { fallbackValidateAddress } from "./crypto/base";
 import { UnsupportedDerivation } from "../../../errors";
+import varuint from "varuint-bitcoin";
```
haammar-ledger, line 10: `varuint-bitcoin` isn't added in _package.json_ can you please add it + update the _dependencies.md_ file?

src/families/bitcoin/getAccountNetworkInfo.ts
```diff
@@ -47,10 +47,12 @@ export async function getAccountNetworkInfo(
   }
   // Fix fees if suggested fee is too low, this is only for viacoin/decred because the fees backend endpoint is broken
   if (
-    (account.currency.id === "viacoin" || account.currency.id === "decred") &&
+    (account.currency.id === "viacoin" ||
+      account.currency.id === "decred" ||
+      account.currency.id === "qtum") &&
```
haammar-ledger, line 52: Not for this PR, but in the future we could consider changing this kind of hardcoded test to use the coin-specific mechanisms that we have here: https://github.com/LedgerHQ/ledger-live-common/blob/develop/src/families/bitcoin/logic.ts#L138

ghost (changes requested):

src/families/bitcoin/wallet-btc/utils.ts
hzheng-ledger: fixed

ghost (approved):

PR #1941, opened 2022-05-05
LIVE-2241: take care about currency of the fees

alexalouit:
## Context (issues, jira)

Better fees and balances computation when the currency is not ATOM, especially during IBC transfer present on the account.

## Expectations


ghost (approved):
:+1: !

PR #1943, opened 2022-05-06
LIVE-2241 Take care about currency for balance computation (#1941)

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-2241

## Description / Usage

Cosmos accounts containing assets other than ATOM had an incorrect ATOM balance that was including the balance of all assets.
This PR makes sure only ATOM is used for the balance computation.

### Before
```
Cosmos legacy cosmos1c...na2l68hq: 111.45 ATOM (98ops) (cosmos1c8qxw4dsj37dy0s2dmsc34ahvkredgna2l68hq on 44'/118'/0'/0/0) #0 js:2:cosmos:cosmos1c8qxw4dsj37dy0s2dmsc34ahvkredgna2l68hq: (! sum of ops 646.386381 ATOM) 0.449751 ATOM spendable. 111.002554 ATOM delegated. 
```

### After
```
Cosmos legacy cosmos1c...na2l68hq: 111.35 ATOM (98ops) (cosmos1c8qxw4dsj37dy0s2dmsc34ahvkredgna2l68hq on 44'/118'/0'/0/0) #0 js:2:cosmos:cosmos1c8qxw4dsj37dy0s2dmsc34ahvkredgna2l68hq: (! sum of ops 646.386381 ATOM) 0.349751 ATOM spendable. 111.002554 ATOM delegated. 
```

## Expectations

### DIFF
--- a/src/families/cosmos/api/Cosmos.ts
+++ b/src/families/cosmos/api/Cosmos.ts
@@ -1,14 +1,17 @@
 import { getEnv } from "../../../env";
 import BigNumber from "bignumber.js";
 import network from "../../../network";
-import { Operation } from "../../../types";
+import { CryptoCurrency, Operation } from "../../../types";
 import { patchOperationWithHash } from "../../../operation";
 
 const defaultEndpoint = getEnv(
   "API_COSMOS_BLOCKCHAIN_EXPLORER_API_ENDPOINT"
 ).replace(/\/$/, "");
 
-export const getAccountInfo = async (address: string): Promise<any> => {
+export const getAccountInfo = async (
+  address: string,
+  currency: CryptoCurrency
+): Promise<any> => {
   try {
     const [
       { accountNumber, sequence },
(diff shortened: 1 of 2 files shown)


juan-cortes (approved):
Would be ideal to add a test for this into the datasets. I'm not sure how difficult it would be to regenerate them and clean up the [libcore references](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/cosmos/test-dataset.ts#L45).

lambertkevin (approved):
LGTM

haammar-ledger:
> Would be ideal to add a test for this into the datasets. I'm not sure how difficult it would be to regenerate them and clean up the [libcore references](https://github.com/LedgerHQ/ledger-live-common/blob/master/src/families/cosmos/test-dataset.ts#L45).

You're definitely right about adding an account in datasets, we just don't have an appropriate one at the moment unfortunately... I've asked for one to our QA but they may not have bandwidth now, so I'll merge this now, but whenever an account is provided we'll have a task to add it to the datasets.

About the libcore accounts, I think these should stay as they actually allow to test migration for users with old builds.

PR #1944, opened 2022-05-09
Simplify a lot "tools" and remove its dep to live-common

gre:
This is a drastic simplification of "tools" project.

Sometimes it's good to do some clean up on legacy stuff. This will be a benefit in light of incoming monorepo, but we also have a real production issue here as Vercel build is broken, it aims to restore it.

 Removing a lot of unused old pages we had, **only keeping the essential of "logsviewer" and "lld-signatures".**
remove the dependency of live-common in logsviewer which have two benefits:
- logsviewer: fixes some cases where a index.html was downloaded instead of the app.json generation! – https://ledgerhq.atlassian.net/browse/TECSUPPORT-1064
- fixes the vercel build to work. Polkadot, Celo, Solana,... these implementations have the tendency to badly break in this webpack project.. – fixes https://ledgerhq.atlassian.net/browse/LIVE-2258

**Note that these pages are dropped:** (correct me if i'm wrong, i don't think these pages are used at all)

- erc20: was a demo made before ERC20 were in live.
- currencies & assets & derivations: was a way to see all the currencies & co. blockchain support team is free to recreate if needed.
- countervalues: was a way to see all the countervalues. live hub team is free to recreate if needed.
- apps: was a way to list all the supported apps on the manager. devices team is free to recreate this later if needed.
- manager: was the "v2" poc. now it's real in LL.
- mcu-repair: same, it's in LL. it's probably dead by now, but if people requires it, we can always put it back.
- bridgestream, bridgetest, qrledger, qrstreambenchmark: was tmp ways to debug  things during LLM development, also performance


ghost (approved):
:pray:

juan-cortes (approved):
🔥🔥🔥

PR #1946, opened 2022-05-09
Many fixes

emmanuelm41:
## Context (issues, jira)
Filecoin integration

## Description / Usage
- add derivation mode on account ids
- increase timeout on specs
- fix mutation flag on bot cli cmd

## Expectations

### DIFF
--- a/src/bot/index.ts
+++ b/src/bot/index.ts
@@ -77,7 +77,7 @@ export async function bot({ currency, family, mutation }: Arg = {}) {
         if (mutation) {
           spec = {
             ...spec,
-            mutations: spec.mutation.filter((m) =>
+            mutations: spec.mutations.filter((m) =>
               new RegExp(mutation).test(m.name)
             ),
           };
--- a/src/families/filecoin/specs.ts
+++ b/src/families/filecoin/specs.ts
@@ -16,7 +16,7 @@ const filecoinSpecs: AppSpec<Transaction> = {
     appName: "Filecoin",
   },
 
-  testTimeout: 2 * 60 * 1000,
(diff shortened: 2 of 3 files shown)


src/families/filecoin/specs.ts
```diff
@@ -16,7 +16,7 @@ const filecoinSpecs: AppSpec<Transaction> = {
     appName: "Filecoin",
   },
 
-  testTimeout: 2 * 60 * 1000,
+  testTimeout: 5 * 60 * 1000,
```
haammar-ledger, line 19: This seems like a lot, users would need to wait 5min to see their transactions confirmed?
Any chance confirmed transactions get seen sooner in the future?

src/bot/index.ts
```diff
@@ -77,7 +77,7 @@ export async function bot({ currency, family, mutation }: Arg = {}) {
         if (mutation) {
           spec = {
             ...spec,
-            mutations: spec.mutation.filter((m) =>
+            mutations: spec.mutations.filter((m) =>
```
haammar-ledger, line 80: Nice catch @emmanuelm41!
cc @gre

ghost (approved):

src/families/filecoin/specs.ts
emmanuelm41: In practice, it took less than 3 minutes. A little more than 2 minutes. I think in the future we will improve it tuning configuration parameters to make it faster.

src/bot/index.ts
emmanuelm41: I tried to use it and it was failing haha

PR #1948, opened 2022-05-11
LIVE-1940 Ledger first inside cosmos validator list

henri-ly:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1940

## Description / Usage

We change how the delegate work on Cosmos according to the new UI.
Also we have some code to clean also because some are unused

## Expectations

### DIFF
--- a/src/families/cosmos/js-buildTransaction.ts
+++ b/src/families/cosmos/js-buildTransaction.ts
@@ -53,22 +53,24 @@ export const buildTransaction = async (
       if (!transaction.validators || transaction.validators.length < 1) {
         isComplete = false;
       } else {
-        transaction.validators.forEach((validator) => {
-          if (!validator.address || validator.amount.lte(0)) {
-            isComplete = false;
-          }
+        const validator = transaction.validators[0];
+        if (!validator) {
+          isComplete = false;
+          break;
+        } else if (!validator.address || transaction.amount.lte(0)) {
+          isComplete = false;
+        }
 
-          msg.push({
-            typeUrl: "/cosmos.staking.v1beta1.MsgDelegate",
-            value: {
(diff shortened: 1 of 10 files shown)


src/families/cosmos/react.ts
```diff
@@ -163,3 +164,35 @@ export function useMappedExtraOperationDetails({
...
+    .filter((validator) => validator.commission !== 1.0)
+    .sort((a, b) => b.votingPower - a.votingPower);
+
+  // move Ledger validator to the first position
+  const ledgerValidator = sortedValidators.find(
+    (v) => v.validatorAddress === LEDGER_VALIDATOR_ADDRESS
+  );
+
+  if (ledgerValidator) {
+    const removeLedgerDuplicate = sortedValidators.filter(
```
haammar-ledger, line 189: Var naming: maybe something more like `sortedValidatorsLedgerFirst`?

src/families/cosmos/js-buildTransaction.ts
```diff
@@ -53,22 +53,23 @@ export const buildTransaction = async (
...
-            value: {
-              delegatorAddress: account.freshAddress,
-              validatorAddress: validator.address,
-              amount: {
-                denom: account.currency.units[1].code,
-                amount: validator.amount.toString(),
-              },
+        const validator = transaction.validators[0];
+        if (!validator) {
+          break;
```
haammar-ledger, line 58: Isn't that an error case? Or should set `isComplete = false`?

ghost (commented):
Wondering if test datasets are covering the delegate scenarios, as you made some changes in `getTransactionStatus` and the tests are still passing without changes.

ghost (approved):

PR #1953, opened 2022-05-12
LIVE-1292 Cover Polkadot existential deposit edge case

alexalouit:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1292

## Description / Usage

When trying to emptied an account, it doesn't work.
This happens when the balance is at ~1 DOT, this is a limitation of the Polkadot blockchain.
The logic of Ledger Live Common was not adapted, this PR fix this.

This PR also contains a fix on the optimistic operation which does not contain the correct type when it is a send max.

## Expectations

### DIFF
--- a/src/families/polkadot/js-signOperation.ts
+++ b/src/families/polkadot/js-signOperation.ts
@@ -42,19 +42,14 @@ const MODE_TO_PALLET_METHOD = {
 };
 
 const getExtra = (type: string, account: Account, transaction: Transaction) => {
-  const extra = MODE_TO_PALLET_METHOD[transaction.mode]
-    ? {
-        palletMethod:
-          MODE_TO_PALLET_METHOD[
-            transaction.mode === "bond" && !isFirstBond(account)
-              ? "bondExtra"
-              : transaction.mode
-          ],
-      }
-    : {};
+  const extra = {
+    palletMethod: MODE_TO_PALLET_METHOD[transaction.mode],
+  };
 
   if (transaction.mode == "send" && transaction.useAllAmount) {
     extra.palletMethod = MODE_TO_PALLET_METHOD["sendMax"];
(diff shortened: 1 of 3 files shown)


henri-ly (approved):

src/families/polkadot/logic.ts
```diff
@@ -3,7 +3,7 @@ import type { Account, OperationType } from "../../types";
 import type { Transaction } from "./types";
 import { getCurrentPolkadotPreloadData } from "./preload";
 
-export const EXISTENTIAL_DEPOSIT = new BigNumber(10000000000);
+export const EXISTENTIAL_DEPOSIT = new BigNumber(11000000000);
```
ghost: `EXISTENTIAL_DEPOSIT` seems to be used in many places and different cases, so to be cautious about regressions could you define here a separate var with value 0.1 DOT and only use it with the change you did in `getSendTransactionStatus`.
With also just a short comment to explain that we take a safety buffer of 0.1 DOT in addition to normal existential deposit.

src/families/polkadot/js-signOperation.ts
```diff
@@ -52,6 +53,10 @@ const getExtra = (type: string, account: Account, transaction: Transaction) => {
       }
     : {};
 
+  if (transaction.mode == "send" && transaction.useAllAmount) {
+    extra.palletMethod = MODE_TO_PALLET_METHOD["sendMax"];
+  }
+
```
ghost: Code readability : can you please rework the `const extra = ...` definition above to merge it with this?
You can rely on if/else as you did here, I think it's way more readable than the multiple ternary operators above.

ghost (changes requested):
Please put the link to the JIRA ticket in the description of the PR + cleanup the title a bit it looks weird :smile:

ghost (approved):
:ok_hand: :+1:

PR #1957, opened 2022-05-12
LIVE-1292 Cover Polkadot existential deposit edge case (#1953)

ghost:
## Context (issues, jira)

https://ledgerhq.atlassian.net/browse/LIVE-1292

## Description / Usage

When trying to emptied an account, it doesn't work.
This happens when the balance is at ~1 DOT, this is a limitation of the Polkadot blockchain.
The logic of Ledger Live Common was not adapted, this PR fix this.

This PR also contains a fix on the optimistic operation which does not contain the correct type when it is a send max.
## Expectations


ghost:
Bot says :heavy_check_mark: !

henri-ly (approved):

PR #1958, opened 2022-05-12
review useAllAmount operations to show the correct errors

emmanuelm41:
## Context (issues, jira)
Filecoin integration

## Description / Usage
Fix some checks we do when the user choose to transfer the entire balance

## Expectations


src/families/filecoin/bridge/account.ts
```diff
@@ -84,10 +84,11 @@ const getTransactionStatus = async (
   const estimatedFees = calculateEstimatedFees(gasFeeCap, gasLimit);
 
   const totalSpent = useAllAmount ? balance : amount.plus(estimatedFees);
-  amount = useAllAmount ? balance.minus(estimatedFees) : amount;
-
-  if (amount.lte(0)) errors.amount = new AmountRequired();
   if (totalSpent.gt(a.spendableBalance)) errors.amount = new NotEnoughBalance();
+  else {
```
ghost: Could you please add { } to the if above, for consistency? Thanks.

emmanuelm41: Done!

src/derivation.ts
```diff
@@ -160,7 +160,7 @@ const modes = Object.freeze({
     tag: "legacy",
   },
   // glif normal derivation
-  glifnormal: {
+  glif: {
```
gre, line 163: we are not supposed to change a derivation id after it went to production (did it went into production?) otherwise it will break existing users "migrating" to the new app with their old data.

emmanuelm41: I did not go to production yet. That is why I have changed it!

src/families/filecoin/test-dataset.ts
```diff
@@ -11,7 +11,7 @@ import { fromTransactionRaw } from "../filecoin/transaction";
 import { CurrenciesData } from "../../types";
 
 const SEED_IDENTIFIER = "f1zx43cf6qb6rd5e4okl7lexnjumxe5toqj6vtr3i";
-const ACCOUNT_1 = "t1d2xrzcslx7xlbbylc5c3d5lvandqw4iwl6epxba";
+const ACCOUNT_1 = "f1zx43cf6qb6rd5e4okl7lexnjumxe5toqj6vtr3i";
```
ghost: Just checking, is this account frozen, i.e will it be never used again?
We need to use frozen accounts here so that we don't need to regenerate the snapshots all the time.

src/families/filecoin/test-dataset.ts
```diff
@@ -96,26 +104,6 @@ const filecoin: CurrenciesData<Transaction> = {
...
-            recipient: ACCOUNT_1,
-            amount: "0",
-          }),
-          expectedStatus: {
-            errors: {
-              amount: new AmountRequired(),
-            },
-            warnings: {},
-          },
-        },
```
ghost: Any reason for removing that one? Testing `AmountRequired` seems to make sense.

src/families/filecoin/test-dataset.ts
emmanuelm41: I didn't know that! Let me change it then so we can leave an account frozen!

emmanuelm41: The account now is frozen!

src/families/filecoin/test-dataset.ts
emmanuelm41: I put the test back where it was!

src/derivation.ts
gre: ok thanks! 👍

ghost (approved):

PR #1960, opened 2022-05-12
LIVE-1560 Algorand NotEnoughBalance errors in bot

alexalouit:
## Context (issues, jira)

Compute fees before to max spendable estimation

https://ledgerhq.atlassian.net/browse/LIVE-1560

## Expectations

### DIFF
--- a/src/families/algorand/js-estimateMaxSpendable.ts
+++ b/src/families/algorand/js-estimateMaxSpendable.ts
@@ -5,6 +5,7 @@ import type { AlgorandTransaction } from "./types";
 import { computeAlgoMaxSpendable } from "./logic";
 import { createTransaction } from "./js-prepareTransaction";
 import { getAbandonSeedAddress } from "@ledgerhq/cryptoassets";
+import { getEstimatedFees } from "./js-getFeesForTransaction";
 
 export const estimateMaxSpendable = async ({
   account,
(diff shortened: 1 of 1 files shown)


src/families/algorand/js-estimateMaxSpendable.ts
```diff
@@ -30,6 +31,8 @@ export const estimateMaxSpendable = async ({
     useAllAmount: true,
   };
 
+  tx.fees = await getEstimatedFees(mainAccount, tx);
```
ghost: Maybe works, but fees are supposed to be set before already, in `prepareTransaction` [here](https://github.com/LedgerHQ/ledger-live-common/blob/5b012ceeca27be920ca31f3227b2b7bdcf51882c/src/families/algorand/js-prepareTransaction.ts#L39).

Would be good to understand how we get here without tx.fees :thinking:

alexalouit: Maybe because https://github.com/LedgerHQ/ledger-live-common/blob/5b012ceeca27be920ca31f3227b2b7bdcf51882c/src/families/algorand/js-prepareTransaction.ts#L27 ?

ghost: Bingo :slightly_smiling_face: good catch.
So I agree the simplest is to call `getEstimatedFees` fees twice indeed...
But instead of setting `tx.fees` then testing it later, can you just `const fees = ...` then use that below? (L50)
Thanks!

alexalouit: fix in 2efced6c85d77963cc0cbde3f0c00a64d9c47de8

ghost (approved):

ghost:
This should've been a `bugfix/` branch, sorry again @lambertkevin :pray:

PR #1961, opened 2022-05-13
[LIVE-1567] Feature: Update NFT types to add multiple medias

lambertkevin:
## Context (issues, jira)
https://ledgerhq.atlassian.net/browse/LIVE-1795

## Description / Usage

We're changing the type of media(s) returned by the metadata service.

## Expectations

### DIFF
--- a/src/types/nft.ts
+++ b/src/types/nft.ts
@@ -3,10 +3,20 @@ import { CryptoCurrencyIds } from ".";
 
 export type NFTStandards = "ERC721" | "ERC1155";
 
+export type NFTMediaSizes = "preview" | "big" | "original";
+
+export type NFTMedias = Record<
+  NFTMediaSizes,
+  {
+    uri: string;
+    mediaType: string; // mime-type
+  }
+>;
+
 export type NFTMetadata = {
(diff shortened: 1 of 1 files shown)


src/types/nft.ts
```diff
@@ -3,10 +3,20 @@ import { CryptoCurrencyIds } from ".";
...
+    uri: string;
+    mediaType: string; // mime-type
+  }
+>;
+
 export type NFTMetadata = {
   tokenName: string | null;
   nftName: string | null;
-  media: string | null;
+  medias: NFTMedias;
```
ghost, line 19: I'm just concerned that impacts the persisted cache (what's in app.json on LLD).
For safety, migration should be tested from a build without this change, with some data in cache, to a build with this change.

lambertkevin: `accounts.nfts` is recreated from scratch each time, so nfts are redefined completely each sync. From all my testing it was never an issue, even with accounts already in cache 👌

lambertkevin:
Closed for the monorepo: https://github.com/LedgerHQ/ledger-live/pull/68

PR #1962, opened 2022-05-14
Fixes on max spendable balance function

emmanuelm41:
## Context (issues, jira)
Filecoin integration

## Description / Usage

- estimate fees when there is not recipient address
- avoid to generate spendable balance as negative

## Expectations

### DIFF
--- a/src/families/filecoin/bridge/account.ts
+++ b/src/families/filecoin/bridge/account.ts
@@ -116,29 +116,27 @@ const estimateMaxSpendable = async ({
   const a = getMainAccount(account, parentAccount);
   const { address } = getAddress(a);
 
+  const recipient = transaction?.recipient;
+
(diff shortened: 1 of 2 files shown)


src/families/filecoin/bridge/account.ts
```diff
@@ -116,29 +116,27 @@ const estimateMaxSpendable = async ({
   const a = getMainAccount(account, parentAccount);
   const { address } = getAddress(a);
 
+  const recipient = transaction?.recipient;
+
   if (!validateAddress(address).isValid) throw new InvalidAddress();
```
ghost, line 121: Unrelated to this PR but I just noticed, probably no need to validate the user's own account address since it's from the Live itself?...

ghost (approved):

PR #1963, opened 2022-05-16
useCurrenciesByMarketcapWithStatus

ThomasLaforge:
## Context (issues, jira)

Current hook to get data by marketcap was async and missing some info.
I created this new one to get loading status and errors to handle those informations on UI.
Here is the issue [LIVE-1982].

Using this one I will be able to get an empty state until the loading or error state switch.

## Description / Usage

```tsx
const { loading, error, currencies } = useCurrenciesByMarketcapWithStatus(
    cryptoCurrencies,
);
```

## Expectations

- [*] **No impact: The changes of this PR have ZERO impact on the userland.** Meaning, we can use these changes without modifying LLD/LLM at all. It will be a "noop" and the maintainers will be able to bump it without changing anything.

[LIVE-1982]: https://ledgerhq.atlassian.net/browse/LIVE-1982?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ

### DIFF
--- a/src/currencies/sortByMarketcap.ts
+++ b/src/currencies/sortByMarketcap.ts
@@ -107,3 +107,36 @@ export const useCurrenciesByMarketcap = <C extends Currency>(
   const tickers = useMarketcapTickers();
   return tickers ? sortByMarketcap(currencies, tickers) : currencies;
 };
+
+type CurrenciesByMarketcapResult = {
+  loading: boolean;
+  error?: Error;
+  currencies?: any[];
+};
+
+export const useCurrenciesByMarketcapWithStatus = <C extends Currency>(
+  currencies: C[]
+): CurrenciesByMarketcapResult => {
+  const [status, setStatus] = useState<CurrenciesByMarketcapResult>({
+    loading: false,
+  });
+
+  useEffect(() => {
+    function getCurrenciesByMarketcap() {
+      setStatus({ loading: true });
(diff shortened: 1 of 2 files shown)


src/currencies/sortByMarketcap.ts
```diff
@@ -107,3 +107,36 @@ export const useCurrenciesByMarketcap = <C extends Currency>(
...
   return tickers ? sortByMarketcap(currencies, tickers) : currencies;
 };
+
+type CurrenciesByMarketcapResult = {
+  loading: boolean;
+  error?: Error;
+  currencies?: any[];
+};
+
+export const useCurrenciesByMarketcapWithStatus = <C extends Currency>(
```
LFBarreto, line 117: @ThomasLaforge can you try and add a unit test case for this hook. should be pretty straight forward creating one starting from the other hook in the file. :pray:

LFBarreto (changes requested):

src/currencies/sortByMarketcap.ts
ThomasLaforge: There is no tests for the equivalent hook. Only for function in it.
I can check how to test a hook ^^

PR #1964, opened 2022-05-16
Solana fixes

konoart:
- Increase tx confirmation timeout to 100 seconds
- Rename simulation failed error


konoart:
@haammar-ledger

